PR #30, opened 2016-06-08
Fix stack alignment for ukvm target

mato:
The x86_64 ABI mandates that (%rsp + 8) be a multiple of 16 on entry to
a user function. GCC relies on this when using SSE instructions and will
ensure correct alignment is preserved in subsequent function calls.

Per testing, virtio target already appears (possibly not intentionally)
to have the correct stack alignment.

### DIFF
--- a/ukvm/ukvm.c
+++ b/ukvm/ukvm.c
@@ -1094,7 +1094,7 @@ int main(int argc, char **argv)
         .rax = 2,
         .rbx = 2,
         .rflags = 0x2,
-        .rsp = GUEST_SIZE,
+        .rsp = GUEST_SIZE - 8,  // x86_64 ABI requires ((rsp + 8) % 16) == 0
         .rdi = GUEST_SIZE,	// size arg in kernel main
         .rsi = kernel_end,	// kernel_end arg in kernel main
         .rdx = BOOT_ARG_AREA,


mato:
FYI/background: Ran into this today while trying to print floats from OCaml which was causing a GPF on `MOVAPS`. Thankfully I remembered doing this before (https://github.com/rumpkernel-attic/rumprun-xen/issues/18) so once identified the fix is trivial.

djwillia:
Thanks @mato, this looks like something terrible to track down.  

If I understand it correctly, you are pushing onto the stack to fix the alignment.  For ukvm, the stack alignment upon entering `kernel_main` entirely depends on where `ukvm` places the stack (https://github.com/djwillia/solo5/blob/master/ukvm/ukvm.c#L1097).  Right now it's being set to GUEST_SIZE, which, if user-supplied in the future may change the alignment and whether the `push` should happen or not.  

If, on the other hand, we force the stack alignment to be correct from `ukvm`, will GCC insert extra stack space if necessary to maintain the correct 16-byte alignment before every function call?  In other words, if `kernel_main` ever changes to have another local variable less than 16 bytes long, will there be problems, or will GCC know to take care of it?

mato:
You're right; the correct place to fix this is by setting `kvm_regs.rsp` to `GUEST_SIZE - 8`. GCC will make sure the correct alignment is used in any later function calls. I'll update this pull request / make a new one with the change.

mato:
@djwillia I've force-pushed a new commit to the branch with the fix in ukvm. (I should have probably just closed this PR and opened a new one, but this seemed easier)

ricarkol:
GUEST_SIZE should be aligned as well. An assert should be enough, but that can wait until we actually get GUEST_SIZE as an argument.

PR #33, opened 2016-06-13
Add solo5_poll(), UKVM_PORT_POLL

mato:
- solo5_poll API to block and wait for I/O. Replaces solo5_cpu_block().
- UKVM_PORT_POLL hypercall to implement solo5_poll() for ukvm target.

### DIFF
--- a/kernel/virtio/time.c
+++ b/kernel/virtio/time.c
@@ -66,11 +66,6 @@ int sleep(uint32_t secs) {
     return 0;
 }
 
-void solo5_cpu_block(uint64_t until_nsecs)
-{
-    cpu_block(until_nsecs);
-}
-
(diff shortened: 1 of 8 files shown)


kernel/virtio/time.c
```diff
@@ -101,3 +96,23 @@ void sleep_test(void) {
...
 }
+
+int solo5_poll(uint64_t until_nsecs)
+{
+    /*
+     * cpu_block() as currently implemented will only poll for the maximum time
+     * the PIT can be run in "one shot" mode. Loop until either I/O is possible
+     * or the desired time has been reached.
+     */
+    for (;;) {
```
mato: This loop has a race with the virtio interrupt handler, fix coming up shortly.

PR #49, opened 2016-06-30
use $(MAKE) and $(CC) instead of make and gcc

hannesm:
(trying to compile this on FreeBSD)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,12 +19,12 @@ all: ukvm_target virtio_target
 
 .PHONY: virtio_target
 virtio_target:
-	make -C kernel virtio
+	$(MAKE) -C kernel virtio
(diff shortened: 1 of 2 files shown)


ricarkol:
Hi hannesm, while you are at it, would you mind changing this one as well please? https://github.com/djwillia/solo5/blob/master/ukvm/Makefile

hannesm:
@ricarkol sure, I also found 3 more `make` occurences... now heading off to ocaml-freestanding

ricarkol:
Thanks Hannes

mato:
This looks good to me, @ricarkol do you want to merge it?

ricarkol:
merged

PR #51, opened 2016-06-30
set style to "Linux style but with 4 spaces instead of tabs"

djwillia:
This should address most of the style for the .c and .h files.  For the most part I did not change files that were direct copies from other projects (e.g., `malloc`, clock code, etc.)  No functional changes are expected. 

I also checked in `checkpatch.pl` from Linux with the leading tab checks and a few other Linux specific checks commented out.  It doesn't catch everything and there are a few things (`extern`s mostly) that are still being warned about, but it should give us a good first pass for style checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,10 @@ and subject to change.
 We also have some simple standalone unikernels written in C to test
 Solo5, see `kernel/test_*` for these.
 
+The coding style for this project is "Linux with 4 spaces instead of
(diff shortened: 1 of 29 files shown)


ricarkol:
Looks good to me

ricarkol:
If anything,  you could squash the three commits.

mato:
Thanks Dan. Reviewed and tested, LGTM. Squashed and merged.

PR #52, opened 2016-07-01
Fix linker scripts

mato:
- switch to --nmagic instead of -z max-page-size, the latter doesn't
  work on FreeBSD (thanks @hannesm for reporting and investigation)
- clean up linker scripts for both virtio and ukvm targets


hannesm:
tested on FreeBSD-11 (amd64, bhyve); works :+1:

PR #53, opened 2016-07-01
use -print-file-name=include instead of -print-search-dirs and sed magic

hannesm:
this is (imho) less brittle...


mato:
Looks good to me, checks out on Alpine (musl) as well. Thanks!

PR #54, opened 2016-07-01
fix C warning (clang 3.8.1)

hannesm:
virtio/virtio.c:240:49: error: missing field 'gso_type' initializer
      [-Werror,-Wmissing-field-initializers]
static struct virtio_net_hdr virtio_net_hdr = {0};


hannesm:
NB: clang version is 3.8.0, not 3.8.1... sorry

PR #55, opened 2016-07-01
CC is cc, not gcc

hannesm:
at least on all the machines I have access to, even on ubuntu linux, you get a cc binary... cc is also common on MacOSX and FreeBSD for the C compiler... (where there might be no gcc installed)...


hannesm:
(not sure how alpine does it, please check before merge)

PR #56, opened 2016-07-01
Revert changes to linker scripts / --nmagic

mato:
--nmagic has side effects which a) break ukvm and b) we don't want in general (results in a single RWX phdr for the entire kernel). Revert those changes, and introduce a possible workaround for FreeBSD.


mato:
For the record, the workaround is here: https://sourceware.org/ml/binutils/2009-08/msg00512.html
(don't set OUTPUT_FORMAT() in linker scripts otherwise -z max-page-size does not work)

hannesm:
this works on FreeBSD (compiles, and the resulting test_hello.virtio) boots!

Issue #26, opened 2016-06-02
Better arg handling for ukvm


djwillia:
`ukvm` cmdline argument handling needs to be fixed.  From @avsm on slack:

> also support the gnu convention of `--` after `<binary>`? Otherwise `-gdb` is ambiguous

Also, the network and disk information will need to be optional when those parts of ukvm are optional modules.

mato:
@djwillia In #41 I had to change the command line for ukvm to be `<disk> <tap> <elf> [--gdb] [args]`, i.e. move `--gdb` to come before the unikernel arguments. I've deliberately not done more than that now as I don't want to step on the work you're doing on the `modularize` branch which affects ukvm command line parsing.

mato:
Fixed ages ago (mostly by #57)

PR #57, opened 2016-07-01
modularize ukvm

djwillia:
This breaks ukvm into a few different modules.  The idea is that it can be built without one or more of the modules, and I'd like to eventually make OPAM know about how to build a monitor for a unikernel based on what it needs.  For the first commit, I'm checking in a ukvm build config (`Makefile.ukvm`) which includes all modules, so functionally everything should stay the same.

This is a bit of an annoying change for a few reasons:
- it moves a bunch of code into different files
- it changes the argument order to `ukvm` yet again (see issue #26 )

This also bumps up on a few things that are changing or will change
- it only supports one instance of each type of device (e.g., net, disk)
- the `poll` code now gets its fd list from the modules (still only net at the moment)

### DIFF
--- a/ukvm/ukvm.c
+++ b/ukvm/ukvm.c
@@ -59,8 +59,9 @@ struct ukvm_module *modules[] = {
 #ifdef UKVM_MODULE_GDB
     &ukvm_gdb,
 #endif
+    NULL,
 };
-#define NUM_MODULES (sizeof(modules) / sizeof(struct ukvm_module *))
+#define NUM_MODULES ((sizeof(modules) / sizeof(struct ukvm_module *)) - 1)
 
 /*
  * Memory map:
@@ -175,6 +176,7 @@ void setup_boot_info(uint8_t *mem,
         cmdline_free -= alen;
         cmdline_p += alen;
     }
+
 }
 
 ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
@@ -727,6 +729,20 @@ void sig_handler(int signo)
     exit(0);
 }
 
+static void usage(const char *prog)
+{
+    int m;
+
+    printf("usage: %s [ CORE OPTIONS ] [ MODULE OPTIONS ] KERNEL", prog);
+    printf(" [ -- ] [ ARGS ]\n");
+    printf("Core options:\n");
+    printf("    --help (display this help)\n");
+    printf("Compiled-in module options:\n");
+    for (m = 0; m < NUM_MODULES; m++)
+        printf("    %s\n", modules[m]->usage());
+    exit(1);
+}
+
 int main(int argc, char **argv)
 {
     int kvm, vmfd, vcpufd, ret;
@@ -735,36 +751,43 @@ int main(int argc, char **argv)
     size_t mmap_size;
     uint64_t elf_entry;
     uint64_t kernel_end;
-    int num_args = 0;
-    char *first_arg;
-    int i;
+    const char *prog;
+    const char *elffile;
+    int matched;
 
-    if (argc < 2) {
-        int m;
+    prog = basename(*argv);
+    argc--;
+    argv++;
 
-        printf("usage: ukvm <elf> [args] --");
-        for (m = 0; m < NUM_MODULES; m++)
-            printf(" %s", modules[m]->usage());
-        printf("\n");
-        return 1;
-    }
+    if (argc < 1)
+        usage(prog);
 
-    const char *elffile = argv[1];
-
-    first_arg = argv[2];
-    for (i = 2; i < argc; i++) {
-        if (!strcmp("--", argv[i]))
-            break;
-        num_args++;
-    }
-
-    printf("Found %d args (%d left)\n", num_args, argc - num_args - 3);
-    for (i = num_args + 3; i < argc; i++) {
+    do {
         int j;
 
-        for (j = 0; j < NUM_MODULES; j++)
-            if (!modules[j]->handle_cmdarg(argv[i]))
+        if (!strcmp("--help", *argv))
+            usage(prog);
+
+        matched = 0;
+        for (j = 0; j < NUM_MODULES; j++) {
+            if (!modules[j]->handle_cmdarg(*argv)) {
+                matched = 1;
+                argc--;
+                argv++;
                 break;
+            }
+        }
+    } while (matched);
+
+    elffile = *argv;
+    argc--;
+    argv++;
+
+    if (argc) {
+        if (strcmp("--", *argv))
+            usage(prog);
+        argc--;
+        argv++;
     }
 
     if (signal(SIGINT, sig_handler) == SIG_ERR)
@@ -825,7 +848,7 @@ int main(int argc, char **argv)
     setup_system(vcpufd, mem);
 
     /* Setup ukvm_boot_info and command line */
-    setup_boot_info(mem, GUEST_SIZE, kernel_end, num_args, &first_arg);
+    setup_boot_info(mem, GUEST_SIZE, kernel_end, argc, argv);
 
     /*
      * Initialize registers: instruction pointer for our code, addends,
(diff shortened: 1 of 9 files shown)


ricarkol:
Hey dan, looks great, just nits:

1. usage for the blk module should be: `disk=xxx` and not `blk=xxx` (in ukvm-blk.c)

this could be part of another PR:
2. update `make ukvm`

ukvm/ukvm.c
```diff
@@ -41,24 +41,26 @@
...
+    &ukvm_blk,
+#endif
+#ifdef UKVM_MODULE_NET
+    &ukvm_net,
+#endif
+#ifdef UKVM_MODULE_GDB
+    &ukvm_gdb,
+#endif
+};
+#define NUM_MODULES (sizeof(modules) / sizeof(struct ukvm_module *))
```
mato: This can be a NULL-terminated array of `(struct ukvm_module *)`, then we don't need the `NUM_MODULES` stuff.

ukvm/ukvm.c
```diff
@@ -577,93 +575,6 @@ void ukvm_port_nanosleep(uint8_t *mem, void *data, struct kvm_run *run)
...
-    struct ukvm_netread *rd = (struct ukvm_netread *) (mem + mem_off);
-    struct timeval zero;
-    fd_set netset;
-    int ret;
-
-    FD_ZERO(&netset);
-    FD_SET(netfd, &netset);
-    zero.tv_sec = 0;
-    zero.tv_usec = 0;
-    ret = select(netfd + 1, &netset, NULL, NULL, &zero);
```
mato: If we set `netfd` to non-blocking then this `select()` call can be removed.

ukvm/ukvm.c
```diff
@@ -577,93 +575,6 @@ void ukvm_port_nanosleep(uint8_t *mem, void *data, struct kvm_run *run)
...
-    if (ret <= 0) {
-        rd->ret = -1;
-        return;
-    }
-
-    rd->len = read(netfd, mem + (uint64_t) rd->data, rd->len);
```
mato: `assert(rd->len > 0)` so that we don't drop errors on the floor? We should also check and handle `EINTR` on all these syscalls.

ukvm/ukvm.c
```diff
@@ -800,61 +710,15 @@ static int vcpu_loop(struct kvm_run *run, int vcpufd, uint8_t *mem,
...
-        return err;
-    }
-
-    /* if the operation was successful, write back the name of the
-     * interface to the variable "dev", so the caller can know
-     * it. Note that the caller MUST reserve space in *dev (see calling
-     * code below)
-     */
-    strcpy(dev, ifr.ifr_name);
+    for (i = 0; i < NUM_MODULES; i++)
```
mato: See comment about NULL-terminating `modules[]`.

ukvm/ukvm.c
```diff
@@ -863,78 +727,48 @@ void sig_handler(int signo)
...
+    const char *elffile = argv[1];
+
+    first_arg = argv[2];
+    for (i = 2; i < argc; i++) {
+        if (!strcmp("--", argv[i]))
+            break;
+        num_args++;
+    }
+
+    printf("Found %d args (%d left)\n", num_args, argc - num_args - 3);
```
mato: Left-over debug print?

mato:
Looking good. I've added a bunch of comments as line notes, not all of which need to be addressed in this PR, some just came up as I read through the code.

I'd like to propose a different syntax for command line arguments. The current setup results in (with all modules configured):

    usage: ukvm <elf> [args] -- blk=<disk.img> net=<tap100> --gdb

Following established UNIX conventions and the principle of least astonishment, something like the following comes to mind. I've also future-proofed it by adding a placeholder for "core" options:

    usage: ukvm [ CORE OPTIONS ] [ MODULE OPTIONS ] KERNEL [ -- ] [ ARGS ]
        Compiled-in modules:
        --disk=IMAGE (description...)
        --net=TAP (description...)
        --gdb (description...)

However, this still leaves two points that should be addressed:

1. A good syntax for specifying other options to modules, such as a user-specified MAC address for `--net`. It'd be nice to come up with something better than QEMU!
2. How to deal with multiple instances of modules. The simplest scheme I can think of is just repeating `--net` and `--disk` multiple times, each creating a new instance.

Thoughts?

djwillia:
@ricarkol I've updated `make ukvm` and changed the argument name.
@mato I changed the argument order to what you suggested, but have not yet made the other changes.  

In terms of NUM_MODULES, I may not be seeing the reason that it should be avoided?  I can change the loops to iterate until NULL, but it seems convenient to know the number of modules to statically allocate structures (e.g., https://github.com/djwillia/solo5/blob/ff0e04e1dcf80a8ccead0a4afab0ed483f05b558/ukvm/ukvm.c#L607).

I'm planning to do the nonblock and check to the ukvm-net tomorrow.

mato:
@djwillia The NUM_MODULES change is just a suggestion, feel free to use what works for you. Once we have multiple instances oif modules the associated structures will not match NUM_MODULES anyway. Otherwise, looks good. Let me know once you want to merge and I'll test this, plus prepare the associated changes for unikernel-runner.

ricarkol:
Thanks dan

djwillia:
@mato I think the issue of supporting multiple device instances is one that may require some more thought (I'm sure NUM_MODULES will change, but it pushes up against this).              
                                                                                                                                                            
Usually, the (general-purpose) monitor would expose as many devices as it discovers on the host or is configured to expose and the unikernel would discover whether enough interfaces (e.g., two NICs) were available at runtime and either adapt its behavior or fail (at runtime) otherwise.  This approach isn't minimal in the sense that we've been advocating with `ukvm`; if the unikernel doesn't need one of them, it's still possible that it could mis-configure one (or use it as an attack vector as in VENOM, for example).  This approach also seems to be headed down a slippery slope where two different applications are implemented as the same unikernel: the one-interface version and the two-interface version.                                                                 
                                                                                                                                                            
With different types of devices it seems that, at compile time, dependencies from the unikernel can dictate what modules `ukvm` should be configured with (for that unikernel).  Implementation-wise, this seems straightforward, as we can either include the module's code or not (at compile time).                 
                                                                                                                                                            
With different instances of the same type of device, conceptually it should be the same: a unikernel (through its implementation or compile-time configuration) should know (at compile time) whether it needs multiple interfaces (e.g., two NICs) or not. Implementation-wise, this seems more tricky as we don't need to include the module's code multiple times, but may need to adapt the module's code to force exactly N instances to be available/configured, where N is the number of interfaces the unikernel depends on.                                                                                                      
                                                                                                                                                            
This isn't the sort of thing we need to decide on this PR, but I thought I'd put the comment here as something to think about.

djwillia:
@mato I think it's ready to merge!

mato:
Couple of things from a basic smoketest:
1. Does not build on my Debian stretch box due to `#include` conflicts in `ukvm-net.c`. Could you remove `linux/if_packet.h`, `net/ethernet.h` and `net/if.h`? These are not required and fix the build error.
2. `ukvm test_hello.ukvm` (i.e. no disk or net) fails with `ukvm: couldn't setup modules`, so there's a logic error somewhere as it's ok to run with no disk or net, if not terribly useful.
3. Invalid arguments, e.g. `ukvm --foo` fail with `ukvm: couldn't open elf: No such file or directory`, should at least include `--foo` in the error message, ideally should complain that this is an invalid option.

djwillia:
@mato thanks, the new commits should address your points 1 and 3 and partially 2.  I left the code in such a way that the module arguments are mandatory (e.g., even if you don't want a disk) but now there is an error message.  If there aren't other problems, I suggest we can merge this but leave issue #26 open to follow up on the argument order and requirements?  What do you think?

mato:
@djwillia OK, looks good to me for an initial version, we can discuss the details of the modularization next week. Merging.

Issue #58, opened 2016-07-05
Memory leak with mirage-net-solo5(?) under load


mato:
Reporting this here as mirage-net-solo5 does not have issues enabled, and it's unclear if that is the root cause.

When running load tests against `static_website` using `siege -c 2 -b -t <TIME> http://10.0.0.2:8080/`, heap use grows rapidly until memory runs out.

Investigation so far:
- The unikernel survives a maximum siege of about `5s` before the ukvm memory allocation of 512MB runs out.
- The interaction of `io-page` with GC is unclear. `mirage-net-solo5` does have the tail-recursion fixes from `mirage-net-unix` in it, but these do not seem to help.
- Solo5 does not currently require page aligned buffers for I/O, so @hannesm made a quick test version of io-page to use `Cstruct` directly rather than the out-of-band page allocator. This can be OPAM-pinned with `opam pin io-page  git://github.com/hannesm/io-page#cstruct`.
- I added a call to `Gc.full_major ()` followed by logging of GC stats to `static_website` every two seconds to attempt to diagnose the GC behaviour.
- There is a separate bug (to be reported) where the unikernel appears to freeze (no GC logging output) after `siege` completes, and does not wake up until receiving at least one network packet.

Example output with GC stats and two `5s` siege runs -- GC values are `live_words` `minor_collections` `major_collections`:

```
2016-07-05 11:18.49: INF [gc] GC: 86061 2 2
2016-07-05 11:18.51: INF [gc] GC: 86068 3 4
2016-07-05 11:18.53: INF [gc] GC: 86068 4 6
2016-07-05 11:18.55: INF [gc] GC: 86068 5 8
2016-07-05 11:18.57: INF [gc] GC: 86068 6 10
Manager: configuration done
ARP: transmitting probe -> 10.0.0.1
ARP: updating 10.0.0.1 -> 0a:3f:94:7c:a1:51
2016-07-05 11:18.59: INF [gc] GC: 3345761 164 21
2016-07-05 11:19.01: INF [gc] GC: 6932473 444 26
//// (freeze until ping)
2016-07-05 11:19.13: INF [gc] GC: 7113271 555 28
2016-07-05 11:19.15: INF [gc] GC: 1038586 567 30
2016-07-05 11:19.17: INF [gc] GC: 93341 568 32
2016-07-05 11:19.19: INF [gc] GC: 93341 569 34
2016-07-05 11:19.21: INF [gc] GC: 93341 570 36
[...]

PR #59, opened 2016-07-05
malloc() and sbrk() fixes

mato:
- simplify sbrk() code path
- page alignment is not required in sbrk()
- enable giving back memory via sbrk() and tell malloc() about it
- don't abort in malloc() if memory exhausted, instead return NULL which
  should trigger OCaml GC

Related to #58, although does not fundamentally affect the issue.

### DIFF
--- a/kernel/malloc.c
+++ b/kernel/malloc.c
@@ -2,8 +2,7 @@
 
 /* dlmalloc configuration */
 #undef WIN32
-#define ABORT PANIC("aborting...")
-#define MORECORE_CANNOT_TRIM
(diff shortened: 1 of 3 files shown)


mato:
@djwillia Please review when you get a moment, should be uncontroversial.

djwillia:
Looks good to me; my only comment is that it looks pretty clear that `sbrk` should be in a common part of the src tree instead of the `virtio`/`ukvm`-specific parts.

mato:
@djwillia Thanks. sbrk() manipulates memory values which are defined in the target-specific mem.c, so that's why I've left it in there for now.

PR #60, opened 2016-07-06
ukvm: Fix overflow if solo5_poll() called too late

mato:
If solo5_poll() on ukvm is called after until_nsecs have passed then we
pass a negative value to UKVM_PORT_POLL which results in waiting for a
very long time.

Also clarify the UKVM_PORT_POLL interface to make it clear that the
timeout is relative, not absolute.


hannesm:
good catch!

Issue #61, opened 2016-07-06
[tracking] solo5 (virtio) on FreeBSD bhyve


hannesm:
status is: it works, but still needs some manual patches.  this uses clang instead of gcc.  the instructions are documented [here](https://hannes.nqsb.io/Posts/Solo5).  I haven't tested more than console yet.

Updates:
- [x] solo5 [PR build system fix](https://github.com/Solo5/solo5/pull/68)
- [x] solo5 [use `-nostdlibinc` and a `/tmp/include`](https://github.com/hannesm/solo5/commit/31c634280a884aee117e626635ebcc95d20aa2ae)
- [x] solo5: `cp` instructions at https://hannes.nqsb.io/Posts/Solo5
- [x] ocaml-freestanding [PR new openlibm](https://github.com/mirage/ocaml-freestanding/pull/4)
- [x] ocaml-freestanding [PR opam fix](https://github.com/mirage/ocaml-freestanding/pull/5)
- [x] ocaml-freestanding FreeBSD Makefile for OCaml (solution: conditional in configure.sh) 
- [x] ocaml-freestanding remove `-Wno-maybe-uninitialized` from nolibc (solution: same as above) see [FreeBSD branch](https://github.com/hannesm/ocaml-freestanding/commits/FreeBSD)
- [x] ocaml-freestanding: `rm sys/cdefs.h`
- [x] mirage using the solo5 PR
- [x] `export LD=/usr/local/bin/ld`

--- old stuff below ---

This issue is to track progress of upstreaming this into the solo5 mainline.  The biggest issue are include files.

- [x] libgcc.a is not needed to get linked in (it seems to be required on arm only, see #50 and https://github.com/mirage/mirage/pull/544), PR #63
- [x] solo5 uses `nostdinc`, but then includes `${CC} -print-file-name=include` as system includes.  It then includes `stddef.h` from there, which drags in `sys/_types.h` on FreeBSD.  There are other headers needed, such as `osreldate.h` (by openlibm).  In order to build `mirage-entropy-solo5` I needed to get rid of the gcc includes to get the intrinsics shipped with clang.  I've no proper idea how to deal with this best: provide our own `stddef.h`?  remove `nostdinc`?  Input is welcome how to solve this (note: the vanilla minios-xen uses `nostdinc` and does not use any system include (thus builds fine with clang); mirage modified minios drags in `stdint.h` and I wasn't able to get it working on FreeBSD).  e.g. grub completely ships their own header files.
[...]

Issue #50, opened 2016-06-30
Linking with libgcc.a unnecessary


mato:
@hannesm suggests that we remove `libgcc.a` from the Solo5 link list entirely. 

At least on `x86_64` our use of `-ffreestanding` to compile everything means it should not be required. See also the comments at https://github.com/mirage/mirage/pull/544.

mato:
Will look into this, need to verify the end to end build of Mirage/Solo5 works as expected.

mato:
Fixed by #63 thanks to @hannesm

PR #63, opened 2016-07-13
no need for libgcc.a (and any LDLIBS)

hannesm:
see #50 and #61 -- removes passing through LDLIBS as well...


mato:
Build-tested on Debian for both ukvm and virtio, merging. Thanks!

Issue #36, opened 2016-06-15
[tracking] Upstream Mirage/Solo5


mato:
This issue tracks the individual tasks required to upstream (or otherwise release) Mirage/Solo5, referring to dependent issues or PRs as appropriate:

Must have:
- [x] Remove dependencies on Mini-OS (#19, #24, djwillia/mirage-platform#2, djwillia/mirage#1).
- [x] Non-busy-looping scheduler (#33, djwillia/mirage-platform#5, djwillia/mirage-net-solo5#1).
- [x] Pass a command line to the unikernel. (#20, #41, djwillia/mirage-platform#6).
- [x] Rename `qemu` target to `virtio` in Mirage tool (djwillia/mirage#2).
- [x] Implement `mirage-bootvar-solo5` to parse bootvars from command line (djwillia/mirage-bootvar-solo5#1).
- [x] Freestanding builds of `zarith` and `gmp`. Currently required for plain HTTP due to mirage/mirage#543. (done in djwillia/opam-solo5)
- [x] Stable enough to pass sustained load tests (e.g. `siege` of HTTP and HTTPS unikernel). Issues found so far: #58, #60. Status: OK, provided the the GC is tuned appropriately.
- [x] Decide on git repository layout, ownership and process of upstreaming. See "preparation for and upstreaming of Mirage/Solo5" comment below.
- [x] Upstreamed!

Nice to have:
- [x] TLS stack working. Requires a Solo5 "port" of `mirage-entropy` and adapting the `nocrypto` build system to use it, plus sorting out the resulting OPAM dependencies (djwillia/mirage#3, djwillia/opam-solo5#4).
- Standalone test programs for Solo5 APIs (#37).

Out of scope / future release:
- Support for >1 network or block device.

mato:
### Preparation for and upstreaming of Mirage/Solo5

1. Decide on repository naming and ownership. We have a bunch of repos that are Solo5 specific. Should we move them into a `solo5/` org or into `mirage/`. Decision: move into `mirage/` since they are parallels to the existing `xen` and `unix` variant repositories that are already there.

  Overall picture of repository/opam/org relationships:

  ````
OPAM package            Git repository (current)          Git repository (upstream)

mirage                  djwillia/mirage#solo5             mirage/mirage (upstream via pr)
mirage-solo5            djwillia/mirage-platform#solo5    
                            >> djwillia/mirage-solo5          mirage/mirage-solo5
[...]

PR #64, opened 2016-07-19
Modular build for ukvm

djwillia:
This PR should be considered with the matching one in djwillia/mirage#solo5. (https://github.com/djwillia/mirage/pull/5)

Most of the changes here are due to moving out the tests from the `kernel` directory, because each test will now have its own `ukvm-bin` generated.  `ukvm-configure` is the script that is called to generate the Makefile to build `ukvm-bin`.


ukvm/ukvm-configure
```diff
@@ -0,0 +1,52 @@
...
+    echo "    UKVM_SRC is /path/to/ukvm"
+    echo "    MODULES can be any combination of: net blk gdb"
+    exit
+fi
+
+UKVM_SRC=`readlink -e $1`
+shift
+UKVM_MODULES=$@
+
+(
```
ricarkol: You can simplify this a lot by using make string operations, and possibly not even needing to have a generated Makefile.ukvm. Something like this:

`UKVM_MODULES=blk net
all:
        gcc $(addprefix -DUKVM_MODULE_,${UKVM_MODULES}) $(addprefix _build-ukvm/ukvm-,${UKVM_MODULES})`

ukvm/ukvm-core.c
```diff
@@ -479,7 +479,7 @@ static void setup_cpuid(int kvm, int vcpufd)
         err(1, "KVM_SET_CPUID2 failed");
 }
 
-
+#if 0
```
ricarkol: I think we will never ever use this functionality. You might as well delete it. If we want it back we have the commits history.

Makefile
```diff
@@ -104,7 +107,10 @@ opam-ukvm-install: solo5-kernel-ukvm.pc ukvm_target
 	cp ukvm/ukvm.h $(OPAM_UKVM_INCDIR)/ukvm.h
 	cp kernel/ukvm/solo5.o kernel/ukvm/solo5.lds $(OPAM_UKVM_LIBDIR)
 	mkdir -p $(OPAM_BINDIR)
-	cp ukvm/ukvm $(OPAM_BINDIR)
+#	cp ukvm/ukvm $(OPAM_BINDIR)
```
mato: Remove rather than comment out?

README.md
```diff
@@ -97,9 +94,14 @@ use to talk to the unikernel:
...
 
+A specialized monitor called `ukvm-bin` is generated as part of the
+build process, so the command line arguments may differ depending on
+the needs of the unikernel.  To see the arguments, run `ukvm-bin` with
+no arguments or `--help`.
+
 To launch the unikernel:
 
-    ukvm /dev/zero tap100 ./mir-stackv4.ukvm
+    ukvm-bin --net=tap100 ./mir-stackv4.ukvm
```
mato: Should be `./ukvm-bin` (unless you have `.` in `$PATH` which is generally considered a bad idea).

README.md
```diff
@@ -138,12 +140,35 @@ Refer to the
...
+You can debug the unikernel running in ukvm using gdb, but need to
+build a `ukvm-bin` with gdb support.  To do so, edit the Makefile
+(generated in the case of Mirage), and add `gdb` to the `UKVM_MODULES`
+make variable.  Then, running `make ukvm-bin` should build a version
+with gdb support.
 
-    sudo $(which ukvm) ~/disk.img tap100 mir-console.solo5-ukvm --gdb
+Start ukvm with the `--gdb` flag, like this:
+
+    ukvm-bin mir-console.solo5-ukvm --gdb
```
mato: `./ukvm-bin`

mato:
@djwillia, this looks good to me, good stuff! I've tested with the companion PR (djwillia/mirage#5) and everything works as expected.

One important piece that is missing is that we can no longer build only `virtio` or `ukvm` and the corresponding tests. This is important at least for @hannesm FreeBSD build. I've made some followup fixes for this and other cleanups to the Makefiles, at https://github.com/mato/solo5/commit/9ce1e32842552d1d007acc7278917815d7dff4f3.

In the interests of getting this into an initial OPAM release of Mirage/Solo5 (#36), I think we should merge this and the companion djwillia/mirage#5 as is, followed up by a PR for my changes above, then I can cut an initial release and submit `solo5-kernel-{ukvm,virtio}` to opam.ocaml.org.

Dan, please confirm is this is OK with you?

ukvm/ukvm-core.c
djwillia: @ricarkol do you mean the interrupt inject code?  perhaps that should be a different issue/PR?

ukvm/ukvm-configure
djwillia: @ricarkol @mato If I understand what you are saying @ricarkol, the `Makefile.ukvm` would be the same in all cases.  The only thing that would change is what `UKVM_MODULES` was set to before the `include Makefile.ukvm`.  Is that right?

ukvm/ukvm-core.c
ricarkol: Yeap, agree: let's have a another one just for that instead (cleanup), which doesn't have to be now.

ukvm/ukvm-configure
ricarkol: Yes, the suggestion is to have one Makefile.ukvm in all cases.

Ah, if you decide to use the example, you will have to change the names to UKVM_OBJECTS (and make the MODULES upper-case). Thanks @mato

djwillia:
Thanks for the comments, it's OK with me to go ahead and merge, put your Makefile changes in and cut the initial release!

mato:
@djwillia Thanks for confirming, merged!

PR #68, opened 2016-07-21
fix make virtio and make clean

hannesm:
the `test_hello` defines both UKVM_TARGETS and VIRTIO_TARGETS, but there's no Makefile.ukvm generated when virtio tests are run: ignore failing to load it (and provide a phony ukvm-clean)

### DIFF
--- a/tests/Makefile.tests
+++ b/tests/Makefile.tests
@@ -13,11 +13,11 @@ CFLAGS+=-I$(SOLO5_DIR)
 %.o: %.c $(HEADERS)
 	$(CC) $(CFLAGS) -c $< -o $@
 
-ifdef UKVM_TARGETS
 Makefile.ukvm: $(UKVM_SRC)/ukvm-configure
 	$(UKVM_SRC)/ukvm-configure $(UKVM_SRC) $(UKVM_MODULES)
 
-include Makefile.ukvm
+.PHONY: ukvm-clean
+-include Makefile.ukvm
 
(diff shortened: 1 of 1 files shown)


djwillia:
Thanks @hannesm 
Isn't `Makefile.ukvm` always generated with this change?

hannesm:
no. the makefile is called with the `virtio` target, thus the Makerules exist, but nobody depends on them!

tests/Makefile.tests
```diff
@@ -13,30 +13,25 @@ CFLAGS+=-I$(SOLO5_DIR)
 %.o: %.c $(HEADERS)
 	$(CC) $(CFLAGS) -c $< -o $@
 
-ifdef UKVM_TARGETS
 Makefile.ukvm: $(UKVM_SRC)/ukvm-configure
 	$(UKVM_SRC)/ukvm-configure $(UKVM_SRC) $(UKVM_MODULES)
 
-include Makefile.ukvm
+.PHONY: ukvm-clean
+-include Makefile.ukvm
```
djwillia: @hannesm is the `-include` directive only respected by certain versions of Make?  On my machine it seems to always build `Makefile.ukvm` so that it can include it?

hannesm: `-` at front always means "don't bail if unsuccessful", which means for me that it tries (and fails) to build `Makefile.ukvm` for inclusion, but continues with the rest of the process... I'm not a gmake expert, and strongly believe there are better solutions :)

djwillia: Ok, thanks, I understand now.  merging

PR #69, opened 2016-07-26
VIRTIO_PCI_QUEUE_PFN is a 32bit register

hannesm:
for (i = 0; i > 2; i++) will never be executed, intension is to run it twice

### DIFF
--- a/kernel/virtio/virtio.c
+++ b/kernel/virtio/virtio.c
@@ -610,7 +610,7 @@ void virtio_config_block(uint16_t base)
     outb(base + VIRTIO_PCI_STATUS, VIRTIO_PCI_STATUS_DRIVER_OK);
 
     outb(base + VIRTIO_PCI_QUEUE_SEL, 0);
-    outw(base + VIRTIO_PCI_QUEUE_PFN, (uint64_t)blkq.vring
+    outl(base + VIRTIO_PCI_QUEUE_PFN, (uint64_t)blkq.vring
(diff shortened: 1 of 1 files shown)


hannesm:
This lets the `_init` code work (more or less, the defaultqueue size is 1024 instead of 256) on FreeBSD/bhyve.

Nevertheless, on transmission or reception of a packet, I actually see the packet in tcpdump, but then get a vm exit:
```
vm exit[0]
reason          VMX
rip             0x000000000010f654
inst_length    3
status          0
exit_reason     2
qualification   0x0000000000000000
inst_type               0
inst_error              0
Abort
```

This is RX side using test_ping_serve, and rip points to the retq of kernel_wait (which is hlt;ret).

When dealing with a TX (mirage-skeleton/ping), I end up with a similar exit (packet was put on the wire, can be observed in the host system), where rip points inside of serial_putc.

According to configuration, the guest uses a single CPU... is it possible that solo5 runs into some race (and we need to lock virtio & console communication)?  A hello world, using only console (and sleep()ing) works fine.

hannesm:
some further investigation leads to `exit_reason` 2 being a triple fault...

hannesm:
@mato this PR is good to merge, if you could test it on Linux, that would be great!  It should improve the virtio protocol when using the virtio backend.

mato:
This LGTM insofar that it doesn't make things any worse than they already are (issue coming up shortly), so merging it. (After testing on Linux)

PR #75, opened 2016-08-17
Adding a blk test kernel for the solo5 interface

ricarkol:
This is a small test for the solo5 blk IO interface. It just writes some sectors and checks that the read content matches.

```
./ukvm-bin --disk=/home/opam/disk.img ./test_blk.ukvm

qemu-system-x86_64 -s -nographic -name foo -m 1024 -kernel test_blk.virtio -device virtio-net,netdev=n0 -netdev tap,id=n0,ifname=tap100,script=no,downscript=no -drive file=/home/opam/disk.img,if=virtio
```

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,4 +1,4 @@
-TESTDIRS=test_hello test_ping_serve
+TESTDIRS=test_hello test_ping_serve test_blk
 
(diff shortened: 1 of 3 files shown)


hannesm:
LGTM, please merge

PR #80, opened 2016-08-26
virtio: Increase max queue size, use single descriptor for RX/TX

ricarkol:
Changes:
1. the max queue size for rx and tx is now 4096 and the final value is not hardcoded.
2. the receive operation now gets one buffer descriptor at a time, instead of two (it was too complicated to have 2).
3. adds some asserts in some places to "assert" our assumptions.

Tested the mir-stackv4 and mirage static-website unikernels in:
1. qemu 2.1.2 on linux (with dhcp and without)
2. Google compute engine (with dhcp=true)
3. qemu 2.6 with this patch to increase the size of the queues (GCE must have this patch or something similar): https://lists.gnu.org/archive/html/qemu-devel/2015-11/msg06194.html

### DIFF
--- a/kernel/virtio/virtio.c
+++ b/kernel/virtio/virtio.c
@@ -91,8 +91,6 @@ struct pkt_buffer {
     uint8_t data[PKT_BUFFER_LEN];
     uint32_t len;
 };
-struct pkt_buffer xmit_bufs[128];
-struct pkt_buffer recv_bufs[128];
 
 #define VIRTIO_BLK_SECTOR_SIZE    512
 struct virtio_blk_hdr {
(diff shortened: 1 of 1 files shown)


kernel/virtio/virtio.c
```diff
@@ -841,6 +846,8 @@ int solo5_net_read_sync(uint8_t *data, int *n)
         return -1;
 
     assert(len <= *n);
+    assert(len <= PKT_BUFFER_LEN);
+    memset(data, 0, *n);
```
mato: This memset is unnecessary.

ricarkol:
Hi @mato , working on the tx side.

ricarkol:
Addressed @mato 's comments:
- buffer len should be 1526
- one buffer descriptor per tx operation

mato:
Thanks @ricarkol. Tested locally here and all seems fine.

hannesm:
still no luck on FreeBSD/bhyve:
```
vm exit[0]
reason          VMX
rip             0x000000000010d908
inst_length     3
status          0
exit_reason     2
qualification 0x0000000000000000
inst_type               0
inst_error              0
Abort
```

which still is a triple fault afaics (trying `test_ping_serve` from this repository)

mato:
Where is 0x10d908?

hannesm:
from objdump:

```
000000000010d907 <kernel_wait>:
  10d907:       f4                      hlt    
  10d908:       c3                      retq   
```

I uploaded my test_ping_serve.virtio to http://berlin.ccc.de/~hannes/test_ping_serve.virtio

PR #85, opened 2016-09-08
Kill all global variables

mato:
Rather than fighting the toolchain to understand why globals are getting allocated as COMMON, I've changed all global variables to `static`.

~~A casualty of this process is `test_ping_serve`, since that was relying on poking around in variables declared in `virtio.c`.~~

~~@djwillia @ricarkol Please confirm you're ok with removing `test_ping_serve` -- I think we discussed this and concluded it's not worth keeping in it's current form?~~

### DIFF
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -41,7 +41,6 @@ virtio/kernel.o \
 virtio/low_level.o \
 virtio/low_level_interrupts.o \
 virtio/mem.o \
-virtio/net.o \
 virtio/pci.o \
 virtio/serial.o \
 virtio/time.o \
--- a/kernel/interrupts.c
+++ b/kernel/interrupts.c
@@ -48,7 +48,7 @@ struct __attribute__((__packed__)) idtptr {
 extern void idt_load(uint64_t idtptr);
 
 /* The actual memory for the IDT is here */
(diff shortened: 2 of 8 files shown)


ricarkol:
@mato I agree, it has to be done. The only thing that will directly affect the development is that building a test_ping_serve unikernel takes ~10 secs, compared to stackv4 that takes around 2 mins here.

mato:
@ricarkol Agreed. I'm just looking at the options for having a basic "is data flowing" network test of some kind (not necessarily IP). There are also stacks like uip (https://github.com/adamdunkels/uip/tree/master/uip) which we could use.

djwillia:
@mato @ricarkol good riddance `test_ping_serve`!!!  (In case it wasn't clear, I'm in agreement with removing that test and making all globals static).

mato:
@djwillia @ricarkol I changed my mind on `test_ping_serve` :-)

Enclosed is a new, improved version. Still doesn't do ARP, but now handles broadcast, so either of the following will get you a reply:

````
$ ping -b -I tap100 10.0.0.255
$ sudo arping -i tap100 <MAC>
````
Also, this version works on `ukvm` as well as `virtio`...

PR #87, opened 2016-09-09
Use virtio ring definitions from virtio_ring.h

ricarkol:
All Virtio's vring related structs are defined in an attached `.h` file in the virtio documentation: http://docs.oasis-open.org/virtio/virtio/v1.0/csprd01/virtio-v1.0-csprd01.html "7 virtio_ring.h". This change adds that file and changes virtio.c to use those definitions instead.

@mato @djwillia

### DIFF
--- a/kernel/virtio/virtio.c
+++ b/kernel/virtio/virtio.c
@@ -17,6 +17,7 @@
  */
 
 #include "kernel.h"
+#include "virtio_ring.h"
 
 /* virtio config space layout */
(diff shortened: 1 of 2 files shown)


ricarkol:
Tested locally with the static_website, ping_serve_test and test_blk unikernels. Tested on GCE with the static_website unikernel (with dhcp).

mato:
LGTM (haven't reviewed the code line by line but I trust that you've tested it). I have some other patches for virtio which I've been working on today to dynamically allocate all the buffers using malloc/memalign as appropriate, will re-work these once this is merged.

Issue #82, opened 2016-08-31
[Tracking] Solo5 tasks for Mirage 3.0 release


mato:
#Solo5 tasks under consideration for the Mirage 3.0 release, with (rough priority) and @mentions for those currently working on a task:
- [x] (1) _CI improvements and end to end testing_ (@mato).
  - `mirage/mirage-skeleton` Travis CI should build Solo5 targets. (done)
  - end to end testing and improved standalone tests: #112, #113, #114, #115, #120
  - Related issues: mirage/mirage#584.
- [x] (1) _Stabilise virtio drivers_ (@ricarkol). 
  - Iterative cleanups so that we always have something working.
  - Aim to move implementation to v1.0 of the spec.
  - Related issues: #78.
- [x] (1) _Audit ukvm interfaces_.
  - `ukvm` hypercall/device interfaces need assertions about invariants, eg. for out of range pointers or buffer sizes passed in by the guest.
  - Related issues: #99, #117
- [x] (1) _Audit ukvm ELF loader_.
  - There are multiple issues with the ukvm ELF loader. See #126.
- [x] (2) _Clean up interrupt and trap handlers_.
  - Switch to a single .S source for interrupt and trap handlers, use gas macros where needed rather than bash-script driven code generation.
  - Add internal interfaces and infrastructure to register/unregister interrupt handlers.
  - Improve debugging value of trap handlers (~~backtraces~~ and other info?).
  - Related issues: #32.
- [x] (2) _FreeBSD support_ (@mato, @hannesm).
  - Priority bumped by the need to test virtio on Byhve.
  - See #61, #102 for details.
- [x] (2) _`ukvm` host feature detection_. 
  - Friendly error message if `ukvm` doesn't support the host CPU or KVM interface.
  - Related issues: #9.
- ~~[ ] (3) _Improved memory protection_.~~
  - Guard pages, verify W^X is used as appropriate, unmapped zero page.
- [x] (4) _Integrated `mkimage` tool_.
  - #110

mato:
@hannesm Should we consider supporting bhyve for Mirage 3.0? Depends on how much time you have from now until the (end-September) planned release.

hannesm:
@mato appreciated.  still needs someone to do the tedious cleanups of include files... I'll focus on opam signing in september, and can't promise much time on bhyve

mato:
[...]

Issue #94, opened 2016-09-12
virtio-net and FreeBSD bhyve


hannesm:
even after recent changes (#87 #92 #91), I cannot get the virtio-net to work on bhyve.  Running `test_ping_serve.virtio` results in the following output:
```
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
multiboot: Using memory: 0x100000 - 0x10000000
TSC frequency estimate is 2594415330 Hz
virtio_config: device_id=1, interrupt_line=5
Found virtio network device with MAC: 00 a0 98 cc 6a 96 
net queue size is 1024/1024
Hello, World
Serving ping on 10.0.0.2 / 10.0.0.255 / 255.255.255.255
With MAC: 00:a0:98:cc:6a:96 (no ARP!)
Received non-ping or unsupported packet, dropped
Received ping, sending reply
Received ping, sending reply
```

I do run `ping 10.0.0.255` in another console on the host system.

A `tcpdump -lnxxxi tap0` shows that solo5 is transmitting packets -- but they contain only `0` bytes - but I see an ICMP echo request also twice on the `tap0` interface!?

```
20:48:45.569970 IP 10.0.0.1 > 10.0.0.255: ICMP echo request, id 56643, seq 0, length 64
        0x0000:  ffff ffff ffff 00bd 588c de00 0800 4500
        0x0010:  0054 b723 0000 4001 ae86 0a00 0001 0a00
        0x0020:  00ff 0800 1edd dd43 0000 57d7 069d 0008
        0x0030:  b25f 0809 0a0b 0c0d 0e0f 1011 1213 1415
        0x0040:  1617 1819 1a1b 1c1d 1e1f 2021 2223 2425
        0x0050:  2627 2829 2a2b 2c2d 2e2f 3031 3233 3435
        0x0060:  3637
20:48:45.570021 IP 10.0.0.1 > 10.0.0.255: ICMP echo request, id 56643, seq 0, length 64
        0x0000:  ffff ffff ffff 00bd 588c de00 0800 4500
        0x0010:  0054 b723 0000 4001 ae86 0a00 0001 0a00
        0x0020:  00ff 0800 1edd dd43 0000 57d7 069d 0008
        0x0030:  b25f 0809 0a0b 0c0d 0e0f 1011 1213 1415
        0x0040:  1617 1819 1a1b 1c1d 1e1f 2021 2223 2425
[...]

PR #95, opened 2016-09-13
Refactor interrupt and trap handlers

mato:
Part of #82 ("Clean up interrupt and trap handlers").


mato:
@ricarkol @djwillia Please review. This is a fairly large change all over, but hopefully for the better (looking at LOC deleted vs. added). I've tested the rewritten trap handlers with manually written code triggering `#DE` and `#PF` and all seems fine. Likewise, the standalone tests pass muster with the new irq handlers.

ricarkol:
taking a look @mato

mato:
@hannesm Asking for the obligatory test on FreeBSD for this when you have a chance, just to ensure I haven't screwed anything up :-)

ricarkol:
lgtm

Tested with the static_website and stackv4 unikernels locally and in GCE. Also tested triggering various interrupts by hand with `int` at various places. Also tested ukvm test_ping_serve and test_blk just in case. All good.

hannesm:
@mato same as usual on FreeBSD/bhyve: interrupts are handled well, but all that is send are zero bytes (see #94).  Thanks (this code looks much cleaner than the previous one)!

PR #96, opened 2016-09-14
Improved assert() and PANIC()

mato:
See individual commits for details. Hopefully nothing controversial here, with the exception of the reduced printf() buffer size.

### DIFF
--- a/kernel/malloc.c
+++ b/kernel/malloc.c
@@ -2,7 +2,7 @@
 
 /* dlmalloc configuration */
 #undef WIN32
-#define ABORT PANIC("dlmalloc: aborting...")
+#define ABORT PANIC("Fatal error")
 #define HAVE_MMAP 0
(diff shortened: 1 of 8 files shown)


kernel/malloc.c
```diff
@@ -19,7 +19,7 @@
 #define LACKS_STDLIB_H
 #define LACKS_SCHED_H
 #define LACKS_TIME_H
-#define ABORT_ON_ASSERT_FAILURE 1
+#define ABORT_ON_ASSERT_FAILURE 0
```
ricarkol: Just wanted to check that this is not a typo

mato: Good catch. It's not a typo, but the DEBUG business in malloc() is slightly confusing. I've pushed some changes to make our position clearer, and fix compliation with DEBUG=1 should anyone need it.

PR #97, opened 2016-09-15
General refactoring/renaming

mato:
As promised, hopefully the last big rename/refactor for a while. No functional changes intended.


ricarkol:
Thanks @mato , changing `low_level_` to `platform_` makes a lot of sense.

PR #98, opened 2016-09-21
virtio: Refactoring and cleanup

ricarkol:
This PR is for two changes:

1. The network header in bhyve's virtio-net implementation is hard coded to use the extra `num_buffers` field (or `vrh_bufs` in freebsd code) even when the `VIRTIO_NET_F_MRG_RXBUF` feature is not negotiated. This proposed change uses the extended network header, asserts the host feature `VIRTIO_NET_F_MRG_RXBUF`, and negotiates it with the host (always). Here is the freebsd code with the hardcoded struct for reference: https://github.com/freebsd/freebsd/blob/master/usr.sbin/bhyve/pci_virtio_net.c#L120

2. The Tx path in bhyve's virtio-net assumes that all Tx'es will have a first descriptor with the header alone. Here is the code. https://github.com/freebsd/freebsd/blob/master/usr.sbin/bhyve/pci_virtio_net.c#L617 has an interesting comment, and https://github.com/freebsd/freebsd/blob/master/usr.sbin/bhyve/pci_virtio_net.c#L631 only copies data for tx after the first descriptor. The proposed change uses two descriptors per Tx operation. All I could find in the virtio 1.0 spec is this:

> "Most common is to begin the data with a header (containing little-endian ﬁelds) for the device to read, and postﬁx it with a status tailer for the device to write.".

Tested this changes with the static_website unikernel locally and in GCE. Also tested the ping_test_serve unikernel in freebsd (after applying @hannesm change to Makefile.common and following the instructions).

### DIFF
--- a/kernel/virtio/virtio_blk.c
+++ b/kernel/virtio/virtio_blk.c
@@ -21,8 +21,8 @@
 #include "virtio_pci.h"
 
 #define VIRTIO_BLK_ID_BYTES       20
-#define VIRTIO_BLK_T_IN           0
-#define VIRTIO_BLK_T_OUT          1
+#define VIRTIO_BLK_T_IN           0 /* read */
+#define VIRTIO_BLK_T_OUT          1 /* write */
 #define VIRTIO_BLK_T_SCSI_CMD     2
 #define VIRTIO_BLK_T_SCSI_CMD_OUT 3
 #define VIRTIO_BLK_T_FLUSH        4
@@ -44,176 +44,86 @@ struct virtio_blk_hdr {
     uint64_t sector;
 };
 
-struct virtio_blk_req {
-    struct virtio_blk_hdr hdr;
-    uint8_t data[VIRTIO_BLK_SECTOR_SIZE];
-
-    volatile uint8_t status;
-    volatile uint8_t hw_used;
-};
-
-#define VIRTQ_BLK_MAX_QUEUE_SIZE 8192
-static struct virtio_blk_req *blk_bufs;
-static uint8_t *blk_data;
 static struct virtq blkq;
 #define VIRTQ_BLK  0
 
 static uint16_t virtio_blk_pci_base; /* base in PCI config space */
 
 static int blk_configured;
-static uint32_t blk_next_avail;
-static uint32_t blk_last_used;
 
 static int handle_virtio_blk_interrupt(void *);
 
-/* WARNING: called in interrupt context */
-static void check_blk(void)
+/* Returns the index to the head of the buffers chain. */
+static uint16_t virtio_blk_op(uint32_t type,
+                              uint64_t sector,
+                              void *data, int len)
 {
-    volatile struct virtq_used_elem *e;
-    struct virtq_desc *desc;
-    int dbg = 0;
+    struct virtio_blk_hdr hdr;
+    struct io_buffer *head_buf, *data_buf, *status_buf;
+    uint16_t head = blkq.next_avail;
 
-    for (;;) {
-        uint16_t data_idx;
-        struct virtio_blk_req *req;
-
-        if ((blkq.used->idx % blkq.num) == blk_last_used)
-            break;
-
-        e = &(blkq.used->ring[blk_last_used % blkq.num]);
-        desc = &(blkq.desc[e->id]);
-        req = (struct virtio_blk_req *)desc->addr;
-
-        if (dbg)
-            printf("INTR: BLK: desc=0x%p: type=%d sector=%d\n",
-                   desc->addr, req->hdr.type, req->hdr.sector);
-
-        req->hw_used = 1;
-
-        data_idx = desc->next;
-        desc = &(blkq.desc[data_idx]);
-
-        if (dbg)
-            printf("INTR: BLK: desc=0x%p data=%08x...\n",
-                   desc->addr, *(uint32_t *)desc->addr);
-
-        data_idx = desc->next;
-        desc = &(blkq.desc[data_idx]);
-
-        if (dbg)
-            printf("INTR: BLK: desc=0x%p status=%d\n",
-                   desc->addr, *(uint8_t *)desc->addr);
-
-        if (dbg)
-            printf("REAP: 0x%p next_avail %d last_used %d\n",
-                   desc->addr, blk_next_avail, blk_last_used);
-
-        blk_last_used = (blk_last_used + 1) % blkq.num;
-    }
-}
-
-
-static struct virtio_blk_req *virtio_blk_op(uint32_t type,
-                                            uint64_t sector,
-                                            void *data, int len)
-{
-    struct virtq_desc *desc;
-    struct virtq_avail *avail;
-    struct virtio_blk_req *req;
-    int dbg = 0;
-
-    if (((blk_next_avail + 3) % blkq.num)
-        == ((blk_last_used * 3) % blkq.num)) {
-        printf("blk buffer full! next_avail:%d last_used:%d\n",
-               blk_next_avail, blk_last_used);
-        return NULL;
-    }
-
-    /* we perform a copy into the blk buffer to make reclaiming easy */
     assert(len <= VIRTIO_BLK_SECTOR_SIZE);
-    req = &blk_bufs[blk_next_avail/3];
 
-    if (req->hw_used) {
-        printf("blk buffer full! no unused buffers!\n");
-        return NULL;
-    }
+    head_buf = &blkq.bufs[head];
+    data_buf = &blkq.bufs[(head + 1) % blkq.num];
+    status_buf = &blkq.bufs[(head + 2) % blkq.num];
 
-    req->hdr.type = type;
-    req->hdr.ioprio = 0;
-    req->hdr.sector = sector;
-    req->hw_used = 0;
+    hdr.type = type;
+    hdr.ioprio = 0;
+    hdr.sector = sector;
 
-    if (type == VIRTIO_BLK_T_OUT)
-        memcpy(req->data, data, len);
+    /* The header buf */
+    memcpy(head_buf->data, &hdr, sizeof(struct virtio_blk_hdr));
+    head_buf->len = sizeof(struct virtio_blk_hdr);
+    head_buf->extra_flags = 0;
 
-    /* it seems like we need 3 descriptors for a single req */
+    /* The data buf */
+    if (type == VIRTIO_BLK_T_OUT) /* write */
+        memcpy(data_buf->data, data, len);
+    else
+        data_buf->extra_flags = VIRTQ_DESC_F_WRITE;
+    data_buf->len = len;
 
-    if (dbg)
-        atomic_printf("REQ BLK: 0x%p type=%d\n", req, req->hdr.type);
+    /* The status buf */
+    status_buf->len = sizeof(uint8_t);
+    status_buf->extra_flags = VIRTQ_DESC_F_WRITE;
 
+    assert(virtq_add_descriptor_chain(&blkq, head, 3) == 0);
 
-    /* the header */
-    desc = &(blkq.desc[blk_next_avail]);
-    desc->addr = (uint64_t)&req->hdr;
-    desc->len = sizeof(struct virtio_blk_hdr);
-    desc->next = (blk_next_avail + 1) % blkq.num;
-    desc->flags = VIRTQ_DESC_F_NEXT;
-
-    if (dbg)
-        atomic_printf("REQ BLK: hdr at 0x%x\n", desc->addr);
-
-    /* the data */
-    desc = &(blkq.desc[(blk_next_avail + 1) % blkq.num]);
-    desc->addr = (uint64_t)req->data;
-    desc->len = len;
-    desc->next = (blk_next_avail + 2) % blkq.num;
-    desc->flags = VIRTQ_DESC_F_NEXT;
-    if (type == VIRTIO_BLK_T_IN)
-        desc->flags |= VIRTQ_DESC_F_WRITE;
-
-    if (dbg)
-        atomic_printf("REQ BLK: data at 0x%x data=%08x...\n",
-                      desc->addr, *(uint32_t *)desc->addr);
-
-    /* the status */
-    desc = &(blkq.desc[(blk_next_avail + 2) % blkq.num]);
-    desc->addr = (uint64_t)&req->status;
-    desc->len = sizeof(uint8_t);
-    desc->next = 0;
-    desc->flags = VIRTQ_DESC_F_WRITE;
-
-    if (dbg)
-        atomic_printf("REQ BLK: status at 0x%x\n", desc->addr);
-
-    if (dbg)
-        atomic_printf("BLK: 0x%p next_avail %d last_used %d\n",
-                      desc->addr, blk_next_avail, blk_last_used);
-
-    avail = blkq.avail;
-    /* Memory barriers should be unnecessary with one processor */
-    blkq.avail->ring[avail->idx % blkq.num] = blk_next_avail;
-
-    avail->idx++;
-    blk_next_avail = (blk_next_avail + 3) % blkq.num;
     outw(virtio_blk_pci_base + VIRTIO_PCI_QUEUE_NOTIFY, VIRTQ_BLK);
 
-    return req;
+    return head;
 }
 
-#if 0
-static struct virtio_blk_req *virtio_blk_flush(uint64_t sector)
+/* Returns the status (0 is OK, -1 is not) */
+static int virtio_blk_op_sync(uint32_t type,
+                              uint64_t sector,
+                              void *data, int *len)
 {
-    return virtio_blk_op(VIRTIO_BLK_T_FLUSH, sector, data, len);
-}
-#endif
-static struct virtio_blk_req *virtio_blk_write(uint64_t sector,
-                                               void *data, int len)
-{
-    return virtio_blk_op(VIRTIO_BLK_T_OUT, sector, data, len);
-}
-static struct virtio_blk_req *virtio_blk_read(uint64_t sector, int len)
-{
-    return virtio_blk_op(VIRTIO_BLK_T_IN, sector, NULL, len);
+    uint16_t head;
+    struct io_buffer *head_buf, *data_buf, *status_buf;
+    uint8_t status;
+
+    head = virtio_blk_op(type, sector, data, *len);
+
+    head_buf = &blkq.bufs[head];
+    data_buf = &blkq.bufs[(head + 1) % blkq.num];
+    status_buf = &blkq.bufs[(head + 2) % blkq.num];
+
+    /* XXX need timeout or something, because this can hang... sync
+     * should probably go away anyway
+     */
+    while (!head_buf->completed)
+        ;
+
+    status = (*(uint8_t *)status_buf);
+    if (status != VIRTIO_BLK_S_OK)
+        return -1;
+
+    if (type == VIRTIO_BLK_T_IN) /* read */
+        memcpy(data, data_buf->data, *len);
+
+    return 0;
 }
 
 void virtio_config_block(uint16_t base, unsigned irq)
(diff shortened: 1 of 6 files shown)


hannesm:
will hopefully soon have time to test! thank you so much for digging into this :D

kernel/virtio/virtio_net.c
```diff
@@ -203,38 +203,40 @@ static void recv_setup(void)
 int virtio_net_xmit_packet(void *data, int len)
 {
     struct virtq_desc *desc;
-    struct virtq_avail *avail;
     int dbg = 0;
 
-    if (((xmit_next_avail + 1) % xmitq.num) ==
-        (xmit_last_used % xmitq.num)) {
+    if (((xmit_next_avail + 2) % xmitq.num) ==
+        ((xmit_last_used * 2) % xmitq.num)) {
```
mato: Shouldn't this be
````
if (((xmit_next_avail + 2) % xmitq.num) >=
        ((xmit_last_used % xmitq.num)) {
````
?
See `check_xmit()`.

Comment (ricarkol) [inline kernel/virtio/virtio_net.c]: That's because the `tx_buf`'s indexes used for data match `xmit_next_avail`. So, the tx_bufs look like this: empty, data, empty, data, empty, data. It should look better with the change you suggest below: getting an xmit_buf for the header as well.

Btw, my plan is to make the xmit_bufs dynamic as well (next PR). Just malloc that memory and free it on the interrupt side.

**Comment by hannesm** [comment]
My take is: implement workarounds in solo5 to get it working right now.  But please report the issues upstream (via https://bugs.freebsd.org/bugzilla/ or https://lists.freebsd.org/mailman/listinfo/freebsd-virtualization); and once they're fixed upsteam (and released), we can remove the workarounds in solo5.

Sorry, still at ICFP in Japan, thus little time and bad internet connectivity

File: kernel/virtio/virtio_net.c (hunk 1 above)
Comment (ricarkol) [inline kernel/virtio/virtio_net.c]: Not using this check anymore. The new version keeps track of free descriptors explicitly (in virtq.num_avail).

File: kernel/virtio/virtio_net.c [hunk 2]
```diff
@@ -203,38 +203,40 @@ static void recv_setup(void)
...
-    assert((len + sizeof(virtio_net_hdr)) <= PKT_BUFFER_LEN);
     memcpy(xmit_bufs[xmit_next_avail].data,
            &virtio_net_hdr, sizeof(virtio_net_hdr));
-    memcpy(xmit_bufs[xmit_next_avail].data + sizeof(virtio_net_hdr),
-           data, len);
-
     desc = &(xmitq.desc[xmit_next_avail]);
-    desc->addr = (uint64_t) xmit_bufs[xmit_next_avail].data;
-    desc->len = sizeof(virtio_net_hdr) + len;
+    desc->addr = (uint64_t)&virtio_net_hdr;
```
ricarkol: Done.

kernel/virtio/virtio_net.c
```diff
@@ -203,38 +203,40 @@ static void recv_setup(void)
...
+    desc->len = sizeof(virtio_net_hdr);
+    desc->next = xmit_next_avail + 1;
```
ricarkol: Moved all this descriptor handling code to virtio_ring.c, handling the wraparound there more carefully.

kernel/virtio/virtio_net.c
```diff
@@ -203,38 +203,40 @@ static void recv_setup(void)
...
+    desc->flags = VIRTQ_DESC_F_NEXT;
+
+    assert(len <= PKT_BUFFER_LEN);
+    memcpy(xmit_bufs[xmit_next_avail + 1].data, data, len);
```
ricarkol: Moved all this descriptor handling code to virtio_ring.c, handling the wraparound there more carefully.

kernel/virtio/virtio_net.c
```diff
@@ -203,38 +203,40 @@ static void recv_setup(void)
...
+    desc = &(xmitq.desc[xmit_next_avail + 1]);
```
ricarkol: Moved all this descriptor handling code to virtio_ring.c, handling the wraparound there more carefully.

kernel/virtio/virtio_net.c
```diff
@@ -203,38 +203,40 @@ static void recv_setup(void)
...
+    desc->addr = (uint64_t) xmit_bufs[xmit_next_avail + 1].data;
```
ricarkol: Moved all this descriptor handling code to virtio_ring.c, handling the wraparound there more carefully.

ricarkol:
@mato "A general problem with our virtio code is that we're not using the various indexes ..._last_used, ..._next_avail in a consistent fashion. " --> cleaned all this. Those comparisons between last_used adn next_avail were done to infer if they are free descriptors. It's much easier to keep track of free descriptors by hand (with a var), so added that.

Moved all that common code dealing with building descriptor chains and keeping track of free descriptors to virtio_ring.c.

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+/* WARNING: called in interrupt context */
+void virtq_handle_interrupt(struct virtq *vq)
+{
+    volatile struct virtq_used_elem *e;
+
+    for (;;) {
+        uint16_t desc_idx;
+        struct io_buffer *head_buf;
+
+        if ((vq->used->idx % vq->num) == vq->last_used)
```
mato: Is last_used free-running? Code in drivers uses (% num) when touching it, doesn't it need the same here? (Also see general comment coming up)

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/*
+ * Create a descriptor chain starting at index head, using vq->bufs also
+ * starting at index head.
+ * Make sure the vq-bufs are cleaned before using them again.
+ */
+int virtq_add_descriptor_chain(struct virtq *vq,
+                               uint16_t head,
+                               uint16_t num)
```
mato: Nit: Inconsistent types. You're using uint32_t for the descriptor indexes in vq.

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+{
+    struct virtq_desc *desc;
+    uint16_t i;
```
mato: Nit: Inconsistent types. As above.

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+
+        /* Overwrite buf->len with the number of bytes written by the device.
+         * This will be 0 for the tx/blk_write case. */
+        head_buf->len = e->len;
+        head_buf->completed = 1;
+        assert(e->len <= MAX_BUFFER_LEN);
+
+        vq->num_avail++;
+        while (vq->desc[desc_idx].flags & VIRTQ_DESC_F_NEXT) {
+            head_buf->completed = 1;
```
mato: Not sure what's the intent here. Do we want to set completed for non-head bufs? If not, then this should go away. If yes, then it's wrong.

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+        desc = &(vq->desc[i]);
+
+        assert(vq->bufs[i].len <= MAX_BUFFER_LEN);
+
+        /*
+         * The first field of a "struct io_buffer" is the "data" field, so in
+         * the interrupt handler we can just cast this pointer back into a
+         * 'struct io_buffer'.
+         */
+        assert((uint64_t) vq->bufs[i].data == (uint64_t) &vq->bufs[i]);
```
mato: These casts should be unnecessary.

mato:
Following up on the "monster" review I just submitted, _if_ I understand the code correctly then I'd prefer that `vq->last_used` and `vq->next_avail` are defined as _free-running_ as it simplifies the code in `virtio_ring.c`:

1. You should be able to do just `vq->next_avail += num` in `virtq_add_descriptor_chain()`.
2. Compare directly against `vq->used->idx`  in the interrupt handler followed by `vq->last_used++` at the end of the loop.

What do you think?

mato:
One more thing, any particular reason why we're now using a chain of 3 buffers per block write request instead of just a single buffer? The spec seems to suggest the latter, I don't particularly care either way since our writes are synchronous, just curious.

kernel/virtio/virtio_blk.c
```diff
@@ -44,176 +44,84 @@ struct virtio_blk_hdr {
...
+    /* XXX need timeout or something, because this can hang... sync
+     * should probably go away anyway
+     */
+    while (!head_buf->completed)
+        ;
 
-    return req;
-}
+    ret = (*(uint8_t *) status_buf) != VIRTIO_BLK_S_OK;
+    if (type == VIRTIO_BLK_T_OUT && ret == 0)
```
ricarkol: It felt because there is a bug in the test (https://github.com/Solo5/solo5/blob/master/tests/test_blk/test_blk.c#L22):

```
        if (sector_write[i] != '0' + i % 10) <== should be sector_read
            /* Check failed */
            return 1;
```

kernel/virtio/virtio_blk.c
```diff
@@ -44,176 +44,84 @@ struct virtio_blk_hdr {
...
 
-struct virtio_blk_req {
-    struct virtio_blk_hdr hdr;
-    uint8_t data[VIRTIO_BLK_SECTOR_SIZE];
-
-    volatile uint8_t status;
-    volatile uint8_t hw_used;
-};
-
 #define VIRTQ_BLK_MAX_QUEUE_SIZE 8192
```
ricarkol: removed

kernel/virtio/virtio_blk.c
```diff
@@ -44,176 +44,84 @@ struct virtio_blk_hdr {
...
-        data_idx = desc->next;
-        desc = &(blkq.desc[data_idx]);
+    /* The status buf */
+    status_buf->len = sizeof(uint8_t);
+    status_buf->extra_flags = VIRTQ_DESC_F_WRITE;
 
-        if (dbg)
-            printf("INTR: BLK: desc=0x%p status=%d\n",
-                   desc->addr, *(uint8_t *)desc->addr);
+    virtq_add_descriptor_chain(&blkq, head, 3);
```
ricarkol: added an assert

kernel/virtio/virtio_blk.c
```diff
@@ -44,176 +44,84 @@ struct virtio_blk_hdr {
...
-    outw(virtio_blk_pci_base + VIRTIO_PCI_QUEUE_NOTIFY, VIRTQ_BLK);
+    /* XXX need timeout or something, because this can hang... sync
+     * should probably go away anyway
+     */
+    while (!head_buf->completed)
+        ;
 
-    return req;
-}
+    ret = (*(uint8_t *) status_buf) != VIRTIO_BLK_S_OK;
```
ricarkol: changed

kernel/virtio/virtio_blk.c
ricarkol: Fixed

kernel/virtio/virtio_net.c
```diff
@@ -181,76 +81,60 @@ int handle_virtio_net_interrupt(void *arg __attribute__((unused)))
...
+    int r;
 
-    memcpy(xmit_bufs[xmit_next_avail].data,
-           &virtio_net_hdr, sizeof(virtio_net_hdr));
-    desc = &(xmitq.desc[xmit_next_avail]);
-    desc->addr = (uint64_t)&virtio_net_hdr;
-    desc->len = sizeof(virtio_net_hdr);
-    desc->next = xmit_next_avail + 1;
-    desc->flags = VIRTQ_DESC_F_NEXT;
+    head = xmitq.next_avail;
```
ricarkol: Changed

kernel/virtio/virtio_net.c
```diff
@@ -288,78 +172,68 @@ void virtio_config_network(uint16_t base, unsigned irq)
...
     /* Remove the virtio_net_hdr */
-    *size = buf->len - sizeof(virtio_net_hdr);
-    return buf->data + sizeof(virtio_net_hdr);
+    *size = buf->len - sizeof(struct virtio_net_hdr);
+    return buf->data + sizeof(struct virtio_net_hdr);
+}
+
+int recv_load_desc(void)
+{
+    memset(recvq.bufs[recvq.next_avail].data, 0, PKT_BUFFER_LEN);
```
ricarkol: Not needed, removed.

kernel/virtio/virtio_net.c
```diff
@@ -288,78 +172,68 @@ void virtio_config_network(uint16_t base, unsigned irq)
...
+    recvq.bufs[recvq.next_avail].len = PKT_BUFFER_LEN;
+    recvq.bufs[recvq.next_avail].extra_flags = VIRTQ_DESC_F_WRITE;
+    return virtq_add_descriptor_chain(&recvq, recvq.next_avail, 1);
 }
 
 void virtio_net_pkt_put(void)
 {
-    recv_load_desc();
+    assert(recv_load_desc() == 0);
```
ricarkol: removed and comment added

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+/*
+ * There is no official max queue size. But we've seen 4096, so let's use the
+ * double of that.
+ */
+#define VIRTQ_NET_MAX_QUEUE_SIZE 8192
+
+/* WARNING: called in interrupt context */
+void virtq_handle_interrupt(struct virtq *vq)
+{
+    volatile struct virtq_used_elem *e;
```
ricarkol: Was like that, so left it "just in case". Removed it now, as it's not really needed.

kernel/virtio/virtio_ring.c
```diff
@@ -0,0 +1,145 @@
...
+
+#include "kernel.h"
+#include "virtio_ring.h"
+#include "virtio_pci.h"
+
+/*
+ * There is no official max queue size. But we've seen 4096, so let's use the
+ * double of that.
+ */
+#define VIRTQ_NET_MAX_QUEUE_SIZE 8192
```
ricarkol: changed

kernel/virtio/virtio_ring.c
ricarkol: removed

kernel/virtio/virtio_ring.c
ricarkol: changed

kernel/virtio/virtio_ring.c
ricarkol: Casting the right side to a (uint8_t *) now

ricarkol:
Thanks for the review @mato 

The usage of last_used and next_avail is now consistent and decided to make them wrap at vq->num.  The only places where %num is used is when they are being read. The reason is that if they are left as free-running, then every read would have to %num and there are too many of those reads in virtio_net, virtio_blk, and virtio_ring (I'm scared of forgetting a %num).

mato:
@ricarkol This LGTM. If you don't have anything else you'd like to do as part of this PR, could you please rebase your branch against master and squash the intermediate commits (as I don't think these are worth keeping). I'll double-check once that's done and merge if everything checks out.

ricarkol:
@mato: rebased

ricarkol:
@mato Made `next_avail` and `last_used` free-running. Does that help with your freebsd experiments?

mato:
@ricarkol No difference in behaviour with your last change. I've done some printf() debugging:

1. Removed the "sending reply" message in `test_ping_serve` (since command line passing seems broken in grub-bhyve).
2. Print `#` when entering `handle_virtio_net_interrupt()`.
3. Print `$` when `isr_status & VIRTIO_PCI_ISR_HAS_INTR` is true.
4. Print `@` when `solo5_poll()` in `test_ping_serve` times out.

The result with `ping -f 10.0.0.255` on my FreeBSD box:
````
Serving ping on 10.0.0.2 / 10.0.0.255 / 255.255.255.255
With MAC: 00:a0:98:f5:bc:45 (no ARP!)
@@#$@#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$#$@@@@@@@@@@@@@@(...)
````
So, what seems to be happening is that we stop receiving `virtio_net` interrupts altogether. However, interrupts as such are still working fine (since we see the `@` being printed once a second).

I've also tried adding an out to `VIRTIO_PCI_QUEUE_NOTIFY` at the end of `virtio_net_recv_pkt_put()`, no luck there either.

mato:
I think I've found the cause of the problem. Take a look at this log: http://pastebin.com/Ly4aBUDK. The `next_avail` and `last_used` values come from `virtio_net_pkt_get()`.

What I think is happening:

1. bhyve is placing packets onto the recvq as fast as possible, in _batches of more than one packet_.
2. Our interrupt handler is updating `last_used` as fast as possible.
3. However, `test_ping_serve` is processing the buffers at a slower rate than that at which they're being added to the virtq.
4. Therefore: `last_used` eventually catches up with `next_avail` at which point the code becomes confused and thinks there are no more packets.

Not sure how to fix this, needs some thought.

ricarkol:
Hi @mato, I can reproduce the problem by adding 10ms sleeps at every iteration in test_ping_serve.

There seems to be two problems on the rx path:

1. When `(next_avail % num) == (last_used % num)` we can't distinguish between a ring with not pending packets (i.e. just after initialization) from a ring without any empty descriptor (i.e. what happens if our test_ping_serve is too slow). This can be "fixed" by removing the `% num` and making next_avail start at 0 (diff is here http://pastebin.com/BqPhe37g).

2. The second problem is that we can't recover after reaching a situation where there are no empty descriptors for the device to put receive buffers. It almost seems as if the device gives up and never tries again. This receive ring is a queue and if this queue gets filled up, is it the virtio behavior to buffer things elsewhere? I guess the answer is in the qemu virtio devices code.

mato:
@ricarkol With your fix for 1) above I get some of the ping traffic coming through and none of the device-side hangs that you describe. For example:
````
--- 10.0.0.255 ping statistics ---
58130 packets transmitted, 5056 packets received, 91.3% packet loss
round-trip min/avg/max/stddev = 0.290/63.275/70.207/10.644 ms
````
Regarding your question in 2): Bhyve will drop packets received from the tap interface if there is no space in the guest receive queue (https://github.com/freebsd/freebsd/blob/master/usr.sbin/bhyve/pci_virtio_net.c#L325). This is normal behaviour for network interfaces when queues fill up.

I've been through both our code and the bhyve-side code yet again and unfortunately can't see anything obviously wrong, so not sure what to try next.

ricarkol:
Tested locally with test_ping_serve, test_blk, and mirage static_website. Tested on GCE with static_website.

hannesm:
Tested (with @mato next to me) this virtio-net (using test_ping_serve) on a naive FreeBSD-12 and bhyve:

- `ping -f` leads to packet loss (>96%)
- `ping -f` to a FreeBSD-11 guest does not lead to packet loss (after setting sysctl `net.inet.icmp.icmplim=0`)
- `ping -i 0.001` no packet loss
- `ping -i 0.0009` packet loss (and delayed replies: 300ms instead of 0.x)

as long as we stay above 10 ms between packets, all is good... as soon as we go below, packets start dropping and getting huge delays

PR #99, opened 2016-09-23
ukvm: Guest interfaces audit and cleanup

mato:
The interesting parts of this change are in 84e98d3; refer to the individual commits for details.

Tested against Solo5 standalone tests and Mirage `static_website`. Manual testing performed to check that the bounds checking and checked overflow macros behave as intended (on GCC 4.x, 5.x and 6.x).

@hannesm Please test against clang. The macros in 2b3f0b9 are written to support it but untested.
@djwillia Please review and let me know if OK with the overall concept.

(Part of #82, "Audit ukvm interfaces").


ukvm/ukvm-blk.c
```diff
@@ -8,74 +8,80 @@
...
-        assert(ret == rd->len);
-        rd->ret = 0;
+    if (rd->sector >= blkinfo.num_sectors) {
+        rd->ret = -1;
+        return;
     }
+
+    ret = lseek(diskfd, blkinfo.sector_size * rd->sector, SEEK_SET);
+    assert(ret != (off_t)-1);
+    ret = read(diskfd, mem +  rd->data, rd->len);
```
djwillia: are we missing a GUEST_CHECK_PADDR for rd->data here?

mato: Good catch! Fixed and re-pushed, also with the change to fix the Travis build.

hannesm:
clang (3.8.0) is happy with these changes (though most are not tested since I can only build the virtio target)

mato:
Still LGTM after the weekend, so merging :-)

PR #100, opened 2016-10-13
Getting rid of virtio interrupt handling

ricarkol:
This change removes all the interrupt handling code from virtio-net and virtio-blk. It's mainly done for simplicity purposes. We can do this at the moment because our IO calls are synchronous, and we poll for receiving network packets anyway. Nothing was previously done on interrupt handling besides advancing our index of used descriptors (i.e. last_used). To know that the device used a descriptor (e.g. it put a packet on the ring), we can just check the ring's used->idx index that gets incremented by the device whenever it uses a descriptor chain.

Tests:
- test_ping_serve and test_blk locally on kvm
- static_website on GCE

Quick perf test (`ping -f` against test_ping_serve on kvm):
- current master: `round-trip min/avg/max/stddev = 0.094/0.285/2.296/0.111 ms`
- after change: `round-trip min/avg/max/stddev = 0.086/0.256/6.012/0.160 ms`

### DIFF
--- a/kernel/virtio/virtio_blk.c
+++ b/kernel/virtio/virtio_blk.c
@@ -51,7 +51,6 @@ static uint16_t virtio_blk_pci_base; /* base in PCI config space */
 
 static int blk_configured;
 
-static int handle_virtio_blk_interrupt(void *);
 
 /* Returns the index to the head of the buffers chain. */
 static uint16_t virtio_blk_op(uint32_t type,
@@ -142,7 +141,7 @@ static int virtio_blk_op_sync(uint32_t type,
     return 0;
 }
 
-void virtio_config_block(uint16_t base, unsigned irq)
+void virtio_config_block(uint16_t base, unsigned irq __attribute__((__unused__)))
 {
     uint8_t ready_for_init = VIRTIO_PCI_STATUS_ACK | VIRTIO_PCI_STATUS_DRIVER;
     uint32_t host_features, guest_features;
@@ -181,7 +180,6 @@ void virtio_config_block(uint16_t base, unsigned irq)
 
     virtio_blk_pci_base = base;
     blk_configured = 1;
-    intr_register_irq(irq, handle_virtio_blk_interrupt, NULL);
 
     /*
      * We don't need to get interrupts every time the device uses our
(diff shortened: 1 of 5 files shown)


kernel/virtio/virtio_blk.c
```diff
@@ -178,7 +200,6 @@ int handle_virtio_blk_interrupt(void *arg __attribute__((unused)))
     if (blk_configured) {
```
mato: Do we need to register a block interrupt handler at all? `virtio_config_block()` sets `VIRTQ_AVAIL_F_NO_INTERRUPT` so this handler should never get called.

kernel/virtio/virtio_net.c
```diff
@@ -71,8 +71,6 @@ int handle_virtio_net_interrupt(void *arg __attribute__((unused)))
     if (net_configured) {
         isr_status = inb(virtio_net_pci_base + VIRTIO_PCI_ISR);
         if (isr_status & VIRTIO_PCI_ISR_HAS_INTR) {
-            virtq_handle_interrupt(&xmitq);
```
mato: Add a comment here to make it clear that this interrupt is just to kick us out of any `solo5_poll()` that may be running?

kernel/virtio/virtio_ring.h
```diff
@@ -103,7 +103,7 @@ struct virtq_used_elem {
 
 struct virtq_used {
         le16 flags;
-        le16 idx;
+        volatile le16 idx;
```
mato: `volatile` is not very meaningful with today's compilers. Can we drop it here?

kernel/virtio/virtio_blk.c
ricarkol: Good point, removed completely.

kernel/virtio/virtio_net.c
ricarkol: added

ricarkol:
Merging

@mato: with LGTM you are saying I'm good to go right? if not, let me know and I can revert.

mato:
Yeah, that's fine

------- Original message -------

PR #102, opened 2016-10-14
FreeBSD build support

mato:
- Introduce a Makeconf/configure.sh combo invoked automatically by 'make'.
- Default for 'make' is to only build targets applicable to the build OS.
- "Compiler-provided" headers are now copied in configure.sh to include-host/
  for in-tree builds, and installed by OPAM to $(OPAM_INCDIR)/include/host.
- Only Linux/gcc and FreeVSD/clang are supported.


configure.sh
```diff
@@ -0,0 +1,81 @@
...
+        ;;
+    FreeBSD)
+        # On FreeBSD/clang we use -nostdlibinc which gives us access to the
+        # clang-provided headers for compiler instrinsics. We copy the rest
+        # (std*.h, float.h) from the host.
+        cc_is_clang || die "Only 'clang' is supported on FreeBSD"
+        INCDIR=/usr/include
+        SRCS_MACH="machine/_stdint.h machine/_types.h machine/endian.h \
+            machine/_limits.h"
+        SRCS_SYS="sys/_null.h sys/_stdint.h sys/_types.h sys/cdefs.h"
```
hannesm: nocrypto (or zarith or gmp or whatever other dependency) also needs `sys/endian.h` (sorry that my blog article is out of date)

hannesm:
Thanks for doing this work (and sorry that I've been too lazy to do it myself)

configure.sh
mato: thx. will check those, only tested building console so far...

mato: nolibc provides a sys/endian.h, and gmp seems to build fine with that provided I add the `CC=cc` change

hannesm:
I confirm that this works on my -CURRENT system!

PR #105, opened 2016-10-18
Print memory map on boot, add stack guard space

mato:
This is useful to be able to quickly eyeball that the memory layout
looks sane and will also aid in debugging addresses in traps / other
crashes.

STACK_GUARD_SIZE is a primitive way of ensuring that (at least) heap
allocation does not overflow into the stack. Ideally we'd want to check
the converse also (stack does not overflow into heap) but the
infrastructure for that is not there yet.

### DIFF
--- a/kernel/ukvm/kernel.c
+++ b/kernel/ukvm/kernel.c
@@ -31,7 +31,6 @@ void _start(struct ukvm_boot_info *bi)
     int ret;
 
     banner();
-    printf("mem_size=%lx, kernel_end=%lx\n", bi->mem_size, bi->kernel_end);
 
     gdt_init();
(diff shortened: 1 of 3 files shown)


mato:
@ricarkol, @djwillia Can you review the `STACK_GUARD_SIZE` / `heap_max` check in `sbrk()`? The idea is that if `heap_top` gets within `STACK_GUARD_SIZE` of the _current_ stack then heap allocations will fail. Obviously this won't work if someone calls `malloc()` from an interrupt handler but that's not intended to work anyway.

mato:
@ricarkol Yeah, there's stuff a bunch of stuff that can be unified, however in the case of kernel.c needs a bit more tweaking with the build system. So, one thing at a time...

PR #106, opened 2016-10-20
Use VAR=value to set implicit variables in Makefile

mato:
VAR?=value does not do what we thought it did.


hannesm:
this is fine, but do we actually need those bindings (`make` provides exactly the same ones, no need to re-bind)... apart from `objcopy` afaics (see also https://www.gnu.org/software/make/manual/make.html#Implicit-Variables which for some reason does not list `LD`)

mato:
I'd like to keep the bindings in `Makefile.common` just to make it obvious to the user that is the place to set them.

hannesm:
LGTM

PR #107, opened 2016-10-20
Prefix all messages with 'Solo5: '

mato:
- Prefix all messages with 'Solo5 ', allowing for easy eyeballing of
"who printed what" in logs.
- Print more / structured information during PCI enumeration and config.
- No functional change intended, but I had to revamp the data structures
passed around in the PCI code, mostly to get the bus/dev numbers out.


mato:
Tested on various targets, including GCE. Merging, nothing to see here :-)

PR #108, opened 2016-10-24
Test ping serve fix checksum

ricarkol:
There was an issue in the checksum function in test_ping_serve, where many packets had incorrect checksums and were reported by `ping -f` as missed.

Other changes are:
1. reverse the quiet/verbose behavior of test_ping_serve, now any arg passed means verbose mode.
2. there was an assert in virtio_net.c that can legitimately fail when the 16 bit indexes wrap over.

### DIFF
--- a/kernel/virtio/virtio_net.c
+++ b/kernel/virtio/virtio_net.c
@@ -103,8 +103,6 @@ int virtio_net_xmit_packet(void *data, int len)
     struct io_buffer *head_buf, *data_buf;
     int r;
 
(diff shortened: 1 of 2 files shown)


mato:
@ricarkol Good catch. Could you replace the checksum function implementation with that in https://tools.ietf.org/html/rfc1071 section 4.1? This also handles the case where an even number of bytes are passed to the function.

mato:
Thanks, LGTM

PR #110, opened 2016-10-27
Add 'solo5-mkimage' tool.

mato:
solo5-mkimage can be used to build MBR partitioned disk images
containing a unikernel. Includes support for GCE-style images (sparse
.tar.gz).

Only runs on Linux and/or in Docker, mainly due to lack of portable DOS
fdisk.


mato:
@ricarkol @sgrove Can you test this when you get a chance and confirm it works for you? 
@djwillia Ok to remove the existing solo5-build-iso.bash in favour of solo5-mkimage?

tools/mkimage/solo5-mkimage.sh
```diff
@@ -0,0 +1,195 @@
...
+    # directories (containing IMAGE and UNIKERNEL) to bind-mount from the host.
+    DESTDIR=$(dirname ${IMAGE}) 2>/dev/null
+    [ ! -d ${DESTDIR} ] && die "can't get parent directory of: ${IMAGE}"
+    SRCDIR=$(dirname ${UNIKERNEL})
+    [ ! -d ${SRCDIR} ] && die "can't get parent directory of: ${UNIKERNEL}"
+    DESTFILE=$(basename ${IMAGE})
+    SRCFILE=$(basename ${UNIKERNEL})
+
+    exec docker run --rm \
+        --tmpfs /tmp:rw,noexec,nosuid,size=2g \
```
ricarkol: Old versions of docker (e.g. mine) don't like this --tmpfs arg.

tools/mkimage/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM alpine:latest
```
ricarkol: Slight preference to use a specific alpine version. Just in case things change on the alpine side and something breaks mysteriously.

mato:
@ricarkol I've addressed all your comments. Shall I remove `solo5-build-iso.bash`? We can always get it back from history/integrate that functionality into `solo5-mkimage` if a use-case for ISO images is found.

ricarkol:
Thanks for the changes @mato

PR #112, opened 2016-11-01
Add 'solo5-run-virtio' tool for launching virtio tests

mato:
Intended for development, testing and CI, not as a general tool for
launching unikernels. Supports kvm, qemu and bhyve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,9 @@
 # About Solo5
 
 Solo5 is most useful as a "base layer" to run
-[MirageOS](https://mirage.io/) unikernels, either on KVM/QEMU or on a
-specialized "unikernel monitor" called `ukvm`.
+[MirageOS](https://mirage.io/) unikernels, either on various existing
+hypervisors (KVM/QEMU, Bhyve) or on a specialized "unikernel monitor" called
+`ukvm`.
 
 # About ukvm
 
@@ -37,7 +38,7 @@ Mirage/Solo5 supports two different targets to `mirage configure`:
 1. `ukvm`: A specialized "unikernel monitor" which runs on Linux
    (`x86_64`) and uses KVM directly via `/dev/kvm`.
 2. `virtio`: An `x86_64` system with `virtio` network and disk
-   interfaces. Use this target for QEMU/KVM, plain QEMU or other
+   interfaces. Use this target for QEMU/KVM, plain QEMU, Bhyve or other
    hypervisors (see below).
 
 Start with a fresh OPAM switch for the Mirage/Solo5 target you will be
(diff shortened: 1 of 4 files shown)


mato:
@hannesm I think this is what you had in mind with "bhyve test run script"? Goes to some lengths to emulate "normal process" behaviour on stdout, i.e. run and `^C` should just work and clean up correctly.

Still a work in progress. Ideally I'd like it to have the exact same options as `ukvm` but unfortunately that's not possible as FreeBSD `getopts` does not support long options?

hannesm:
yes, thanks. this looks good to me!

hannesm:
another thing might be supporting nmdm for console (see https://www.freebsd.org/doc/handbook/virtualization-host-bhyve.html#virtualization-bhyve-nmdm) - AFAICS this does not need any support in solo5, just a different parameter for `com1`.  Allows headless startups!

mato:
@hannesm The script already uses nmdm to get the "it's just stdio" behaviour :-)

tools/run/solo5-virtio-run.sh
```diff
@@ -0,0 +1,186 @@
...
+        hv_addargs -s 3:0,virtio-blk,${BLKIMG}
+    fi
+
+    hv_addargs ${VMNAME}
+
+    # Fake a console using nmdm(4). Open 'A' end first and keep it open, then
+    # set some sane TTY parameters and launch bhyve on the 'B' end.
+    # Bhyve and cat will respond to interactive SIGINT correctly.
+    # XXX Can occasionally leak /dev/nmdm devices, oh well ...
+    cat ${TTYA} &
```
mato: presumably you want the console output to go somewhere in any case? if not you can always `>/dev/null`

hannesm: only in a failure case, and then I'm happy to `tip /dev/nmdmXXB` manually...

mato: ah, so you want the `nmdm` device, but don't want the `cat`?

hannesm: yes, in the rc.d scenario this is exactly what I want :)

mato:
Made some more tweaks, command line passing now works on Bhyve (I'd forgotten about the need for "placeholder"). Updated the docs to point to this script for launching virtio unikernels. We are good to go, merging...

README.md
```diff
@@ -103,15 +109,18 @@ To launch the unikernel:
...
 
-**Running with qemu on Linux**
+**Running with KVM/QEMU on Linux, or Bhyve on FreeBSD**
 
 To launch the unikernel:
 
-    qemu-system-x86_64 -nographic -vga none -kernel ./mir-stackv4.virtio \
-        -device virtio-net,netdev=n0 \
-        -netdev tap,id=n0,ifname=tap100,script=no,downscript=no
+    solo5-run-virtio -n tap100 ./mir-stackv4.virtio
```
hannesm: should be `solo5-run-virtio.sh`

mato: OPAM installs it without the `.sh` extension

PR #116, opened 2016-11-04
we should be using uaddll_ instead of uaddl_ for uint64_t

djwillia:
I found this because of warnings when using clang instead of gcc.  I think that we need to deal with `uint64_t` in those macros, but am a little bit unsure at the moment because of the fact that some pointers passed are required to be 32-bit due to GUEST_PIO32_TO_PADDR(x)

@mato @ricarkol, can one of you take a look?  Thanks!

### DIFF
--- a/ukvm/ukvm-private.h
+++ b/ukvm/ukvm-private.h
@@ -65,11 +65,12 @@
  *
  * Note that (p), (l) and (sz) should be uint64_t or compatible types.
  */
-#define GUEST_CHECK_PADDR(p, l, sz) \
+#include <inttypes.h>
+#define GUEST_CHECK_PADDR(p, l, sz)                                     \
     {                                                                          \
         uint64_t __e;                                                          \
         if ((p >= l) || uaddll_overflow(p, sz, __e) || (__e >= l))              \
-            errx(1, "%s:%d: Invalid guest access: paddr=0x%llx sz=%lu",         \
+            errx(1, "%s:%d: Invalid guest access: paddr=0x%"PRIu64 " sz=%lu",         \
                     __FILE__, __LINE__, p, sz);                                \
     }
 
--- a/ukvm/ukvm-private.h
+++ b/ukvm/ukvm-private.h
@@ -7,8 +7,8 @@
 #define ALIGN_UP(_num, _align)    (_align * ((_num + _align - 1) / _align))
 
 /*
- * Define uaddl_overflow(a, b, result) for compilers without
- * __builtin_uaddl_overlow().  Based on "Catching Integer Overflows in C"
+ * Define uaddll_overflow(a, b, result) for compilers without
+ * __builtin_uaddll_overlow().  Based on "Catching Integer Overflows in C"
  * (https://www.fefe.de/intof.html).
  */
 #if defined(__GNUC__)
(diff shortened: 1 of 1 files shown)


mato:
What are the exact warnings you're seeing? `clang` is used on FreeBSD and it's not complaining there.

djwillia:
I see from the travis check and now compiling back on another machine that this is a `unsigned long long` vs `unsigned long` having different meanings on different systems.  The warnings I was getting were:

```
uhvf-blk.c:60:5: warning: incompatible pointer types passing 'uint64_t *' (aka 'unsigned long long *') to parameter of type 'unsigned long *' [-Wincompatible-pointer-types]
    GUEST_CHECK_PADDR(rd->data, GUEST_SIZE, rd->len);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./ukvm-private.h:72:25: note: expanded from macro 'GUEST_CHECK_PADDR'
        if ((p >= l) || uaddl_overflow(p, sz, __e) || (__e >= l))              \
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~
./ukvm-private.h:29:60: note: expanded from macro 'uaddl_overflow'
#define uaddl_overflow(a,b,r) __builtin_uaddl_overflow(a,b,&r)
                                                           ^~
uhvf-blk.c:60:23: warning: format specifies type 'unsigned long' but the argument has type 'uint64_t' (aka 'unsigned long long') [-Wformat]
    GUEST_CHECK_PADDR(rd->data, GUEST_SIZE, rd->len);
    ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./ukvm-private.h:74:41: note: expanded from macro 'GUEST_CHECK_PADDR'
                    __FILE__, __LINE__, p, sz);                                \
```

On FreeBSD, are you building the `ukvm` host side with `clang`?  If not, I'm not sure that file will be compiled... Or maybe I'm missing a compile flag that affects `long` vs `long long`?

djwillia:
Ah it looks like the `ukvm` host side is not being compiled on FreeBSD (which makes perfect sense, since it's not linux and won't have some of the headers/KVM stuff).  
https://github.com/Solo5/solo5/blob/master/configure.sh#L64

mato:
Yeah, I only tested those macros manually on clang. Can you try keeping the change from `uaddl` to `uadll`, *but* adding `<inttypes.h>` and changing the `errx()` invocation to:

    if ((p >= l) || uaddl_overflow(p, sz, __e) || (__e >= l))              \
            errx(1, "%s:%d: Invalid guest access: paddr=0x%" PRIx64 "sz=%" PRIu64,         \
                    __FILE__, __LINE__, p, sz);

djwillia:
@mato you read my mind :)  as ugly as those PRI things are, I think that is probably the right solution, as we will be checking some 64-bit addresses.  Your fix is the right one, I will fix mine (I mistakenly used PRIu64 in the wrong place and didn't alter the other...

mato:
@djwillia Yeah, the printf formatting for these things is generally horrible. I'm wondering if we should use `%zu` for the 2nd conversion (and update the comment in GUEST_CHECK_PADDR) since we only ever pass a `size_t` as `sz`.

mato:
Superseded by #124.

PR #119, opened 2016-11-10
test-blk is using byte offsets and virtio_blk not cleaning up IO bufs before using them

ricarkol:
This PR is for two issues:
- `test_blk` is using `solo5_blk_*_sync` incorrectly, as these use sector numbers and not byte offsets.
- `virtio_blk_op` is not cleaning up the IO bufs completely before reusing them. The consequence is that IO bufs can reuse the flags of previous IOs, which happens when wrapping over the ring.


mato:
LGTM. Thanks!

PR #120, opened 2016-11-10
Add a test suite

mato:
TL;DR: `sudo tests/setup-tests.sh && sudo tests/run-tests.sh`.

Finally, a test suite that runs end to end tests using the standalone test programs in `tests/`. See tests/README.md for usage.

The test harness is a rather gnarly shell script, this is the simplest I could get to that has the basic functionality we need. There are more interesting things that can be done with automated tests, but those will probably need a test harness written in something better than shell.

Please review. 

@hannesm There is a race condition _somewhere_ in the innards of `solo5-run-virtio.sh` for Bhyve which may cause random hard hangs (not sure why `timeout` is not doing the right thing in this case). Will debug in due course.

Part of #82, (1) "end to end CI".

### DIFF
--- a/tests/test_ping_serve/test_ping_serve.c
+++ b/tests/test_ping_serve/test_ping_serve.c
@@ -263,6 +263,8 @@ static void send_garp(void)
 
 static void ping_serve(int verbose)
 {
+    unsigned long received = 0;
+
     /* XXX this interface should really not return a string */
     char *smac = solo5_net_mac_str();
     for (int i = 0; i < HLEN_ETHER; i++) {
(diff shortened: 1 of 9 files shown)


tests/test_ping_serve/test_ping_serve.c
```diff
@@ -317,6 +319,10 @@ static void ping_serve(int verbose)
         if (solo5_net_write_sync(buf, len) == -1)
             puts("Write error\n");
 
+        received++;
+        if (received == 100000)
```
ricarkol: Can you add a message here saying that we are exiting because of this limit please.

It would be nice to pass this number as a parameter as I usually test virtio-net by letting this run for a while. Do you mind if i create a PR afterwards making this an arg? (unless you want to do it as part of this PR)

mato:
@ricarkol I've implemented argument passing to the test unikernels in 039f206, does this address your comments?

mato:
@ricarkol I've implemented argument passing to the test unikernels in 039f206, does this address your comments?

PR #121, opened 2016-11-10
ukvm: Clean up error messages

mato:
- Ensure all error messages go to stderr.
- Consistently use err()/errx()/warn()/warnx() everywhere.
- Slightly less terse error messages for users.
- Cosmetic cleanups for usage errors.
- Remove some more commented out/debug code.
- Remove "useless" success messages (CPUID, blk/net setup).

No functional change intended with the exception of the last point (silent operation).

@djwillia I think this addresses "ukvm host feature detection" in #82. #9 talks about a few other things:
- TSC stability: that's a guest issue, I may add some detection to Solo5 for this if needed
- guest page size support which is moot for the moment since 2MB pages are supported on all x86_64 processors)

### DIFF
--- a/ukvm/ukvm-blk.c
+++ b/ukvm/ukvm-blk.c
@@ -108,15 +108,11 @@ static int setup(int vcpufd, uint8_t *mem)
     /* set up virtual disk */
     diskfd = open(diskfile, O_RDWR);
     if (diskfd == -1)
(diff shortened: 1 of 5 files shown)


djwillia:
Looks good to me, I think as far as ukvm host feature detection, this is good (and I agree that the TSC thing is a guest issue.

PR #122, opened 2016-11-13
put pkg-config files into share/pkgconfig, rather than lib/pkgconfig

hannesm:


mato:
Closing due to different solution used

PR #124, opened 2016-11-21
Make no assumptions about type of uint64_t in GUEST_CHECK_PADDR 

mato:
GUEST_CHECK_PADDR relies on the incorrect assumption that (uint64_t) is the same type as (unsigned long) and uses __builtin_uaddl_overflow() for checked arithmetic. This does not hold for all compiler configurations.

### DIFF
--- a/ukvm/ukvm-private.h
+++ b/ukvm/ukvm-private.h
@@ -1,19 +1,17 @@
 #ifndef __UKVM_PRIVATE_H__
 #define __UKVM_PRIVATE_H__
 
+#include <inttypes.h>
+#include <err.h>
+
 #define GUEST_SIZE      0x20000000 /* 512 MBs */
 #define GUEST_PAGE_SIZE 0x200000   /* 2 MB pages in guest */
 
 #define ALIGN_UP(_num, _align)    (_align * ((_num + _align - 1) / _align))
 
-/*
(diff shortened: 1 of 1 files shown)


djwillia:
@mato It seems like I'm still getting some warnings because of `size_t` vs. `unsigned long long`.  E.g.,:
```
warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
    GUEST_CHECK_PADDR(wr->data, GUEST_SIZE, wr->len);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```
If I use the `%zu` for the `size_t`, as you suggested in https://github.com/Solo5/solo5/pull/116, the warnings go away.  Is there a reason we can't use `%zu`?
```
#define GUEST_CHECK_PADDR(p, l, sz) \
    {                                                                          \
        uint64_t __e;                                                          \
        if ((p >= l) || add_overflow(p, sz, __e) || (__e >= l))                \
            errx(1, "%s:%d: Invalid guest access: "                            \
                "paddr=0x%" PRIx64 "sz=%zu",                          \ 
                    __FILE__, __LINE__, p, sz);                                \
    }
```

Issue #117, opened 2016-11-07
GUEST_CHECK_PADDR width issues


djwillia:
On the host side, `ukvm` checks addresses passed into the guest to make sure that they don't exceed the guest memory.  The macro `GUEST_CHECK_PADDR` (https://github.com/Solo5/solo5/blob/master/ukvm/ukvm-private.h#L68) is supposed to do this.

There was a question about whether uint64_t pointers should be checked with `uaddll_overflow` or `uaddl_overflow`.   It seems that this is not entirely straightforward, as different compilers may have different builtins or macros to do this that treat `unsigned long` and `unsigned long long` differently. 

The start of the discussion between @mato and @djwillia was in this PR (https://github.com/Solo5/solo5/pull/116)

mato:
@djwillia Can you check if the version in #124 works for you without any compiler warnings on clang/osx? (But don't merge yet)

I tested with:
- Debian stable, GCC 4.9.2, uses macros
- Debian stretch, GCC 6.2.0, uses builtins
- Debian stretch, clang 3.8.1, uses builtins (manually set `UKVM_CC=clang` appropriately)
- Travis, GCC 4.something, uses macros

mato:
(commenting here rather than in the PR(s))

@djwillia Regarding the issue with `size_t` and using `%zu` for it, that's probably fine, however I have yet to find time to re-run my tests on the various compilers with it.

What I'm more worried about is that the change relies on using the generic `__builtin_add_overflow` rather than the more strongly typed `__builtin_uaddl_overflow`. Based on the [clang](http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins) and [gcc](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) documentation this _seems_ fine, but I'm not 100% sure.

mato:
I've updated #124 and manually tested it with the following test program:
````
#include <stdint.h>
#include <stddef.h>

#include "ukvm/ukvm-private.h"
#include <assert.h>

#define TEST_CHECK_PADDR(p, l, sz) \
    {                                                                          \
        uint64_t __e;                                                          \
        if ((p >= l) || add_overflow(p, sz, __e) || (__e >= l))                \
            warnx("%s:%d: Invalid guest access: "                              \
                    "paddr=0x%" PRIx64 ", sz=%zu",                             \
                    __FILE__, __LINE__, p, sz);                                \
    }

[...]

PR #127, opened 2016-11-23
ELF loader audit

### LINKED ISSUE #126: ukvm: Multiple issues with ELF loader
While reviewing #123 and trying to write a test to go into `tests/` for #73 I found (at least) the following issues with the ELF loader in ukvm:

1. In `load_code()` the test for a loadable segment is incorrect, `phdr.p_type` is not a bitmap.
2. Return values from `mprotect()` are not checked.
3. The `mprotect()` call which is supposed to write-protect `.text` is passed a `len` which is not a multiple of the page size.
4. (related) Trying to write to a global `char *` in the guest produces an `EFAULT` exit from `KVM_RUN` where it should succeed.

I'm working on auditing that code and fixing the above.

mato:
Fixes #126. See individual commits for details.

@djwillia Please review.

### DIFF
--- a/ukvm/ukvm-core.c
+++ b/ukvm/ukvm-core.c
@@ -467,9 +467,19 @@ static int vcpu_loop(struct kvm_run *run, int vcpufd, uint8_t *mem)
         int i, handled = 0;
 
         ret = ioctl(vcpufd, KVM_RUN, NULL);
-        if (ret < 0 &&
-                (errno != EINTR && errno != EAGAIN)) {
-            err(1, "KVM: ioctl (RUN) failed");
+        if (ret == -1 && errno == EINTR)
+            continue;
+        if (ret == -1) {
+            if (errno == EFAULT) {
+                struct kvm_regs regs;
+                ret = ioctl(vcpufd, KVM_GET_REGS, &regs);
+                if (ret == -1)
+                    err(1, "KVM: ioctl (GET_REGS) failed after guest fault");
+                errx(1, "KVM: host/guest translation fault: rip=0x%llx",
+                        regs.rip);
+            }
+            else
+                err(1, "KVM: ioctl (RUN) failed");
         }
 
         for (i = 0; i < NUM_MODULES; i++) {
--- a/ukvm/ukvm-core.c
+++ b/ukvm/ukvm-core.c
@@ -248,7 +248,7 @@ static void load_code(const char *file, uint8_t *mem,     /* IN */
             errx(1, "%s: Invalid segment: paddr=0x%" PRIx64 ", memsz=%zu",
                     file, paddr, memsz);
         }
-        _end = ALIGN_UP(paddr + memsz, align);
+        _end = ALIGN_UP(result, align);
         if (_end >= GUEST_SIZE) {
             errx(1, "%s: Invalid segment: paddr=0x%" PRIx64 \
                     ", _end=0x%" PRIx64, file, paddr, _end);
--- a/ukvm/ukvm-core.c
+++ b/ukvm/ukvm-core.c
@@ -154,11 +154,11 @@ ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
  *
  * Memory will look like this after the elf is loaded:
  *
- * *mem                    *p_entry                                 *p_end
- *   |             |                    |                |            |
- *   |    ...      | .text .rodata      |   .data .bss   |            |
- *   |             |        code        |   00000000000  | empty page |
- *   |             |  [PROT_EXEC|READ]  |                | PROT_NONE  |
+ * *mem                    *p_entry                   *p_end
+ *   |             |                    |                |
+ *   |    ...      | .text .rodata      |   .data .bss   |
+ *   |             |        code        |   00000000000  |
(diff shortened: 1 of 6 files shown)


ukvm/ukvm-core.c
```diff
@@ -224,45 +224,55 @@ static void load_code(const char *file, uint8_t *mem,     /* IN */
...
-        /* Protect the executable code */
-        if (phdr[ph_i].p_flags & PF_X)
-            mprotect((void *) dst, memsz, PROT_EXEC | PROT_READ);
 
+        if ((paddr >= GUEST_SIZE) || add_overflow(paddr, filesz, result) ||
+            (result >= GUEST_SIZE)) {
+            errx(1, "%s: Invalid segment: paddr=0x%" PRIx64 ", filesz=%zu",
+                    file, paddr, filesz);
+        }
+        if (add_overflow(paddr, memsz, result) || (result >= GUEST_SIZE)) {
```
ricarkol: I'm not 100% sure this can't overflow after the ALIGN_UP below. It doesn't seem to because (but wanted to double check with you): if `(paddr_memsz <= 1e64)` then `ALIGN_UP(paddr+memsz, 4096) <= 1e64`. And that will be true if paddr and memsz are both `uint64_t` so `add_overflow` becomes `UaddLL_overflow`, and if the check is `<=` and not `<`. Wouldn't it be safer to make memsz an uint64_t just to be sure the correct add_overflow is used?

`(result >= GUEST_SIZE)` is not needed here as that check is done one line below.

mato: >  if (paddr_memsz <= 2^64) then ALIGN_UP(paddr+memsz, 4096) <= 2^64
I don't follow. `align` comes from the ELF file, so we can't make any assumptions about its value.

Regarding using an `uint64_t` for `memsz`, that shouldn't matter as `paddr` and `result` are already both `uint64_t`, so the result will be computed using the correct types (see #117, #126) If I've got that behaviour wrong then #126 is also wrong :-/.

The `(result -> GUEST_SIZE)` check is just to trigger the correct error message about `memsz` being invalid rather than `_end`. I can simplify a bit more and do this (re-use `result`):

````
    if ((paddr >= GUEST_SIZE) || add_overflow(paddr, filesz, result) ||
        (result >= GUEST_SIZE)) {
        /* ... complain about filesz */
    }
    if (add_overflow(paddr, memsz, result) || (result >= GUEST_SIZE)) {
        /* ... complain about memsz */
    }
    _end = ALIGN_UP(result, align);
    if (_end >= GUEST_SIZE) {
    /* ... complain about _end */
    }
````

What do you think?

Comment (ricarkol) [inline ukvm/ukvm-core.c]: Looks good, mainly wanted to double check and have a full understanding of it. All makes sense, thanks!

Comment (ricarkol) [inline ukvm/ukvm-core.c]: `if (paddr_memsz <= 2^64) then ALIGN_UP(paddr+memsz, ALIGN) <= 2^64` is true as long as 2^64 is a multiple of ALIGN. Which I'm guess is the case, as ALIGN seems to always be 4096.

Comment (mato) [inline ukvm/ukvm-core.c]: That's true, but the point of these checks is that the data in the ELF file is supplied by the guest, so could be crafted to be malicious. So a little extra paranoia can't hurt.

Comment (mato) [inline ukvm/ukvm-core.c]: See f47d93b, that replaces ALIGN_UP with a bitwise operation and addresses the case where it's not a power of 2 by just ignoring the value.

**Comment by ricarkol** [comment]
Hi Mato, there's still a problem. `2**64` is still overflowing: max_value(uint64_t) is 0xffffffffffffffff and not 0x1000000000000), so even if align is a power of two, we will overflow. For example, if align=4096 and `paddr+memsz=0xffffffffffffffff-100`, then result will be 0.

```#include <stdio.h>

int main()
{
        unsigned long long align = 4096;
        unsigned long long result = 0xffffffffffffffff - 100;
        printf("%llx\n", (result + (align - 1)) & -align);
}
```
===> outputs 0

mato:
(copying comments from Slack, for the record)

I realise the `﻿⁠⁠⁠⁠(x + (a -1 )) & -a)`﻿⁠⁠⁠⁠ computation can overflow, however if it overflows into anything _﻿⁠⁠dangerous﻿⁠_⁠ it'll get caught by the check here: https://github.com/Solo5/solo5/blob/master/ukvm/ukvm-core.c#L260.

If the final value of `_end` is bogus but not dangerous, we don't care (garbage in, garbage out!).

mato:
(After further discussion.) There is still an edge case here. The guest could potentially supply a value of ﻿⁠⁠⁠⁠align﻿⁠⁠⁠⁠ that would overflow in such a way that (for example) ﻿⁠⁠⁠⁠_end == paddr﻿⁠⁠⁠⁠ rendering the following call to ﻿⁠⁠⁠⁠mprotect()﻿⁠⁠⁠⁠ useless. So the code really needs to use `add_overflow()` for the alignment computation as well.

mato:
Reopened #126 so that I don't forget to address this last point.

mato:
@ricarkol Addressed in #129, take a look.

PR #130, opened 2016-11-24
Ukvm cleanups

mato:
See individual commits for details. Nothing controversial here.

### DIFF
--- a/ukvm/ukvm-core.c
+++ b/ukvm/ukvm-core.c
@@ -457,10 +457,10 @@ static int vcpu_loop(struct kvm_run *run, int vcpufd, uint8_t *mem)
             continue;
 
         switch (run->exit_reason) {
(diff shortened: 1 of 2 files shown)


ukvm/ukvm-core.c
```diff
@@ -382,48 +381,6 @@ static void setup_cpuid(int kvm, int vcpufd)
...
-     * We are using a TSC-based clock, but this is an example for
-     * delivering a timer interrupt once a second.
-     */
-    if (0) {
-        for (;;) {
-            sleep(1);		/* every 1 s */
-
-            pthread_mutex_lock(&interrupt_mutex);
-
-            run->request_interrupt_window = 1;
```
ricarkol: This was the old primitive way of having a "tick" interrupt, definitely not needed anymore.

PR #131, opened 2016-11-24
ukvm-configure: Correct dependencies in generated Makefile

mato:
1. Make _build_ukvm/ an order-only prerequisite, thus not rebuilding
bits of ukvm every time due to the timestamp on the directory changing.

2. Add $(MAKEFILE_LIST) to ukvm dependencies, forcing a rebuild of ukvm
if the generated Makefile has changed (e.g. from re-running of
ukvm-configure).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ include Makefile.common
 .PHONY: all
 all: ukvm virtio
 .DEFAULT_GOAL := all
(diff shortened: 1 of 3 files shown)


mato:
Added a couple more things, all with the goal of reproducible behaviour with make and parallel builds.

mato:
Nothing controversial here, merging this one

PR #137, opened 2016-11-29
use _build/ukvm instead of _build-ukvm for building

hannesm:
part of https://github.com/mirage/mirage/pull/703


mato:
No longer required

PR #138, opened 2016-12-06
Fix bug in non-builtin add_overflow() when (b < 1)

mato:
Commit d0cf32cd5e9d7959c5f66800df1a821b8db68296 introduced the generic
add_overflow() macro for compilers without checked arithmetic builtins.
Unfortunately it also introduced a copy-paste-error for the case where
(b < 1).

This showed up as an "Invalid guest access" in ukvm_port_puts() with a
p->len of 0. This change fixes the typo and the invalid guest access no
longer triggers.


mato:
@djwillia FYI - you definitely want this fix if you're using the macros elsewhere

PR #143, opened 2016-12-09
Optionally accept a pre-existing open tap device fd instead of a name

ijc25:
This allows a more privileged parent process to take care of opening
/dev/net/tun etc and simply pass the open fd to the ukvm child, meaning ukvm
does not now require access to the tun device.

https://github.com/mato/docker-unikernel-runner/pull/4 makes use of this new functionality.


mato:
Looks good. Thanks!

Issue #141, opened 2016-12-08
virtio: bhyve limits vmname to 8 characters


hannesm:
"mir-nqsbio" doesn't work with bhyve flawlessly (affects `tools/run/solo5-run-virtio.sh`)

we should crop the vmname if the given thing is too large.

we may want to remove "mir-" prefix from all mirage generated binaries (to be fixed separately in mirage tool)...

mato:
An 8 character limit is a bit daft. Perhaps we could just use `vm<PID>` for the purposes of `solo5-run-virtio`?

hannesm:
mirage now does not add a `mir-` prefix anymore, see https://github.com/mirage/mirage/pull/725

maybe the PID approach is good for `solo5-run-virtio` if the kernel name (without `.virtio`) is longer than 8 characters.  another option would be to sub the first 8 characters and hope there is no collision.

mato:
Fixed in #145.

PR #145, opened 2016-12-12
FreeBSD/bhyve: Use vm<PID> for VM name, VM cleanup change

mato:
- Address #141 by using VM<PID> for the bhyve VM name.
- Rather than blowing away an existing bhyve VM of the same name each time we
run, move the cleanup to be done in solo5-run-virtio.sh SIGINT/SIGTERM processing.


mato:
@hannesm Does this work for you? Using `vm<PID>` for the name always is easiest, it's also consistent with the `/dev/nmdmXXX` naming.

hannesm:
LGTM

Issue #144, opened 2016-12-10
tsc time source broken on bhyve


hannesm:
I get some monotonic time in the right interval, but it is not gettimeofday (`2016-12-12T21:52:17-00:00` in VM vs `Sat 10 Dec 2016 22:09:30 CET` on host).

`kernel/virtio/tscclock.c` is in use.

while investigating, I found that the code getting the time still uses the OCaml runtime and its float conversion, we should use a timeval in mirage-clock-freestanding and get rid of the floating point conversions.

mato:
I can reproduce this on `bhyve`, but not on software-only `qemu` (which uses the same TSC clock code as bhyve). Will investigate some more and see what I can find, my guess is we might be tickling something in bhyve's RTC emulation which is giving us wrong values for wall time.

`mirage-clock-freestanding` using float conversions is a separate issue, should be reported there.

mato:
I've found what looks like the root cause of the problem. The `tscclock` RTC code is correct, however on KVM and QEMU (irrespective of whether or not the RTC is involved) the TSC begins counting at zero when the VM is initialised. On Bhyve the TSC appears to be either passed through directly from the host or otherwise initialised to an unspecified base value.

The Solo5 clock code uses the TSC base value as the base for monotonic time (see https://github.com/Solo5/solo5/blob/master/kernel/virtio/tscclock.c#L203 for the `tscclock` case) and assumes that this starts at zero at boot time. The wall clock value value is only read _once_, then used as an offset against monotonic time. Unfortunately this assumption is violated in the bhyve case so the wall clock TOD is incorrect.

mato:
Fixed in #146.

PR #146, opened 2016-12-13
tscclock: Correctly compute RTC epoch offset

mato:
The tscclock code assumed that the RDTSC reading used to compute
time_base at boot time is zero (or close enough). This does not
necessarily hold (see #144 for issues on Bhyve, but can affect QEMU/KVM
and other virtio hypervisors as well).

Correct the computation to

    rtc_epochoffset = rtc_boot - time_base

@djwillia, @ricarkol Please review. This only affects virtio targets without `pvclock`.

### DIFF
--- a/kernel/virtio/tscclock.c
+++ b/kernel/virtio/tscclock.c
@@ -169,7 +169,7 @@ uint64_t tscclock_monotonic(void) {
  * Calibrate TSC and initialise TSC clock.
  */
 int tscclock_init(void) {
-    uint64_t tsc_freq;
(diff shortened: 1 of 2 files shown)


mato:
@hannesm Could you please test on your hardware w/ bhyve and see if this produces the expected result? Note the change to ask bhyve to keep the RTC in UTC in 80057bf.

hannesm:
works as advertised! thanks!

djwillia:
Looks good to me

PR #123, opened 2016-11-18
remove LD hacks

hannesm:
Makefile.common included first Makeconf, and set AS, CC, LD unconditionally
afterwards.  This lead to the LD in Makeconf never having any effect.

It turns out (I didn't yet dig which commit) that FreeBSD's ld (version 2.17.50)
works fine with Solo5 now.  I tested this with a) solo5 examples,
b) mirage-skeleton examples up to stackv4 and static_website_tls.

This commit a) removes the LD detection (executed on FreeBSD only), b) moves the .include in Makefile.common to a later point so that we can actually override these variables in Makeconf.  It worked before #106, but not after (before `?=` was used, after `=`).

@mato a review would be appreciated.  This complements https://github.com/mirage/mirage/pull/692 pretty well.

Here is `readelf -S mir-stackv4.virtio` compiled with ld 2.17.50:
```
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0] (null)            NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .bootstrap        PROGBITS         0000000000100000  00001000
       0000000000000020  0000000000000000  AX       0     0     4
  [ 2] .text             PROGBITS         0000000000101000  00002000
       00000000001116cf  0000000000000000  AX       0     0     4096
  [ 3] .rodata           PROGBITS         0000000000213000  00114000
       00000000000076a8  0000000000000000   A       0     0     16
  [ 4] .eh_frame         X86_64_UNWIND    000000000021a6a8  0011b6a8
       00000000000336b0  0000000000000000   A       0     0     8
  [ 5] .got              PROGBITS         000000000024e000  0014f000
       000000000000e718  0000000000000008  WA       0     0     8
  [ 6] .data             PROGBITS         000000000025c740  0015d740
       00000000000ba440  0000000000000000  WA       0     0     64
  [ 7] .bss              NOBITS           0000000000317000  00217b80
       0000000000019694  0000000000000000  WA       0     0     16
  [ 8] .comment          PROGBITS         0000000000000000  00217b80
       0000000000000052  0000000000000001  MS       0     0     1
  [ 9] .debug_aranges    PROGBITS         0000000000000000  00217be0
       0000000000001890  0000000000000000           0     0     16
  [10] .debug_pubnames   PROGBITS         0000000000000000  00219470
       0000000000007501  0000000000000000           0     0     1
[...]

PR #147, opened 2016-12-14
remove unnecessary depext to binutils on FreeBSD since #123 is merged

hannesm:


hannesm:
@mato this should IMHO be a no-brainer

PR #148, opened 2016-12-19
Support user-provided mac address via --mac= CLI

ijc25:


ukvm/ukvm-net.c
```diff
@@ -177,10 +179,16 @@ static int handle_exit(struct kvm_run *run, int vcpufd, uint8_t *mem)
 
 static int handle_cmdarg(char *cmdarg)
 {
-    if (strncmp("--net=", cmdarg, 6))
+    if (!strncmp("--net=", cmdarg, 6)) {
+        netiface = cmdarg + 6;
+        return 0;
+    } else if (!strncmp("--mac=", cmdarg, 6)) {
+        snprintf(netinfo.mac_str, sizeof(netinfo.mac_str), cmdarg + 6);
```
mato: This should pass `%s` as the format string to `snprintf()` to prevent format-string-injection bugs or use `strncpy()` / `strlcpy()`.

ukvm/ukvm-net.c
```diff
@@ -177,10 +179,16 @@ static int handle_exit(struct kvm_run *run, int vcpufd, uint8_t *mem)
 
 static int handle_cmdarg(char *cmdarg)
 {
-    if (strncmp("--net=", cmdarg, 6))
+    if (!strncmp("--net=", cmdarg, 6)) {
+        netiface = cmdarg + 6;
+        return 0;
+    } else if (!strncmp("--mac=", cmdarg, 6)) {
```
mato: Can we call the option `--net-mac`? This makes it clear which module it belongs to.

ukvm/ukvm-net.c
mato: It'd be good to check that the argument to this option is something that at least looks like a MAC address. Perhaps using `sscanf()` and checking that the correct number of items were scanned?

ijc25:
Your two additions (help and switch from `err` to `warnx`) LGTM, thanks.

mato:
Looks good, merging. Thanks!

PR #149, opened 2016-12-21
nolibc migration for tests (fixes #37)

### LINKED ISSUE #37: Standalone test programs for Solo5 APIs
Currently the standalone test programs are limited to "Hello World" and "ping" (virtio only). We should have standalone tests for all of the Solo5 APIs, for both ukvm and virtio. This will require integrating nolibc from ocaml-freestanding to build the tests as post-namespacing Solo5 does not export any libc-like functionality.

DrDub:
This is  modified copy of the freestanding nolibc,
per my understanding of the discussion in the bug
report.

What has been modified is the removal of the fmt_fp
function of vfprintf.c (with their corresponding flags
e, f, g and ) and the dependency on openlibm (which
means <math.h> is a noop).

The provided test cases tests solo5 memory allocation
by allocating various strings, growing them and deleting
them non-consecutively, with an eye on memory corruption. 
I believe that was the main solo5 API left untested.

If this is in the right direction, I can expand this PR to remove 
the code clones from nolibc in the other test cases, although 
it might be better to restructure the test harness to link to this, 
which might be better left to core contributors. Just let me know.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,4 +1,4 @@
-TESTDIRS=test_hello test_globals test_ping_serve test_blk
+TESTDIRS=test_hello test_globals test_ping_serve test_blk test_nolibc
 
 UKVM_TESTS=$(subst test, _test_ukvm, $(TESTDIRS))
 VIRTIO_TESTS=$(subst test, _test_virtio, $(TESTDIRS))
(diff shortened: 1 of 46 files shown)


mato:
Historic, closing. malloc has since gone from the public API, so the immediate motivation for this is no longer there.

PR #153, opened 2017-01-23
Standardise license headers and copyright notices

mato:
This commit unifies the license headers and copyright notices in Solo5
and ukvm source files. Original source files use one of the following
standard headers:

For the Solo5 unikernel:

    Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file

    This file is part of Solo5, a unikernel base layer.

For ukvm:

    Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file

    This file is part of ukvm, a unikernel monitor.

The header is followed by the standard ISC license text.

Where a source file contains existing third-party licenses or copyright
notices these have been preserved and the Solo5 header has been added to
the top of the file.

Further, this commit updates the copyright notice in the top-level
LICENSE file to read:

    Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file

and adds a new AUTHORS file containing:

1. copyright notices for corporate contributors,
2. a list of individual contributor's names and their affiliation(s),
    if any.

I have populated the AUTHORS file with all contributors to date based on
the information in the git history.

PR #154, opened 2017-01-24
Documentation updates for 0.2.0 release

mato:
Now that #153 is in, this PR is the last step before tagging an 0.2.0 release for MirageOS 3.

Changes:
* I've reformatted `CHANGES` as `CHANGES.md` for consistency with other OPAM packages, and added a list of what I think are the most important changes since the initial release, mostly extracted from memory and #82.
* I've removed the majority of the MirageOS-specific instructions in `README.md` and/or replaced them with generic instructions referring to `test_ping_serve`. The main reason for this is that I don't want to duplicate what should go in the MirageOS documentation for the 3.0 release, and I don't want to include information which will be instantly out of date after the release ("use the `mirage-dev` OPAM remote", etc.)

@yomimono, @hannesm What do you think of the 2nd point above? It'd be even better to be able to point the user to a canonical URL for documentation about building with the Solo5 targets, but I don't know what that is...?

@djwillia Now would also be a good time to re-work the Acknowledgments section at the end of `README.md` if you want to do that...?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,2 +1,0 @@
-0.1.1 (21 July 2016)
-* Initial release for publishing to opam.ocaml.org.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ High-level user-visible changes in this release:
   Solo5 (`virtio` backed) unikernels on QEMU/KVM and FreeBSD/bhyve.
 * Improvements to the standalone test programs under `tests/`, including an
   automated test suite.
+* Standardised license headers and copyright notices across the codebase and
+  introduced an AUTHORS file listing contributors and their affiliations.
 
 Notable changes under the hood:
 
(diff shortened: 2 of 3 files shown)


yomimono:
>  I don't want to duplicate what should go in the MirageOS documentation for the 3.0 release, and I don't want to include information which will be instantly out of date after the release ("use the mirage-dev OPAM remote", etc.)

Very sensible.

> a canonical URL for documentation about building with the Solo5 targets, but I don't know what that is...?

@avsm might have an opinion on this, but it's a good candidate for a tutorial item since I think it would have to include instructions on how to run the artifact as well (which we should have for all the backends, not just the solo5 ones).

hannesm:
sounds good to me, and yes, I agree there should be somewhere a getting started which shows the different backends in a more central place

mato:
Small update to `CHANGES.md` pushed, adding an acknowledgements section for people who contributed to the release with non-code contributions and mentioning the license standardisation and AUTHORS file.

djwillia:
@mato Thanks for this - it looks good to me.  This PR seems like a good place to also get rid of the Acknowledgements section in README.md, seeing as it's out of date and from here on out CHANGES.md can contain acknowledgements anyway.

mato:
@djwillia Done in abdf2c8. Looks like we're good to go!

PR #155, opened 2017-02-01
remove USE_LOCKS from malloc

djwillia:
As part of https://github.com/Solo5/solo5/pull/150, I've been working on building Solo5 (and ultimately uhvf) locally on MacOSX.  Along the way, I was using clang 3.9, which threw some warnings concerning the USE_LOCKS macros in dlmalloc.  We don't use malloc locks anyway, so it seemed best to remove that code.

### DIFF
--- a/kernel/malloc.c
+++ b/kernel/malloc.c
@@ -654,26 +654,6 @@ MAX_RELEASE_CHECK_RATE   default: 4095 unless not HAVE_MMAP
 /* The maximum possible size_t value has all bits set */
 #define MAX_SIZE_T           (~(size_t)0)
 
-#ifndef USE_LOCKS /* ensure true if spin or recursive locks set */
-#define USE_LOCKS  ((defined(USE_SPIN_LOCKS) && USE_SPIN_LOCKS != 0) || \
-                    (defined(USE_RECURSIVE_LOCKS) && USE_RECURSIVE_LOCKS != 0))
-#endif /* USE_LOCKS */
-
-#if USE_LOCKS /* Spin locks for gcc >= 4.1, older gcc on x86, MSC >= 1310 */
-#if ((defined(__GNUC__) &&                                              \
-      ((__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 1)) ||      \
-       defined(__i386__) || defined(__x86_64__))) ||                    \
-     (defined(_MSC_VER) && _MSC_VER>=1310))
-#ifndef USE_SPIN_LOCKS
-#define USE_SPIN_LOCKS 1
-#endif /* USE_SPIN_LOCKS */
-#elif USE_SPIN_LOCKS
-#error "USE_SPIN_LOCKS defined without implementation"
-#endif /* ... locks available... */
-#elif !defined(USE_SPIN_LOCKS)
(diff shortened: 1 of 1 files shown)


hannesm:
Mirage3 beta tester here.  I recommend to merge this PR and make another release of Solo5.  On FreeBSD, I otherwise (with `solo5-kernel-virtio-0.2.0`) and clang 3.9.0 (tags/RELEASE_390/final 280324):

```
cc -nostdlibinc -ffreestanding -mno-red-zone -isystem /usr/home/hannes/devel/mirage/solo5/include-host -std=gnu99 -Wall -Wextra -Werror -O2 -g -c malloc.c -o malloc.o
malloc.c:642:5: error: macro expansion producing 'defined' has undefined behavior
      [-Werror,-Wexpansion-to-defined]
#if USE_LOCKS /* Spin locks for gcc >= 4.1, older gcc on x86, MSC >= 1310 */
    ^
malloc.c:638:22: note: expanded from macro 'USE_LOCKS'
#define USE_LOCKS  ((defined(USE_SPIN_LOCKS) && USE_SPIN_LOCKS != 0) || \
                     ^
malloc.c:642:5: error: macro expansion producing 'defined' has undefined behavior
      [-Werror,-Wexpansion-to-defined]
malloc.c:639:22: note: expanded from macro 'USE_LOCKS'
                    (defined(USE_RECURSIVE_LOCKS) && USE_RECURSIVE_LOCKS != 0))
                     ^
malloc.c:1543:5: error: macro expansion producing 'defined' has undefined behavior
      [-Werror,-Wexpansion-to-defined]
#if USE_LOCKS
    ^
```

mato:
Can we achieve the same effect with a one line change adding

    #define USE_LOCKS 0

...at the top of `malloc.c` where we define the other "configuration" directives?

If not, why not?

@hannesm Why haven't we run into this before? Is it something to do with a newer `clang`?

hannesm:
@mato `#define USE_LOCKS 0` at the top of `malloc.c` works as well (compiles fine with clang-3.9).

and we didn't run into this before since earlier compilation was done with clang-3.8 (3.9 was merged to FreeBSD-CURRENT at the end of November, it looks like I didn't recompile solo5 since then on my laptop). FreeBSD-11 (the latest release, I deploy my solo5 unikernels on such a machine) still uses clang 3.8.0.

mato:
Ok, that makes sense. @djwillia Can you confirm `#define USE_LOCKS 0` works for you as well? If yes then I'll make that one line change and cut a new release of Solo5 at the same time.

djwillia:
@mato yes, `#define USE_LOCKS 0` works for me as well.  Is the reason you want to keep that code around because you think the code should remain as similar to stock dlmalloc as possible, or because you think we may want locks in the future, or some other reason?  I had opted for removal as I feel the code is easier to read without as many nontrivial `#ifdefs`, and I don't think we will need locks for the foreseeable future.

mato:
@djwillia Both reasons that you mention and further: The first question that came into my mind when I read the change in isolation was "_Why_ is this change _removing_ the lock code rather than just turning it off, given that there's a perfectly good mechanism for the latter?"

Anyhow, thanks for acking that the one-liner works, I'll now make a new PR with that change.

mato:
Superseded by #156.

Issue #151, opened 2017-01-17
ukvm on ARM


djwillia:
ukvm (and Solo5) is currently x86 specific.  It would be great if they ran on ARM as well.  I believe most of the mirageOS components already work with ARM.  A couple of discussion points:
- how architecture independent can we make the solo5 code (e.g., by pushing hw-specific setup out of the kernel and into ukvm)
- how should the ukvm code be arranged to cleanly support multiple platforms (e.g., KVM/Linux, Hypervisor.framework/MacOSx, vmmapi.h/FreeBSD?) and multiple architectures (e.g., x86, ARM) at the same time?

Weichen81:
Hi Daniel,

My colleague Kaly had contacted you before. We're trying to port ukvm and Solo5 to ARM. But we are in  the very first stage. We have done some investigation and set up a development environment on ARM server. Currently, we can compile the sample apps in mirage-skeleton, but of course, they could not run.

In our consider, we can refer to Linux about the architecture independent design. We can create
an "arch" folder to store different architecture files. For example, use "arch/arm64" to store arm64
architecture files like cpu.S, cpu_vectors.S and etc. And we can split the system APIs that are used in kernel.c into two parts, use intr_init as an example:
1. Abstract interface: 
    void intr_init(void) 
    {
        ...
        arch_intr_init();
        ...
    }
2. Architecture implementation: arch_intr_init in arch/arm64/intr.c
 
About the ukvm code, I think it has combined with the x86_64 and KVM/Linux. I am sorry, I am not
very familiar with Hypervisor.framework/MacOSx, vmmapi.h/FreeBSD. The first plan could be using #ifdef CONFIG_ARM64 to add ARM64 support.

Above are the initiatory considers from us,  we'd like to discuss with you to push the multiple architectures/platforms supports of solo5 and ukvm :) .

Regards.

djwillia:
Ideally, all architecture-dependent code for Solo5 should be in one of the backends: `kernel/virtio/` or `kernel/ukvm/`.  Clearly this is not the case now, because of things like `kernel/cpu.S` that you point out.

It should be possible to move all architecture-dependent code from `kernel/` to `kernel/<backend>`, by trivially duplicating the code in either backend.

So, assuming `kernel/` has no architecture-dependent code left, we can consider each backend separately.  I believe `kernel/virtio` will need an architecture subdir because it expects to run on fully emulated hardware.  

[...]

PR #157, opened 2017-02-15
refactor to remove/clarify arch-specific parts of ukvm unikernel

djwillia:
The purpose of these changes is to move all asm and x86-specific code for the ukvm path into one place.  This should help with the ARM port (https://github.com/Solo5/solo5/issues/151) and hopefully the Hypervisor.framework patch too (https://github.com/Solo5/solo5/pull/150).

The main changes are:
* rely on ukvm to set up hardware state (gdt, sse)
* interrupt handling is completely removed from ukvm
* all assembly (inline or otherwise) for the ukvm target is in `kernel/cpu.S`, `kernel/cpu.h`, or `kernel/ukvm/ukvm.h`

Practically, this means some code moved from `kernel/` to `kernel/virtio`.

Some of the inline assembly in `kernel/` was in `malloc.c`, but it had to do with locks, which are unnecessary and targeted for removal in PR https://github.com/Solo5/solo5/pull/155, which I have also merged here.


mato:
AFAICT this change is incomplete -- you remove the arch/cpu-specific code from `kernel/ukvm` but I don't see any code in the monitor that replaces it?

djwillia:
You are right, there is more being removed than added back in.  If we don't support guest handling of interrupts (which we don't; we had interrupt injection code in ukvm at some point but removed it), then it turns out that we do not need to enable interrupts in the guest.  This in turn means that we do not need an IDT to be set up in the guest (BTW, this is all related to one of your comments in https://github.com/Solo5/solo5/pull/150).  

A GDT and page table mapping were already being done in ukvm and do not need to be redone in the guest.  The enabling of SSE for floating point is now in ukvm https://github.com/Solo5/solo5/pull/157/files#diff-2db47873a953eeb11019d73e0e2d16bfR302, but I believe that is the only addition to ukvm from what was already there.

ukvm/ukvm-core.c
```diff
@@ -294,8 +294,18 @@ static void load_code(const char *file, uint8_t *mem,     /* IN */
 
 static void setup_system_64bit(struct kvm_sregs *sregs)
 {
-    sregs->cr0 |= X86_CR0_PE;
-    sregs->efer |= EFER_LME;
+    uint64_t cr0 = (X86_CR0_NE | X86_CR0_PE | X86_CR0_PG)
+        & ~(X86_CR0_NW | X86_CR0_CD);
+    uint64_t cr4 = X86_CR4_PAE;
```
ricarkol: cr4:`X86_CR4_PAE`  and cr0:`X86_CR0_PG` are already set in `setup_system_page_tables`. You might want to remove them from `setup_system_page_tables` or here.

kernel/malloc.c
```diff
@@ -654,26 +654,6 @@ MAX_RELEASE_CHECK_RATE   default: 4095 unless not HAVE_MMAP
 /* The maximum possible size_t value has all bits set */
 #define MAX_SIZE_T           (~(size_t)0)
 
-#ifndef USE_LOCKS /* ensure true if spin or recursive locks set */
```
ricarkol: Just a reminder that you should not delete this code anymore. I guess git will tell you that.

ukvm/ukvm-cpu.h
```diff
@@ -42,19 +42,26 @@
...
 
 /*
- * Intel CPU features in CR4
+ * CPU control registers
  */
-#define X86_CR4_PAE_BIT		5 /* enable physical address extensions */
-#define X86_CR4_PAE		_BITUL(X86_CR4_PAE_BIT)
+#define X86_CR0_PE 0x00000001 /* Protected mode Enable */
+#define X86_CR0_NE 0x00000020 /* Numeric Error enable (EX16 vs IRQ13) */
+#define X86_CR0_PG 0x80000000 /* PaGing enable */
```
ricarkol: This might be a bit of a pain to change (and it's up to you), but using `_BITUL` made it easier to check for the flag number directly. For example, mapping 0x80000000 to 31 is not as "straightforward" as before.

mato:
Superseded by #192.

PR #158, opened 2017-02-21
Explicitly define USE_LOCKS to 0 in dlmalloc

mato:
Clang 3.9 does not parse the default value of USE_LOCKS (0). This
change explicitly sets the value to 0 to appease the compiler. No
functional change intended.


mato:
Supersedes #155.

PR #163, opened 2017-03-06
Ukvm FPU setup & test case

Kensan:
Setup FPU in UKVM monitor instead of guest and add test case for floating point/FPU operation.


mato:
This looks good to me. Given that the same code also sets up SSE, could you add some code to the test that verifies that SSE instructions are also working? Inline `__asm__` is fine, since otherwise you'd have to force the test to build at -O3 to force the compiler to eg. vectorize loops.

Kensan:
I extended the test case as requested.

mato:
Thanks, I added a small fix to the test invocation (and found some other bugs in the process, will fix). This is good to merge.

PR #165, opened 2017-03-08
Fix check for KVM 32-bit gap

mdlemay:
The existing check seems to be incorrect and inconsistent with the comment that precedes it.

### DIFF
--- a/ukvm/ukvm-core.c
+++ b/ukvm/ukvm-core.c
@@ -628,7 +628,7 @@ int main(int argc, char **argv)
      * slots: one before the pci gap, and one after it.
      * Reference: kvmtool x86/kvm.c:kvm__init_ram()
      */
-    assert(GUEST_SIZE < KVM_32BIT_GAP_SIZE);
+    assert(GUEST_SIZE < KVM_32BIT_GAP_START);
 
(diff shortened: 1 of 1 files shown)


ricarkol:
That check was added just in case KVM emulates the PCI gap just below address 4GB. I'm not sure if it does. I guess the easiest way to check is by writing and reading into/from it (i.e. starting at 0xe0000000). In any case, kvmtool and qemu call KVM_SET_USER_MEMORY_REGION with two memory regions, one before the hole and one after; we might want to do the same to be safe.

mdlemay:
I was actually just pointing out that it looks like GUEST_SIZE is checked against the wrong variable: KVM_32BIT_GAP_SIZE instead of KVM_32BIT_GAP_START.

ricarkol:
I could swear I saw the line as deleted! anyway, good catch.

ricarkol:
Should be safe to merge, so merging.

mato:
Belated comment: My understanding of the "32 bit memory hole" is that it's not actually _imposed_ by KVM but by the machine being emulated, i.e. it's there so that ACPI tables/PCI/other stuff can be mapped into that space. For our use of KVM in ukvm it can probably be removed entirely, which will happen after someone verifies that my statements here are true :-)

PR #166, opened 2017-03-09
Replace kernel/lib.c with musl-derived implementations

mato:
This change replaces the functions defined in kernel/lib.c with
musl-derived implementations, which are then also used by the tests.

This change is motivated by an immediate need for known-good
implementations of these functions. I had not reviewed the original
implementations until running into major bugs in them today. Eventually
we want a better integration of "nolibc" into both kernel (Solo5) and
application (test) code, so this particular arrangement is temporary.

For simplicity the musl-derived implementations have been stripped down
to use the generic "fallback" implementations where possible.

### DIFF
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -64,6 +64,8 @@ all: virtio ukvm
 virtio: virtio/solo5.o 
 ukvm: ukvm/solo5.o 
 
+CFLAGS+=-D__SOLO5_KERNEL__
+
 %.o: %.c $(HEADERS)
 	$(CC) $(CFLAGS) -c $< -o $@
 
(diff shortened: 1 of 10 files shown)


mato:
@djwillia @ricarkol The TL;DR of this PR is: "spot the bugs in `strcmp()`". This is the _minimal_ change to just replace the implementations wholesale without having to think about it much so that I can get on with other things. #149 (or something like it) is still on the cards for both kernel and tests but that will take more time.

mato:
Merging this as the existing implementations need to be terminated with extreme prejudice.

hannesm:
do you plan to do a solo5 release containing these changes?  would be appreciated.  I tested compilation on FreeBSD (CURRENT with clang-3.9, and 11 with clang-3.8) successfully.

mato:
On Thursday, 09.03.2017 at 05:19, Hannes Mehnert wrote:
> do you plan to do a solo5 release containing these changes?  would be appreciated.  I tested compilation on FreeBSD (CURRENT with clang-3.9, and 11 with clang-3.8) successfully.

Yes.

PR #168, opened 2017-03-13
Paths for Syslinux files in Arch Linux

gjaldon:
Running `solo5-mkimage -f tar blog.tar.gz ./main.native` was failing for me since it can't find `mbr.bin`, `libcom32.c32` and `mboot.c32`. 

Adding the paths for those in this PR creates an image without error. I did encounter another error when uploading the image output Google Compute Engine which may be unrelated:
```
Loading unikernel.bin... ok
Invalid Multiboot image: neither ELF header nor a.out kludge found
boot: 
```
Any idea how to resolve the error above?

### DIFF
--- a/tools/mkimage/solo5-mkimage.sh
+++ b/tools/mkimage/solo5-mkimage.sh
@@ -145,11 +145,14 @@ type mtools >/dev/null 2>/dev/null || die "need mtools installed"
 type mkdosfs >/dev/null 2>/dev/null || die "need dosfstools installed"
 
 SYSLINUX_MBR=$(maybefind /usr/lib/syslinux/mbr/mbr.bin \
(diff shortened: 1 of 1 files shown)


mato:
The change looks fine. I've no idea why you're getting that error, what version of syslinux?

gjaldon:
I'm using syslinux 6.03 which seems to be the latest release.

mato:
I have Debian 6.03+dfsg-14.1 here and it works fine. I'm merging this change since it works fine, please open a separate issue for the syslinux problem.

gjaldon:
Disregard the other issue I mentioned. I was creating an image out of the wrong unikernel (the unikernel created for unix).

PR #169, opened 2017-03-13
Replace ifconfig with ip command

gjaldon:
It seems net-tools is supposed to be phased out and scripts are discouraged from using them. https://lwn.net/Articles/710533/

Only found out about this since I noticed my OS didn't have `ifconfig` by default when I ran `solo5-run-virtio`. Turns out some distros no longer include it by default and expecting people to use iproute2 instead.

### DIFF
--- a/tools/run/solo5-run-virtio.sh
+++ b/tools/run/solo5-run-virtio.sh
@@ -80,7 +80,13 @@ while true; do
         ;;
     -n)
         NETIF="$2"
-        ifconfig ${NETIF} >/dev/null || die "no such network interface: ${NETIF}"
+        # Check dependencies
+        type ip &>/dev/null ||
+            type ifconfig &>/dev/null ||
+            die "need ip or ifconfig installed"
+        ip a show ${NETIF} &>/dev/null ||
+            ifconfig ${NETIF} &>/dev/null ||
+            die "no such network interface: ${NETIF}"
         shift; shift
         ;;
     -q)


mato:
This script needs to be OS-agnostic, it's used at least on FreeBSD as well as Linux. I would suggest trying "ip" followed by "ipconfig" on Linux, and only the latter on anything else.

gjaldon:
The latest commit checks for `ip` or `ifconfig` and raises if those aren't installed. It then proceeds to use those commands to check if the network interface passed via `-n` exists. Is this behavior satisfactory?

Removed the previous commit, btw.

tools/run/solo5-run-virtio.sh
```diff
@@ -80,7 +80,13 @@ while true; do
         ;;
     -n)
         NETIF="$2"
-        ifconfig ${NETIF} >/dev/null || die "no such network interface: ${NETIF}"
+        # Check dependencies
+        type ip &>/dev/null ||
+            type ifconfig &>/dev/null ||
```
mato: I didn't know there was such a thing as `&>`. Nice.

mato:
Yes, this is fine. Thanks!

tools/run/solo5-run-virtio.sh
hannesm: I'm not sure which shell(s) support `&>`, my `/bin/sh` does not (also, http://www.shellcheck.net/ complains:
```
SC2039: In POSIX sh, 'type' is undefined.
SC2039: In POSIX sh, &> is undefined.
```

mato: `type` is provided in `/usr/bin` on FreeBSD so that's fine. You're right about `&>` being a (presumably) bashism. I'll fix that.

hannesm: thanks.  and yes, `type` seems to be POSIX (see http://pubs.opengroup.org/onlinepubs/9699919799/utilities/type.html).

PR #174, opened 2017-03-31
remove side effect from assert

djwillia:
I just noticed that this assert statement has the actual write call in it, causing a side-effect to the assertion.  Not an issue now, but it might cause a bug later: if someone redefines assert to do nothing rather than perform the check, they would assume no semantic changes.  I didn't see any other asserts in ukvm with this problem.

### DIFF
--- a/ukvm/ukvm-core.c
+++ b/ukvm/ukvm-core.c
@@ -402,9 +402,11 @@ void ukvm_port_puts(uint8_t *mem, uint64_t paddr)
 {
     GUEST_CHECK_PADDR(paddr, GUEST_SIZE, sizeof (struct ukvm_puts));
     struct ukvm_puts *p = (struct ukvm_puts *)(mem + paddr);
+    int ret;
 
     GUEST_CHECK_PADDR(p->data, GUEST_SIZE, p->len);
-    assert(write(1, mem + p->data, p->len) != -1);
+    ret = write(1, mem + p->data, p->len);
+    assert(ret != -1);
 }
 
 void ukvm_port_poll(uint8_t *mem, uint64_t paddr)


ricarkol:
There are these other asserts with side-effects :D :

```
kernel/ukvm/time.c:    assert(pvclock_init() == 0);
kernel/virtio/virtio_net.c:    assert(virtq_add_descriptor_chain(&recvq, recvq.next_avail & mask, 1) == 0);
kernel/virtio/virtio_blk.c:    assert(virtq_add_descriptor_chain(&blkq, head, 3) == 0);
tests/test_ping_serve/test_ping_serve.c:        assert(solo5_net_read_sync(buf, &len) == 0);
```

mato:
Redefining `assert()` to do nothing is, in my book, _bad_. The code I write is written with `assert()` used to enforce contracts (among other things). Perhaps we should define a `ukvm_assert()` and explicitly document it as "mandatory, do not remove"?

Notwithstanding the above, `assert()` with side-effects is hard to read so I'm all for changing those to improve code readability.

However, if we decide to do either of the above, let's do it after #171 lands and let's do it all in one go rather than many small PRs changing individual cases.

djwillia:
I agree.  Should we close this and create an issue named "remove assert side-effects for readability" or something so that we don't forget?

mato:
> I agree. Should we close this and create an issue named "remove assert side-effects for readability" or something so that we don't forget?

Yes, perhaps a general issue to discuss what the error handling should look like. See also my comment here about the _modus operandi_: https://github.com/Solo5/solo5/pull/178#discussion_r110109969

mato:
Historic, closing.

PR #178, opened 2017-04-05
ukvm_module_net: Check if interface is up and running

Kensan:
Prior to attaching to a tap device, check that the network inferface is up and running. This gives the user more meaningful feedback regarding this error.

Furthermore, switch to wait-for-packet variant in the ping serve test because ukvm does no longer raise an assertion.

### DIFF
--- a/ukvm/ukvm_module_net.c
+++ b/ukvm/ukvm_module_net.c
@@ -83,11 +83,13 @@ static int tap_attach(const char *ifname)
     }
 
     /*
-     * Verify that the interface exists. If we don't do this then we get
-     * "create on open" behaviour on most systems which is not what we want.
+     * Verify that the interface exists and is up and running. If we don't do
+     * this then we get "create on open" behaviour on most systems which is not
+     * what we want.
      */
     struct ifaddrs *ifa, *ifp;
     int found = 0;
+    int up = 0;
 
     if (getifaddrs(&ifa) == -1)
         return -1;
@@ -95,6 +97,7 @@ static int tap_attach(const char *ifname)
     while (ifp) {
         if (strcmp(ifp->ifa_name, ifname) == 0) {
             found = 1;
+            up = ifp->ifa_flags & (IFF_UP | IFF_RUNNING);
             break;
         }
         ifp = ifp->ifa_next;
@@ -102,6 +105,10 @@ static int tap_attach(const char *ifname)
     freeifaddrs(ifa);
     if (!found) {
         errno = ENOENT;
+        return -1;
+    }
+    if (!up) {
+        errno = ENETDOWN;
         return -1;
     }
 
--- a/ukvm/ukvm_module_net.c
+++ b/ukvm/ukvm_module_net.c
@@ -46,6 +46,8 @@
 
 #elif defined(__FreeBSD__)
 
+#include <net/if.h>
+
 #else /* !__linux__ && !__FreeBSD__ */
 
 #error Unsupported target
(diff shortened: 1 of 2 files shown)


ukvm/ukvm_module_net.c
```diff
@@ -178,8 +178,10 @@ static void hypercall_netwrite(struct ukvm_hv *hv, ukvm_gpa_t gpa)
     int ret;
 
     ret = write(netfd, UKVM_CHECKED_GPA_P(hv, wr->data, wr->len), wr->len);
-    assert(wr->len == ret);
-    wr->ret = 0;
+    if (wr->len == ret)
+        wr->ret = 0;
+    else
+        wr->ret = -1;
 }
```
mato: This is problematic since host-side errors can also be permanent (eg. "Interface is down"). What specific error were you getting, and under what conditions?

Kensan: That was actually the issue I observed. I started the unikernel without setting the interface up. With the change this sequence of events is not fatal any longer:

- Start test_ping_server unikernel
- Bring up and configure tap interface
- Ping 10.0.0.2

Looking at the error handling code in [test_ping_serve.c ](https://github.com/Solo5/solo5/blob/master/tests/test_ping_serve/test_ping_serve.c#L248) which checks if net write returns -1 (cannot happen with the current code) gave me the impression that it was intended for users to be able to handle errors.

mato: There are two problems here:

1. The _modus operandi_ for error handling in Solo5, the ukvm hypercall API and ukvm itself has not really been thought through. This needs solving, but is a hard problem to get right, hence the aggressive use of "fail fast" for now.
2. This specific problem which I would describe as "you get an obscure error message if the host side of the tap interface is down, despite the guest having started up correctly".

I think the right solution for (2), without going into (1), is that we check that the interface is indeed up in ukvm `tap_attach()` by adding a check for `IFF_UP` (and `IFF_RUNNING`?) at [ukvm_module_net.c:96](https://github.com/Solo5/solo5/blob/master/ukvm/ukvm_module_net.c#L96) and have the function return `ENETDOWN` if not.

I would leave the assertion on the write path alone for now.

What do you think?

Kensan: Thanks for the clarification. I was not aware of (1) and read too much into the error checking code in test_ping_serve.c.

In this case, failing hard with an assertion in the write path is definitely better and I agree with your solution for (2).

mato: Sounds good. Do you want to have a go at the solution for (2)?

Kensan: Yes, will update the PR once I have something.

mato:
I pushed a minor fix for FreeBSD, otherwise LGTM. Thanks!

hannesm:
this breaks FreeBSD at runtime:  I have the sysctl `net.link.tap.up_on_open=1`, which will put the interface into UP state once it has been `open()`ed.  But this PR first checks that the device is `IFF_UP | IFF_RUNNING` _before_ the call to `open`.  I'm not sure what the semantics of `ifa_flags` is -- on Linux the interface is already up when it is present, but no consumer does use it yet?

mato:
@hannesm The semantics of tap interfaces are weird everywhere. On Linux all the scripts I use (which are also what's described in the documentation) require the user to configure the tap interface upfront and bring it up manually, so this was my attempt at emulating the same semantics on FreeBSD. I must have tested just the version without #178, thanks for catching.

ukvm/ukvm_module_net.c
hannesm: @mato I think this is the most verbose thread about error handling in ukvm (/solo5) -- could we move it into a separate issue (since discussions on merged PRs is slightly annoying) -- #174 seems to be related as well.

PR #164, opened 2017-03-07
Solo5 verbosity

ricarkol:
Adding two verbosity related args for solo5 unikernels: `--quiet` and `--debug`.

The changes:

- Most current `printf`s were changed to `log(INFO,...)` and `log(WARN,...)`, and the default verbosity level was set to `INFO` (everything that was logged is still logged).
- Parsing and "eating" the --quiet and --debug args is done inside the unikernel at `_start` before calling `solo5_app_main`. To make things simpler, --quiet/--debug args will be understood as --quiet/--debug only if they are the first argument(s) to the unikernel. For example, `${UNIKERNEL} --quiet blabla` and `${UNIKERNEL} --debug --quiet bleble` will run in quiet mode, while `${UNIKERNEL} blabla --quiet` won't. The cmdline argument to the applications for the previous examples would be `blabla`, `bleble` and `blabla --quiet` respectively.  A large set of examples of how things are parsed can be found at `test-verbosity.sh`.
- Added a test-quiet test into `run-tests.sh`.
- Added a bunch of other tests into `test-verbosity.sh`.

This is part of #161 .

PR #183, opened 2017-04-10
Various fixes to self-tests

mato:
Fix some bugs introduced by #164 and others discovered in the process of fixing them. Tests now pass again on FreeBSD. See individual commits for details.

/cc @ricarkol

### DIFF
--- a/tests/run-tests.sh
+++ b/tests/run-tests.sh
@@ -298,7 +298,7 @@ for T in ${TESTS}; do
         OPTS=
     fi
     printf "%-32s: " "${NAME}"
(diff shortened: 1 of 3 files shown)


ricarkol:
Thanks @mato . Sorry for breaking stuff, didn't test freeBSD. Talking about that, can surf CI use freeBSD? also, what about using pytest or bats?

hannesm:
@ricarkol I tried to set it up and run into https://github.com/surf-build/surf/issues/63 -- it may be that someone familiar with npm has suggestions (also, unclear to me whether there were API changes between nodegit-0.16 and 0.18).

PR #189, opened 2017-04-11
ukvm_net: move check whether interface is up to after open() on FreeBSD

hannesm:
This fixes solo5-net running on FreeBSD (update to #178) with `tap` interfaces which are only up after `open()` was called.

EDIT: this moves the check under the `#ifdef linux` conditional.  It also does not check for the interface being `up` on FreeBSD.

### DIFF
--- a/ukvm/ukvm_module_net.c
+++ b/ukvm/ukvm_module_net.c
@@ -109,12 +109,13 @@ static int tap_attach(const char *ifname)
         errno = ENOENT;
         return -1;
     }
+
+#if defined(__linux__)
+
     if (!up) {
         errno = ENETDOWN;
         return -1;
     }
-
-#if defined(__linux__)
 
     int err;
     struct ifreq ifr;


ukvm/ukvm_module_net.c
```diff
@@ -109,13 +107,14 @@ static int tap_attach(const char *ifname)
         errno = ENOENT;
         return -1;
     }
-    if (!up) {
+
+#if defined(__linux__)
+
+    if (! (ifp->ifa_flags & (IFF_UP | IFF_RUNNING))) {
```
mato: This is a use-after-free, `ifa` (and thus `ifp`) is freed on line 105 (or 107 in the original).

I suggest to keep the original structure (collecting `up` inside the for loop) and just make the `if (!up)` conditional on `__linux__`.

hannesm: force-pushed... interesting that no C compiler told me about this use-after-free...

PR #192, opened 2017-04-24
kernel: Refactoring for multi-arch support

mato:
This is the guest-side follow-up to #171, refactoring the kernel side to allow for multi-arch support.

Briefly, this separates out the kernel arch-specific code into `cpu_<architecture>.[ch]` and `cpu_vectors_<architecture>.S`. It also includes the build system infrastructure needed for multiarch. See individual commits for details.

/cc #151 #190


mato:
@djwillia @ricarkol (and anyone else interested) Please review...

Weichen81:
@mato I have reviewed the changes. I think they can make ARM64 guest side code landing easily.
I have some concern about the build scripts. Can we provide an option for user to select ukvm or
virtio? Because these two kernels are exclusive while we are using opam install. And the ARM64 porting
currently can only work with ukvm :)

Regards,
Wei Chen

mato:
@Weichen81 I would just add a check to force `BUILD_VIRTIO=no` when the target architecture is not `x86_64`. FWIW you can already build `ukvm` only by doing `make ukvm` at the top level.

kernel/cpu_x86_64.c
```diff
@@ -0,0 +1,196 @@
...
+{
+    log(INFO, "Solo5: trap: type=%s ec=0x%lx rip=0x%lx rsp=0x%lx rflags=0x%lx\n",
+        traps[num], regs->ec, regs->rip, regs->rsp, regs->rflags);
+    if (num == 14)
+        log(INFO, "Solo5: trap: cr2=0x%lx\n", regs->cr2);
+    PANIC("Fatal trap");
+}
+
+/* keeps track of how many stacked "interrupts_disable"'s there are */
+int intr_depth = 1;
```
Kensan: Should this be volatile since it may be accessed by nested interrupts?

mato:
@Kensan @diwillia I have pushed 2ef936d to address the outcome of our discussion about consolidating x86 FPU/SSE unit setup in #190. This compiles and passes all tests on both Linux and FreeBSD. Please take a look and let me know what you think.

kernel/cpu_x86_64.c
mato: Per the comment in `idt_fillgate()` (https://github.com/mato/solo5/blob/multiarch/kernel/cpu_x86_64.c#L57), we run all interrupt handlers with interrupts disabled. So this should be fine as is.

Kensan: Sorry I misread, I wrongly had the impression that the variable was also accessed when handling exceptions/traps.

Kensan:
@mato The change looks fine to me.

A note regarding Muen/#190: I will want to override **fpu_init()** (by splitting it into a separate compilation unit) since the control register accesses create traps. On Muen I really only want to perform the MXCSR initialization part. Will adapt #190 accordingly once this PR lands on master.

mato:
I've reworked these changes to incorporate the outcome of the out-of-band discussions around FPU/SSE setup, which was to keep things as they are for now (i.e. do the majority of setup on the monitor side) and add a `platform_init()` which can be used for platform-specific initialisation such as LDMXCSR for the Muen port.

This led to somewhat more refactoring than I expected, mostly on the virtio side. The outcome of this is that I've managed to unify and clean up `mem.c` for both platforms which is a good thing. Also, the virtio initialisation / bootstrap flow is now much clearer. See the individual commits for details.

While I was at it I also excised / `#ifdef`'d most of the remaining arch-dependent code with the exception of the call to `tscclock_init()` from `time_init()` on the ukvm side. 

@Weichen81: You'll need to add an appropriate clock for ARM64 and `#ifdef` the call to it's init function (if any). I've also addressed your comment about not building virtio on !x86_64.

@djwillia Please take a look and let me know if this is OK to merge.

Weichen81:
Hi mato,

Thanks for addressing the virtio issue of ARM64. And we are trying to use the arm generic timer to implement the clock functions for ARM64. It will be completed in the next few days.

Regards,

在 2017年5月5日，22:25，Martin Lucina <notifications@github.com<mailto:notifications@github.com>> 写道：


I've reworked these changes to incorporate the outcome of the out-of-band discussions around FPU/SSE setup, which was to keep things as they are for now (i.e. do the majority of setup on the monitor side) and add a platform_init() which can be used for platform-specific initialisation such as LDMXCSR for the Muen port.

This led to somewhat more refactoring than I expected, mostly on the virtio side. The outcome of this is that I've managed to unify and clean up mem.c for both platforms which is a good thing. Also, the virtio initialisation / bootstrap flow is now much clearer. See the individual commits for details.

While I was at it I also excised / #ifdef'd most of the remaining arch-dependent code with the exception of the call to tscclock_init() from time_init() on the ukvm side.

@Weichen81<https://github.com/Weichen81>: You'll need to add an appropriate clock for ARM64 and #ifdef the call to it's init function (if any). I've also addressed your comment about not building virtio on !x86_64.

@djwillia<https://github.com/djwillia> Please take a look and let me know if this is OK to merge.

kernel/virtio/solo5.lds
```diff
@@ -3,19 +3,12 @@ ENTRY(_start)
...
-    }
-
-    . = ALIGN(0x1000);
-
     /* Code */
     _stext = .;
 
     .text :
     {
+        *(.data.multiboot)
```
mato: That's correct. The requirements are that the header be within the first 8k of the executable and be 4-byte aligned, both of which are taken care of. I tested GCE with these commits and everything works.

kernel/virtio/boot.S
```diff
@@ -27,29 +27,28 @@
  */
 
 #include "multiboot.h"
+#include "../cpu_x86_64.h"
```
mato: We can do that if (unlikely) virtio ever gets ported to a non x86_64 architecture.

kernel/virtio/boot.S
```diff
@@ -141,42 +140,58 @@ END(_start)
 
 .data
 .align 64
-.globl cpu_gdt64
-cpu_gdt64:
+gdt64:
 	.quad 0x0000000000000000
-	.quad 0x00af9b000000ffff	/* 64bit CS		*/
+	.quad GDT_DESC_CODE_VAL		/* 64bit CS		*/
```
mato: It does not, LIMIT is bits (0 - 15) and (48 - 51) all of which are 1. The values were taken directly from the result of the computation in `ukvm_x86_setup_gdt()`. The reason I made this change was to deliberately remove any confusion/differences between the values used by `ukvm` and those used by `virtio`.

ricarkol: You are right, sorry for the confusion. The field changed is type which for a code segment means:
```
before: 0x00af9b000000ffff -> type = 0xb (Execute-Only, accessed)
now: 	0x00af99000000ffff -> type = 0x9 (Execute/Read, accessed)
```

kernel/virtio/solo5.lds
ricarkol: Awesome, thanks for the test.

ricarkol:
LGTM (just in case)

Issue #16, opened 2016-05-07
gdb breakpointing is relying on single stepping every instruction


ricarkol:
Right now, in order to break on a specific function we are single stepping on every single instruction running in the VM and comparing the current PC to a list of breakpoints. This is extremely slow because the vcpu is exiting for every single instruction (although it works). Ideally, we could use some other technique:
1. mark the page as invalid so we only get exits on the page that includes the breakpoint (the instruction).
2. update the guest memory with some breakpoint instruction, so we get an exit just for the breaking instruction.
3. use some KVM API for this. There has to be some KVM api to break on a specific PC. Can't find it in the docs and QEMU does not seem to be using any. But there has to be something...

mato:
Fixed in #198.

PR #193, opened 2017-05-26
ARM64 support

Weichen81:
This pull request is to enable the arm64 support on ukvm.
the issue is：
https://github.com/Solo5/solo5/issues/151


mato:
@Weichen81 Passes basic `tests/run-tests.sh` for me on an ARM64 box. Will now do a full review.

/cc @avsm @djwillia

mato:
@Weichen81 Network I/O does not work for me:
````
$ tests/test_ping_serve/ukvm-bin --net=tap100 tests/test_ping_serve/test_ping_serve.ukvm 
ukvm-bin: /home/mato/solo5/ukvm/ukvm_core.c:101: Invalid guest access: gpa=0x11711ffffd50, sz=16
$ tests/test_ping_serve/ukvm-bin --net=tap100 tests/test_ping_serve/test_ping_serve.ukvm 
ukvm-bin: /home/mato/solo5/ukvm/ukvm_core.c:101: Invalid guest access: gpa=0x276b1ffffd50, sz=16
````
This is after having run `tests/setup-tests.sh` as root to bring up the `tap100` interface.

**Comment by Weichen81** [comment]
@mato It's a bug of a uninitialized variable on stack. This will make the high 32-bit gpa become random
and cause the UKVM_CHECKED_GPA_P failed in some applications.
I have sent a patch to fix this bug.

File: ukvm/ukvm_hv_kvm_aarch64.c [hunk 1]
```diff
@@ -0,0 +1,137 @@
...
+ * AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * ukvm_hv_kvm_x86_64.c: x86_64 architecture-dependent part of KVM backend
+ * implementation.
```
mato: -> aarch64

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -0,0 +1,137 @@
...
+            handled = ((*fn)(hv) == 0);
+        if (handled)
+            continue;
+
+        struct kvm_run *run = hvb->vcpurun;
+
+        switch (run->exit_reason) {
+        case KVM_EXIT_MMIO: {
+            if (!run->mmio.is_write || run->mmio.len != 4)
+                errx(1, "Invalid guest port access: port=0x%llx", run->mmio.phys_addr);
```
mato: "port access" -> "MMIO access"

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -0,0 +1,137 @@
...
+
+            if (run->mmio.phys_addr < UKVM_HYPERCALL_PIO_BASE ||
+                    run->mmio.phys_addr >=
+                    (UKVM_HYPERCALL_PIO_BASE + (UKVM_HYPERCALL_MAX << 2)))
+                errx(1, "Invalid guest port access: port=0x%llx", run->mmio.phys_addr);
```
mato: As above

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -0,0 +1,137 @@
...
+            /* Guest has halted the CPU, this is considered as a normal exit. */
+            if (nr == UKVM_HYPERCALL_HALT)
+                return;
+
+            ukvm_hypercall_fn_t fn = ukvm_core_hypercalls[nr];
+            if (fn == NULL)
+                errx(1, "Invalid guest hypercall: num=%d", nr);
+
+            ukvm_gpa_t gpa = 0;
+            memcpy(&gpa, run->mmio.data, run->mmio.len);
```
mato: This should be either a direct pointer dereference as in the `x86_64` version of the code, or, if you want to use memcpy() then bound the memcpy on `sizeof gpa`, not `run->mmio.len`. Rationale: Someone will move the code around later and create a buffer overflow by accident.

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -0,0 +1,137 @@
...
+
```
mato: Extra blank line

ukvm/ukvm_hv_kvm.h
```diff
@@ -25,11 +25,20 @@
 #ifndef UKVM_HV_KVM_H
 #define UKVM_HV_KVM_H
 
+#include <linux/kvm.h>
+
```
mato: Why is this needed?

ukvm/ukvm_elf.c
```diff
@@ -197,6 +217,17 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
...
+#ifdef __aarch64__
+        ukvm_region_list->regions[ph_i].slot = ph_i;
+        ukvm_region_list->regions[ph_i].guest_phys_addr = (uint64_t)paddr;
+        ukvm_region_list->regions[ph_i].memory_size = _end - paddr;
+        ukvm_region_list->regions[ph_i].userspace_addr = (uint64_t)daddr;
+        ukvm_region_list->regions[ph_i].flags = (phdr[ph_i].p_flags & PF_X)
+                                                ? KVM_MEM_READONLY
+                                                : 0;
+        ukvm_region_list->count++;
+#endif
```
mato: The ELF loader cannot depend on hypervisor specifics. If I understand the `ukvm_region_list` correctly, it serves two purposes for ARM64:

1. To have a way of creating the MMIO region for hypercalls.
2. To set up the guest page tables to match the ELF program headers, ie. in practice to get a read-only mapping for the first program header (containing `.text`, `.rodata`, `.eh_frame`).

Thinking about simplifying this:

1. Is always at a fixed address, so can be done by an appropriate `#ifdef` in `ukvm_hv_init()`.
2. For x86_64 we cheat: The guest page tables are all RWX and set up in a simple loop. However, we still get a read-only mapping since that's done by the `mprotect()` call in the ELF loader. The only difference is that instead of a guest-side fault on attempting to write to those pages we get an `EFAULT` back from `ioctl(KVM_RUN)` instead.

Could we use the same approach for ARM64, or is there some other need for `ukvm_region_list` that I've missed?

mato:
@Weichen81 I can confirm that c0e6a73 fixes networking, thanks.

kernel/cpu_vectors_aarch64.S
```diff
@@ -0,0 +1,121 @@
+/* Linkage for ARM */
```
Weichen81: Ahh, I missed it, I will fix it in the updated changes.

kernel/cpu_vectors_aarch64.S
```diff
@@ -0,0 +1,121 @@
...
+ * Exception trap types
+ *-----------------
+ */
+#define ET_SYNC    0
+#define ET_IRQ     1
+#define ET_FIQ     2
+#define ET_ERROR   3
+
+/*
+ * Invalid exception handlers
```
Weichen81: That's because, we think all exceptions we get in guest are invalid. For example, If we got a page fault exception in guest. Because we have allocated memory for all pages. This page fault exception is abnormal, so it's invalid, we have to panic the guest.

kernel/cpu_vectors_aarch64.S
```diff
@@ -0,0 +1,121 @@
...
+
+/*
+ * Exception vectors.
+ *
+ * AArch64 unikernel runs in EL1 mode using the SP1 stack. The vectors
+ * don't have a fixed address, only alignment (2^11) requirements.
+ */
+	.pushsection ".exception.text", "ax"
+
+    .align  11
```
Weichen81: We had tried in another way to get the address of this extra link section without the change of linker script. Just like use readelf to dump the vector table symbol from elf image. But we found it's too complex. It's more complex than add a link section in linker scripts.

kernel/ukvm/tscclock.c
```diff
@@ -30,6 +30,12 @@ static uint64_t tsc_base;
 /* Multiplier for converting TSC ticks to nsecs. (0.32) fixed point. */
 static uint32_t tsc_mult;
 
+#ifdef __x86_64__
+#define read_cpu_ticks cpu_rdtsc
+#elif __aarch64__
+#define read_cpu_ticks cpu_cntvct
+#endif
+
```
Weichen81: ok

tests/test_fpu/test_fpu.c
```diff
@@ -43,6 +44,16 @@ int solo5_app_main(char *cmdline __attribute__((unused)))
...
+#else
+    __asm__(
+        "ldr d0, %0\n"
+        "ldr d1, %0\n"
+        "fmul v0.2s, v1.2s, v0.2s\n"
+        "str d0, %0\n"
+        : "=m" (c)
+        : "m" (c)
+    );
+#endif
```
Weichen81: Thanks for your reminder. I will update it.

kernel/cpu_aarch64.h
```diff
@@ -0,0 +1,107 @@
...
+
+    __asm__ __volatile__("mrs %0, cntvct_el0" : "=r" (val)::);
+    return val;
+}
+
+static inline uint64_t mul64_32(uint64_t a, uint32_t b)
+{
+    uint64_t prod;
+
+    __asm__ __volatile__("mul %0, %1, %2" : "=r" (prod) : "r" (a), "r" (b));
```
Weichen81: Yes, it's a normal multiplication. I will change it in updates.

kernel/cpu_aarch64.h
```diff
@@ -0,0 +1,107 @@
...
+
+    return prod;
+}
+
+/* accessing devices via port space */
+
```
Weichen81: Yes, we have defined a single function for hypercaller. These I/O functions for ARM64 are no longer needed. I will remove them in updates.

ukvm/ukvm_elf.c
```diff
@@ -120,15 +129,15 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
...
+     * 4. Target instruction must be set to the correct architecture.
      */
     if (hdr.e_ident[EI_MAG0] != ELFMAG0
             || hdr.e_ident[EI_MAG1] != ELFMAG1
             || hdr.e_ident[EI_MAG2] != ELFMAG2
             || hdr.e_ident[EI_MAG3] != ELFMAG3
             || hdr.e_ident[EI_CLASS] != ELFCLASS64
             || hdr.e_type != ET_EXEC
-            || hdr.e_machine != EM_X86_64)
+            || hdr.e_machine != E_MACHINE_TARGET)
```
Weichen81: OK, that seems better.

ukvm/ukvm_elf.c
Weichen81: Hi Mato,
Your understanding about ukvm_region_list is right.
But the section address and size like .text is not fixed. It will change as we add or remove code.
About the mprotect, we had tried it already. But mapping a read-only VMA is only allowed if the
memory region is configured as read-only. If we set all guest memory to RWX, the mprotect will
failed when we try to make an area be read-only.

ukvm/ukvm_guest.h
```diff
@@ -73,6 +73,43 @@ static inline void ukvm_do_hypercall(int n, volatile void *arg)
 }
 #    endif
 
+#elif __aarch64__
+/*
+ * PIO base address used to dispatch hypercalls.
+ */
+#define UKVM_HYPERCALL_PIO_BASE 0x500
```
Weichen81: Yes. This would be more sensible.

ukvm/ukvm_guest.h
```diff
@@ -73,6 +73,43 @@ static inline void ukvm_do_hypercall(int n, volatile void *arg)
...
+
+#    ifdef UKVM_HOST
+/*
+ * Non-dereferencable monitor-side type representing a guest physical address.
+ */
+typedef uint64_t ukvm_gpa_t;
+#    else
+/*
+ * On aarch64, the PIO address must be 32-bit aligment it we access this
```
Weichen81: Thanks.

ukvm/ukvm_guest.h
```diff
@@ -73,6 +73,43 @@ static inline void ukvm_do_hypercall(int n, volatile void *arg)
...
+ * address in 32-bit. So the real hypercall ID will be calculated as:
+ * UKVM_HYPERCALL_PIO_BASE + (ID << 2).
+ *
+ * This only supports sending 32-bit pointers; raise an assertion if a bigger
+ * pointer is used.
```
Weichen81: Yes, it's not a limitation on ARM64. But I agree to keep it for consistency.

ukvm/ukvm_guest.h
```diff
@@ -73,6 +73,43 @@ static inline void ukvm_do_hypercall(int n, volatile void *arg)
...
+ *
+ * On aarch64 the compiler-only memory barrier ("memory" clobber) is
+ * sufficient across the hypercall boundary.
+ */
+static inline void ukvm_do_hypercall(int n, volatile void *arg)
+{
+#    ifdef assert
+    assert(((uint64_t)arg <= UINT32_MAX));
+#    endif
+
```
Weichen81: Ok. I will remove it.

kernel/cpu_aarch64.c
```diff
@@ -0,0 +1,67 @@
...
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "ukvm/kernel.h"
+
+void intr_init(void)
+{
+
+}
```
Weichen81: Ok

kernel/cpu_aarch64.c
```diff
@@ -0,0 +1,67 @@
...
+static const char *exception_modes[]= {
+        "Synchronous Abort",
+        "IRQ",
+        "FIQ",
+        "Error"
+};
+
+void cpu_trap_handler(int el, int mode)
+{
+    printf("Exception caught in EL%d:%s\n", el, exception_modes[mode]);
```
Weichen81: Yes, that would be more sensible.

ukvm/ukvm_hv_kvm_aarch64.c
Weichen81: Ok

ukvm/ukvm_hv_kvm_aarch64.c
Weichen81: On ARM64, when I use a direct pointer run->mmio.data[8], I always got the gcc "strict aliasing".
In this case, I have to memcpy instead. But your comment is right, I should not use the run->mmio.len directly. I will fix it.

But gpa is a 64-bit data, so the upper 32-bit is random. This is
will cause the UKVM_CHECKED_GPA_P failed in some applications.
In order to address this issue, we reset the gpa to 0 before get
address from run->mmio.data. In this case, the unused 32-bit value
should be set to 0.

ukvm/ukvm_hv_kvm.h
Weichen81: This because the ukvm_region_list contains kvm_user_space_memory_region. And this structure has been defined in linux/kvm.h. Place this include here may break the x86_64 freebsd build. I had sent
a patch to fix it. But I think I better to find a way to remove this include from this header file.

Weichen81:
> Generally this is great, and it looks like the multiarch refactoring is holding up, with the exception of:

>UKVM_HYPERCALL_HALT: I think we should do this for all architectures anyway, I can do that after this PR is merged. The call should accept a struct with an actual exit code and not be a special case in the VCPU loop.

That's better. But if I remove current halt hypercall from ARM64. The test applications could not terminate appropriately. Can we keep it in this PR and revert it when you new halt hypercall is
ready?

> ukvm_region_list won't work as designed. See my comment on the code in the ELF loader.

I will try to avoid using ukvm_region_list.

>So, good stuff. Could you also please add yourself to AUTHORS under "Corporate" and/or "Individual" contributor sections as appropriate?

Thanks. I will do it. And this week I will be in Japan for OSS 2017. I think I have to address the comments
after I am back from the conference.

mato:
> +
> +/*
> + * Invalid exception handlers
> 
> That's because, we think all exceptions we get in guest are invalid. For example, If we got a page fault exception in guest. Because we have allocated memory for all pages. This page fault exception is abnormal, so it's invalid, we have to panic the guest.

Aborting/panicking the guest is indeed what we do for all exceptions.
However, the comment confused me since it reads as "These exception
handlers are invalid". Perhaps just drop "invalid" from the terminology?

mato:
> + *
> + * AArch64 unikernel runs in EL1 mode using the SP1 stack. The vectors
> + * don't have a fixed address, only alignment (2^11) requirements.
> + */
> +	.pushsection ".exception.text", "ax"
> +
> +    .align  11
> 
> We had tried in another way to get the address of this extra link section without the change of linker script. Just like use readelf to dump the vector table symbol from elf image. But we found it's too complex. It's more complex than add a link section in linker scripts.

Wouldn't ".align 2048" be sufficient to get the required alignment without
the need for an extra section?

mato:
> Your understanding about ukvm_region_list is right.
> But the section address and size like .text is not fixed. It will change as we add or remove code.
> About the mprotect, we had tried it already. But mapping a read-only VMA is only allowed if the
> memory region is configured as read-only. If we set all guest memory to RWX, the mprotect will
> failed when we try to make an area be read-only.

If I understand correctly you're saying that the guest-side page tables and
host-side VMA protection bits have to match up exactly? That seems... odd.
Is it an architectural or KVM/ARM limitation of some kind?

Aside: Regarding the general idea of a "region list", I'm not against
having such a thing but it would need to be a hypervisor (KVM, vmm, Hvf)
agnostic API, not something specific to one particular (platform, arch)
combination. Implementing such a generic API is complex so if we can avoid
it for now that would be good.

mato:
> On ARM64, when I use a direct pointer run->mmio.data[8], I always got the gcc "strict aliasing".

What specific pointer access did you try to use? What was the exact
warning?

Something like

gpa = *(uint32_t *)&run->mmio.data[0];

?

I'll have a play with this and also look at what other code does (e.g.
QEMU).

mato:
> That's better. But if I remove current halt hypercall from ARM64. The test applications could not terminate appropriately. Can we keep it in this PR and revert it when you new halt hypercall is
> ready?

Yes, that's [keeping the hypercall in this PR as is] what I meant, sorry if
it wasn't clear.

> Thanks. I will do it. And this week I will be in Japan for OSS 2017. I think I have to address the comments
> after I am back from the conference.

Good luck with the talk and let us know how it goes.

ukvm/ukvm_cpu_aarch64.c
```diff
@@ -0,0 +1,495 @@
...
+        err(1, "KVM: ioctl (SET_USER_MEMORY_REGION) slot=%d failed", used_slot);
+
+    used_slot++;
+    /* Register elf regions */
+    for (idx = 0; idx < ukvm_region_list->count; idx++) {
+        uint64_t va_end, pa_end;
+        elf_region = ukvm_region_list->regions + idx;
+
+        /* Update slot number with used slot before elf regions */
+        elf_region->slot = used_slot;
```
ricarkol: No need for the variable used_slot. You can use idx here instead.

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -0,0 +1,137 @@
...
+    /* Setup aarch64 system registers */
+    ukvm_aarch64_setup_system(hvb->vmfd, hvb->vcpufd);
+
+    /* Setup aarch64 guest vector table */
+    ukvm_aarch64_setup_vectors(hvb->vcpufd, gpa_ep - VBAR_TABLE_LEN);
+
+    /* Setup aarch64 core registers */
+    ukvm_aarch64_setup_core(hv, gpa_ep, gpa_kend);
+
+    /* Free the memeory regions list, we won't use it any more */
```
ricarkol: typo on memeory

ukvm/ukvm_cpu_aarch64.c
```diff
@@ -0,0 +1,495 @@
...
+        ret = ioctl(vmfd, KVM_SET_USER_MEMORY_REGION, elf_region);
```
ricarkol: Is `userspace_addr` set to something like `hv->mem + xxx`? it seems that on the ELF loading side, `userspace_addr` is pointing to the buffer used to read the elf, and that buffer is `hv->mem`. So we might be OK. Although this is getting too complicated. Do we really need the regions?

Weichen81:
>>That's because, we think all exceptions we get in guest are invalid. For example, If we got a page fault exception in guest. Because we have allocated memory for all pages. This page fault exception is abnormal, so it's invalid, we have to panic the guest.

>Aborting/panicking the guest is indeed what we do for all exceptions.
However, the comment confused me since it reads as "These exception
handlers are invalid". Perhaps just drop "invalid" from the terminology? 

Yes, It seems "invalid" not accurate here. I will drop it in updates.

Weichen81:
>>*
>>* AArch64 unikernel runs in EL1 mode using the SP1 stack. The vectors
>>* don't have a fixed address, only alignment (2^11) requirements.
>>*/
>>.pushsection ".exception.text", "ax"
>> .align  11


>> We had tried in another way to get the address of this extra link section without the change of linker script. Just like use readelf to dump the vector table symbol from elf image. But we found it's too complex. It's more complex than add a link section in linker scripts.

>Wouldn't ".align 2048" be sufficient to get the required alignment without
the need for an extra section?

2048 alignment is enough. But without an extra section in linker script, I can't place the vector table at
the beginning of the image. I have to read the vector table symbol from elf image to get the address
of vector table.

ukvm/ukvm_cpu_aarch64.c
Weichen81: @ricarkol Thanks, I will fix it.

ukvm/ukvm_cpu_aarch64.c
Weichen81: Hmm, I need to re-consider it and try to make it be simple. It's best to avoid using regions if possible.

ukvm/ukvm_hv_kvm_aarch64.c
Weichen81: Thanks.

Weichen81:
>> Your understanding about ukvm_region_list is right.
 But the section address and size like .text is not fixed. It will change as we add or remove code.
 About the mprotect, we had tried it already. But mapping a read-only VMA is only allowed if the
 memory region is configured as read-only. If we set all guest memory to RWX, the mprotect will
 failed when we try to make an area be read-only.

>If I understand correctly you're saying that the guest-side page tables and
host-side VMA protection bits have to match up exactly? That seems... odd.
Is it an architectural or KVM/ARM limitation of some kind?

It's the KVM/ARM limitation. I am not sure this limitation still existed in the latest code, 
I will check it later.

>Aside: Regarding the general idea of a "region list", I'm not against
having such a thing but it would need to be a hypervisor (KVM, vmm, Hvf)
agnostic API, not something specific to one particular (platform, arch)
combination. Implementing such a generic API is complex so if we can avoid
it for now that would be good.

Yes, this API should be hypervisor agnostic.  If we could not avoid this "region list", 
 I will refine this API.

Weichen81:
>>Thanks. I will do it. And this week I will be in Japan for OSS 2017. I think I have to address the comments
 after I am back from the conference.

>Good luck with the talk and let us know how it goes.

Thanks,
I just come back from Tokyo today. Both the talk and demo were going well :)
Because my talk is based on ARM64, so most feedbacks are:
1. Will we support ARM32 and i386?
2. The Hardware requirements of running unikernels on uKVM.
3. Can we run this demo on raspberry PI? -->> Of course, we did this demo on NeoPI.
4. The Memory limits of unikernels on uKVM
5. They  were care about the CPU consumption of GUEST MMU translation on ARM64.
6. Some audiences were confused by the name of uKVM. They thought this is a replacement
    of KVM.

mato:
On Tuesday, 06.06.2017 at 07:34, Wei Chen wrote:
> 2048 alignment is enough. But without an extra section in linker script, I can't place the vector table at
> the beginning of the image. I have to read the vector table symbol from elf image to get the address
> of vector table.

I don't follow. Can you point me to the code that uses the address of the
vector table?

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -0,0 +1,137 @@
...
+    struct ukvm_hvb *hvb = hv->b;
+
+    /* Setup aarch64 guest memroy */
+    ukvm_aarch64_setup_memory(hvb->vmfd, hv->mem, 0, hv->mem_size);
+
+    /* Setup aarch64 system registers */
+    ukvm_aarch64_setup_system(hvb->vmfd, hvb->vcpufd);
+
+    /* Setup aarch64 guest vector table */
+    ukvm_aarch64_setup_vectors(hvb->vcpufd, gpa_ep - VBAR_TABLE_LEN);
```
Weichen81: @mato Here we use the vector table address to initialize the ARM VBAR_EL1 register.
gpa_ep point to _start. The linker script make the ld to place the vector table before _start.
So we can use the gpa_ep and vector table reserved size to calculate the start address of vector table.

Weichen81:
@mato 
>> On ARM64, when I use a direct pointer run->mmio.data[8], I always got the gcc "strict aliasing".
What specific pointer access did you try to use? What was the exact
warning?

>Something like
gpa = *(uint32_t *)&run->mmio.data[0];
?
I'll have a play with this and also look at what other code does (e.g.QEMU).

Yes, I had tried to use:
ukvm_gpa_t gpa = 0;
gpa = *(ukvm_gpa_t *)&run->mmio.data[0];

But got following error:
/home/Wei/Merge/arm64-solo5/ukvm/ukvm_hv_kvm_aarch64.c: In function ‘ukvm_hv_vcpu_loop’:
/home/Wei/Merge/arm64-solo5/ukvm/ukvm_hv_kvm_aarch64.c:115:13: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
             gpa = *(ukvm_gpa_t *)&run->mmio.data[0];
             ^

mato:
@Weichen81:
> On ARM64, when I use a direct pointer run->mmio.data[8], I always got the gcc "strict aliasing".

Based on studying what `kvmtool` does, you can do something like this (in `ukvm_hv_kvm_aarch64.c`):
````
static inline uint32_t mmio_read32(void *data)
{
    return *(uint32_t *)data;
}

/* ... in the vcpu_loop ... */

    ukvm_gpa_t gpa = mmio_read32(run->mmio.data);
    
````

This is one of the rare cases where I _think_ the compiler is just getting in the way with its strict aliasing rules and the above can be used safely. I've tested this change on an ARM64 machine and it seems to work but you might want to confirm via disassembly that it's doing the right thing.

mato:
@Weichen81

> @mato Here we use the vector table address to initialize the ARM VBAR_EL1 register.

So, perhaps a naive question but can't the guest do this in `cpu_init()`?
It seems to me to be equivalent to the IDT setup for x86_64 which is also
done on the guest side.

That way we'd need nothing more than the alignment, and the monitor does
not need to use any "magic" to look for the exception table in the guest
binary.

Weichen81:
Regards,
在 2017年6月8日，21:16，Martin Lucina <notifications@github.com<mailto:notifications@github.com>> 写道：

@Weichen81

> @mato Here we use the vector table address to initialize the ARM VBAR_EL1 register.

So, perhaps a naive question but can't the guest do this in `cpu_init()`?
It seems to me to be equivalent to the IDT setup for x86_64 which is also
done on the guest side.


Of course, we can. In the first version before multi-arch refactor. We do this in the guest side. But considering to make the guest arch independent, we move this setting to monitor. It seems I had considered too much :)

That way we'd need nothing more than the alignment, and the monitor does
not need to use any "magic" to look for the exception table in the guest
binary.

Yeah, I agree. Since we can't make guest arch independent in current stage, it's better to move this setting back to guest to keep linker scripts to be simple.

mato:
> Yeah, I agree. Since we can't make guest arch independent in current stage, it's better to move this setting back to guest to keep linker scripts to be simple.

It's not just the linker scripts, those are solvable. I'm more concerned about the monitor needing to fish around in the guest binary. This could be solved by doing symbol lookup but I'd rather not do that at this stage. So yeah, lets do it on the guest side.

Weichen81:
Regards,

在 2017年6月8日，21:12，Martin Lucina <notifications@github.com<mailto:notifications@github.com>> 写道：


@Weichen81<https://github.com/weichen81>:

On ARM64, when I use a direct pointer run->mmio.data[8], I always got the gcc "strict aliasing".

Based on studying what kvmtool does, you can do something like this (in ukvm_hv_kvm_aarch64.c):

static inline uint32_t mmio_read32(void *data)
{
    return *(uint32_t *)data;
}

/* ... in the vcpu_loop ... */

    ukvm_gpa_t gpa = mmio_read32(run->mmio.data);



This is one of the rare cases where I think the compiler is just getting in the way with its strict aliasing rules and the above can be used safely. I've tested this change on an ARM64 machine and it seems to work but you might want to confirm via disassembly that it's doing the right thing.

Yes, this is very rare case. Thank you very much for finding above solution! I will verify it.

Weichen81:
@mato This week I have some urgent task of Xen. So I will resolve the branch conflicts and address above
comments a little later.

Weichen81:
@mato I have re-based my changes to latest solo5 master. If I push those changes to this branch, should I
reopen a PR?

mato:
@Weichen81 Sorry for the delay, I was on holiday. Thanks for the update, I'll look at it over the next few days.

mato:
@Weichen81:

This is looking good. However, most importantly, the guest memory setup still seemed too complex and introduced aarch64-specifics in the ELF loader. Especially the latter is something I'd like to avoid if at all possible, so, I did some experimenting and come up with the following change: https://github.com/mato/solo5/commit/57cb954ec6afb686eeb2b5a6e042c1311e36d584

From the commit message for that change:

----

This change simplifies the guest memory region and page table setup onaarch64 by:

 - Using a single KVM memory region for the entire guest physical memory.
 - Using a 2-level paging scheme with 1GB pages for the guest.

This brings the code in line with what the x86_64 code does, and removes the need to have aarch64-specific code in the ELF loader.

Guest memory protection is _de facto_ unchanged and behaves the same way as on x86_64; writing to a read-only section of the guest kernel (eg. *solo5_app_main) will cause an exit from ukvm_hv_vcpu_loop() with "KVM: host/guest translation fault" due to the underlying mprotect() performed by the ELF loader.

I will add a test case to demonstrate this behaviour in a future change.

----

Let me know what you think. I've tested this on the ARM64 hardware I have available and it operates as expected.

Some further general comments, roughly in order of importance:

1. Can you please put the global symbols defined in `kernel/cpu_vectors_aarch64.S` into the `cpu_` namespace? This follows existing conventions used in the ukvm code.
2. Unrelated to this change, `test_time` fails for me some of the time with "Time is not passing". Can you reproduce the same problem? (Just run the test manually from the command line.)
3. The code split between `ukvm_cpu_aarch64.c` and `ukvm_hv_kvm_aarch64.c` is not quite right. If you look at the x86_64 code as an example, the code in `ukvm_cpu_x86_64.c` is  _independent_ of the underlying hypervisor, so eg. that module does not include any KVM header files. Code which is arch- and hypervisor- dependent is in `ukvm_hv_kvm_<...>.c`. This should be just some simple shuffling of functions around.
4. `ukvm_cpu_aarch64.h` has a _lot_ of constants in it, most of which are not used anywhere. It would be nice to trim this down somewhat, I'm sure there's stuff in there which ukvm will not need in the medium term.

Weichen81:
@mato mato
>This is looking good. However, most importantly, the guest memory setup still seemed too complex and introduced aarch64-specifics in the ELF loader. Especially the latter is something I'd like to avoid if at all possible, so, I did some experimenting and come up with the following change: mato/solo5@57cb954

>From the commit message for that change:

>This change simplifies the guest memory region and page table setup onaarch64 by:

>Using a single KVM memory region for the entire guest physical memory.
Using a 2-level paging scheme with 1GB pages for the guest.
This brings the code in line with what the x86_64 code does, and removes the need to have aarch64-specific code in the ELF loader.

I think this change is good. Why we separated guest memory to so many ranges was a limit
in Linux ARM KVM API (see: liunx/virt/kvm/arm/mmu.c: kvm_arch_prepare_memory_region).
Before multi-arch refactor, we called mprotect before  KVM_SET_USER_MEMORY_REGION ioctl.
This would trigger following restriction.
		/*
		 * Mapping a read-only VMA is only allowed if the
		 * memory region is configured as read-only.
		 */
		if (writable && !(vma->vm_flags & VM_WRITE)) {
			ret = -EPERM;
			break;
		}
But after multi-arch refactor, we had changed the order of KVM_SET_USER_MEMORY_REGION ioctl
and mprotect. This would avoid this restriction.

So the simplify is OK. I have tested it on my ARM64 server. I am sorry not to observe this order changed
while I was updating the changes.

>Guest memory protection is de facto unchanged and behaves the same way as on x86_64; writing to a read-only section of the guest kernel (eg. *solo5_app_main) will cause an exit from ukvm_hv_vcpu_loop() with "KVM: host/guest translation fault" due to the underlying mprotect() performed by the ELF loader.
I will add a test case to demonstrate this behaviour in a future change.

>Let me know what you think. I've tested this on the ARM64 hardware I have available and it operates as expected.

I had done a simple test, it works as expected too :)
ukvm-bin: KVM: host/guest translation fault: pc=0x105d44

>Some further general comments, roughly in order of importance:

>Can you please put the global symbols defined in kernel/cpu_vectors_aarch64.S into the cpu_ namespace? This follows existing conventions used in the ukvm code.

Ok, it's my mistake, I will do it to follow the existing conventions.

>Unrelated to this change, test_time fails for me some of the time with "Time is not passing". Can you reproduce the same problem? (Just run the test manually from the command line.)

Yes, I have reproduced it:
**** Solo5 standalone test_time ****

ta:4fa80 tb:4fa80
ERROR: time is not passing

It seems two solo5_clock_monotonic() return the same value. It found the GCC compiled the c=a*b to MADD instruction not MUL instruction we had used before.  But I think it would not be the problem,
I need do some further investigation to find the root reason.

>The code split between ukvm_cpu_aarch64.c and ukvm_hv_kvm_aarch64.c is not quite right. If you look at the x86_64 code as an example, the code in ukvm_cpu_x86_64.c is independent of the underlying hypervisor, so eg. that module does not include any KVM header files. Code which is arch- and hypervisor- dependent is in ukvm_hv_kvm_<...>.c. This should be just some simple shuffling of functions around.

Yes, that splits are not quite clear. I will move KVM related code to ukvm_hv_kvm_aarch64.c

>ukvm_cpu_aarch64.h has a lot of constants in it, most of which are not used anywhere. It would be nice to trim this down somewhat, I'm sure there's stuff in there which ukvm will not need in the medium term.

Yes, they are useless in current stage, I will do a purge :)

Weichen81:
@mato 
I looked into the code of tscclock and test_time.
I objdump the tscclock_monotonic:
  7 0000000000000000 <tscclock_monotonic>:
  8    0:   d53be042        mrs     x2, cntvct_el0
  9    4:   90000004        adrp    x4, 0 <tscclock_monotonic>
 10    8:   91000083        add     x3, x4, #0x0
 11    c:   f9400080        ldr     x0, [x4]
 12   10:   b9400861        ldr     w1, [x3,#8]
 13   14:   cb000040        sub     x0, x2, x0
 14   18:   9b017c00        mul     x0, x0, x1
 15   1c:   f9400861        ldr     x1, [x3,#16]
 16   20:   f9000082        str     x2, [x4]
 17   24:   8b010000        add     x0, x0, x1
 18   28:   f9000860        str     x0, [x3,#16]
 19   2c:   d65f03c0        ret

Call solo5_clock_monotonic two times sequentially will use 24 instructions. If we count until the
CPU save the multiplication result to x0 in secondary call, it will use 19 instructions.

I have a CPU will 500MIPS, and the timer of ARM64 is 250Mhz, the resolution is 4ns.
4ns is enough for about 20 instructions.  So it's possible that when we call solo5_clock_monotonic
secondly, the timer had changed its value.

Can we change the test_time to following?
    /*
     * Verify that monotonic time is passing
     */
    uint64_t ta, tb;
    ta = solo5_clock_monotonic();
    tb = solo5_clock_monotonic();
    if (!(tb >= ta)) { ===>>>> Cover the equal case, or add a delay to two calls?
        solo5_console_dbg(ta, tb);
        puts("ERROR: time is not passing\n");
        return 1;
    }

mato:
@Weichen81 

Thanks for tracking down the cause of the `test_time` failure. I've addressed it in https://github.com/mato/solo5/commit/402e296ec7a10aa20caa45c47063b1cd53ee4a9c. We don't want to use a poll-style delay for that test since the test should verify that time is passing independently of any exits to the monitor.

Weichen81:
@mato 
I have updated the changes to address the comments of last version. I delete the branch by mistake. So I have to reopen the PR.

Regards,

mato:
@Weichen81 The latest set of changes you pushed looks wrong, there are a bunch of commits that are already on `master`, and I don't see the new changes. Also, you appear to have merged `master` into your branch.

Can you please try and rebase a clean set of changes onto `master`, then force-push your branch? You can do that without the need to re-create the branch or PR. (But, as always, perform the rebase on a copy of your git clone if you're not sure what the results will be.)

Weichen81:
@mato 
Sorry, it seems I have done some mistakes while rebasing the changes. I think it's better to close this PR and create a new one.

mato:
You don't need to close the PR, just force-push the branch again with the correct set of commits.

Weichen81:
Ok, I will try to update the branch and reopen this PR.

Weichen81:
@mato 
This PR could not be reopened any more, the github hint is "arm64-refactor" has been force-updated or recreated. I had deleted the remote branch before, so I can't be reopened.
I have created a new PR https://github.com/Solo5/solo5/pull/209
This one seems has correct commits set.

Regards,

PR #196, opened 2017-06-04
ukvm/console: return actual number of bytes written

hannesm:
instead of the string length

### DIFF
--- a/kernel/ukvm/console.c
+++ b/kernel/ukvm/console.c
@@ -29,7 +29,7 @@ int platform_puts(const char *buf, int n)
 
     ukvm_do_hypercall(UKVM_HYPERCALL_PUTS, &str);
 
-    return str.len;
+    return str.ret;
 }
 
 int solo5_console_write(const char *, size_t)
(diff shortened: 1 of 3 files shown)


hannesm:
this addresses https://github.com/mirage/mirage-flow/issues/4#issuecomment-306041566 and comes along with https://github.com/mirage/mirage-solo5/pull/19 and https://github.com/mirage/mirage-console-solo5/pull/12 (but this change in here is very independent, and does not change any API)

mato:
There are multiple issues with the `solo5.h` APIs, I'm working on a minimal refresh that corrects the most egregious problems without actually redesigning anything.

In the specific case of the console, I'm going to change `solo5_console_write()` to just return `void`; I don't see any point in trying to "recover" from failed console writes.

PR coming up once it's ready, we can discuss the changes there.

hannesm:
my motivation was to unify the semantics of console writes across backends, please see https://github.com/mirage/mirage-flow/issues/4#issuecomment-306041566 for more information.  but i agree that error recovery from partial writes is slightly overengineering -- but I'd still like to have a unified semantics here (as in: writing to the console may fail or be partial), and properly documented error cases.

hannesm:
superseeded by https://github.com/Solo5/solo5/pull/201

Issue #176, opened 2017-04-04
Refactor/rewrite ukvm gdb module


mato:
The GDB module code is hacked together from various sources and is generally in a dire state. It should be refactored (probably implies a rewrite) into two separate parts:

1. The GDB server and protocol code, ideally arch- and backend- independent and engineered from an actual protocol description (or reading the GDB source code, no idea how much documentation exists for the protocol).
2. Initially an x86_64 (KVM ...) backend implementation since that's what is there today.

Ideally as part of the refactoring the 1st part would be arch- and backend- independent (as in the changes in #171), then the 2nd part could gain features like using hardware breakpoints rather than single-stepping, plus fix various other outstanding issues (#16, #25).

Help wanted. Any takers? This is an extremely useful and fairly self-contained project.

ricarkol:
Will start working on this. Does anybody want to collaborate?

Issue #25, opened 2016-05-31
The ukvm GDB stub is not sending the appropriate terminate message when exiting


ricarkol:
When ukvm runs in --gdb mode and exits, the client complains with this:

```
A problem internal to GDB has been detected,                                                                                                                 
further debugging may prove unreliable.                                                                                                                      
Quit this debugging session? (y or n) Quit   
```

The fix is to catch SIGTERM and other signals and send the appropriate message to the gdb client.

mato:
Fixed in #198.

PR #198, opened 2017-06-05
Cleanup of the GDB server code in ukvm

ricarkol:
Changes:
1. Removed all arch and backend dependent code from ukvm/ukvm_module_gdb.c
2. Cleaned up ukvm/ukvm_module_gdb.c, fixed issue #25.
3. Implemented some new GDB commands like: writing memory and registers.
4. Added an empty stub for freebsd (just so freebsd builds successfully).
5. Breakpoints without relying on stepping: hardware and software breakpoints #16.

This PR fixes: #176  #16  #25 

Tests:
- tested without mirage with this:
    https://gist.github.com/ricarkol/3e9768639c1e52b86c195d579d818279#file-gdb-test-sh
- tested breakpointing and stepping over some mirage apps. Used this pin:
    `opam pin add mirage git://github.com/ricarkol/mirage#gdb-ukvm`
- built on freebsd (didn't run ukvm)

### DIFF
--- a/kernel/queue.h
+++ b/kernel/queue.h
@@ -82,7 +82,7 @@
 
 /* XXX: Consider moving this to some common libraries directory for solo5 and
  *      ukvm. */
-#ifndef UKVM_HOST
+#ifdef __SOLO5_KERNEL__
 #    include "kernel.h"
 #endif
 
--- a/kernel/queue.h
+++ b/kernel/queue.h
@@ -80,7 +80,11 @@
  * For details on the use of these macros, see the queue(3) manual page.
  */
 
-#include "kernel.h"
+/* XXX: Consider moving this to some common libraries directory for solo5 and
+ *      ukvm. */
+#ifndef UKVM_HOST
+#    include "kernel.h"
+#endif
 
 /*
  * Singly-linked List definitions.
--- a/tests/test_hello/GNUmakefile
+++ b/tests/test_hello/GNUmakefile
@@ -19,6 +19,6 @@
 UKVM_TARGETS=test_hello.ukvm ukvm-bin
 VIRTIO_TARGETS=test_hello.virtio
 MUEN_TARGETS=test_hello.muen
-UKVM_MODULES=gdb # Just check that it builds.
+UKVM_MODULES=
 
 include ../Makefile.tests
--- a/tests/test_hello/GNUmakefile
+++ b/tests/test_hello/GNUmakefile
@@ -19,6 +19,6 @@
 UKVM_TARGETS=test_hello.ukvm ukvm-bin
 VIRTIO_TARGETS=test_hello.virtio
 MUEN_TARGETS=test_hello.muen
-UKVM_MODULES=
+UKVM_MODULES=gdb # Just check that it builds.
 
 include ../Makefile.tests
--- a/ukvm/ukvm.h
+++ b/ukvm/ukvm.h
@@ -146,14 +146,51 @@ extern struct ukvm_module ukvm_module_gdb;
  * GDB specific functions to be implemented on all backends for all
  * architectures.
  */
-int ukvm_gdb_read_registers(struct ukvm_hv *hv, uint8_t *reg, uint64_t *len);
-int ukvm_gdb_write_registers(struct ukvm_hv *hv, uint8_t *reg, uint64_t len);
+
+/*
+ * Fills *registers with a stream of hexadecimal digits for each register in
+ * GDB register order, where each register is in target endian order.
+ * Returns 0 if success, -1 otherwise.
+ */
+int ukvm_gdb_read_registers(struct ukvm_hv *hv, uint8_t *reg, size_t *len);
+
+/*
+ * Writes the shadow registers from a stream of hexadecimal digits for each register in
+ * GDB register order, where each register is in target endian order.
+ * Returns 0 if success, -1 otherwise.
+ */
+int ukvm_gdb_write_registers(struct ukvm_hv *hv, uint8_t *reg, size_t len);
+
+/*
+ * Enagle single stepping. Returns 0 if success, -1 otherwise.
+ */
 int ukvm_gdb_enable_ss(struct ukvm_hv *hv);
+
+/*
+ * Disable single stepping. Returns 0 if success, -1 otherwise.
+ */
 int ukvm_gdb_disable_ss(struct ukvm_hv *hv);
+
+/*
+ * Reads the current KVM exit code and maps it to a GDB signal value.
+ * Returns 0 if success, -1 otherwise.
+ */
 int ukvm_gdb_read_last_signal(struct ukvm_hv *hv, int *signal);
+
+/*
+ * Add a breakpoint of type software or hardware, at address addr.  len is
+ * typically the size of the breakpoint in bytes that should be inserted
+ * Returns 0 if success, -1 otherwise.
+ */
 int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
-                            uint64_t addr, uint32_t len);
+                            ukvm_gpa_t addr, size_t len);
+
+/*
+ * Remove a breakpoint of type software or hardware, at address addr.  len is
+ * typically the size of the breakpoint in bytes that should be inserted.
+ * Returns 0 if success, -1 otherwise.
+ */
 int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, uint32_t type,
-                               uint64_t addr, uint32_t len);
+                               ukvm_gpa_t addr, size_t len);
 
 #endif /* UKVM_H */
--- a/ukvm/ukvm.h
+++ b/ukvm/ukvm.h
@@ -142,4 +142,18 @@ extern struct ukvm_module ukvm_module_blk;
 extern struct ukvm_module ukvm_module_net;
 extern struct ukvm_module ukvm_module_gdb;
 
+/*
+ * GDB specific functions to be implemented on all backends for all
+ * architectures.
+ */
+int ukvm_gdb_read_registers(struct ukvm_hv *hv, uint8_t *reg, uint64_t *len);
+int ukvm_gdb_write_registers(struct ukvm_hv *hv, uint8_t *reg, uint64_t len);
+int ukvm_gdb_enable_ss(struct ukvm_hv *hv);
+int ukvm_gdb_disable_ss(struct ukvm_hv *hv);
+int ukvm_gdb_read_last_signal(struct ukvm_hv *hv, int *signal);
+int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
+                            uint64_t addr, uint32_t len);
+int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, uint32_t type,
+                               uint64_t addr, uint32_t len);
+
 #endif /* UKVM_H */
--- a/ukvm/ukvm.h
+++ b/ukvm/ukvm.h
@@ -34,6 +34,7 @@
 #include "ukvm_cc.h"
 #define UKVM_HOST
 #include "ukvm_guest.h"
+#include "ukvm_gdb.h"
 
 /*
  * Hypervisor {arch,backend}-independent data is defined here.
@@ -148,21 +149,22 @@ extern struct ukvm_module ukvm_module_gdb;
  */
 
 /*
- * Fills *registers with a stream of hexadecimal digits for each register in
- * GDB register order, where each register is in target endian order.
+ * Fills *reg with a stream of hexadecimal digits for each guest register
+ * in GDB register order, where each register is in target endian order.
  * Returns 0 if success, -1 otherwise.
  */
 int ukvm_gdb_read_registers(struct ukvm_hv *hv, uint8_t *reg, size_t *len);
 
 /*
- * Writes the shadow registers from a stream of hexadecimal digits for each register in
- * GDB register order, where each register is in target endian order.
+ * Writes all guest registers from a stream of hexadecimal digits for each
+ * register in *reg. Each register in *reg is in GDB register order, and in
+ * target endian order.
  * Returns 0 if success, -1 otherwise.
  */
 int ukvm_gdb_write_registers(struct ukvm_hv *hv, uint8_t *reg, size_t len);
 
 /*
- * Enagle single stepping. Returns 0 if success, -1 otherwise.
+ * Enable single stepping. Returns 0 if success, -1 otherwise.
  */
 int ukvm_gdb_enable_ss(struct ukvm_hv *hv);
 
@@ -182,7 +184,7 @@ int ukvm_gdb_read_last_signal(struct ukvm_hv *hv, int *signal);
  * typically the size of the breakpoint in bytes that should be inserted
  * Returns 0 if success, -1 otherwise.
  */
-int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
+int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, gdb_breakpoint_type type,
                             ukvm_gpa_t addr, size_t len);
 
 /*
@@ -190,7 +192,7 @@ int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
  * typically the size of the breakpoint in bytes that should be inserted.
  * Returns 0 if success, -1 otherwise.
  */
-int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, uint32_t type,
+int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, gdb_breakpoint_type type,
                                ukvm_gpa_t addr, size_t len);
 
 #endif /* UKVM_H */
--- a/ukvm/ukvm.h
+++ b/ukvm/ukvm.h
@@ -149,6 +149,13 @@ extern struct ukvm_module ukvm_module_gdb;
  */
 
 /*
+ * Check if backend implements GDB support. Returns 0 if supported, -1 if not.
+ * Note backends not implementing GDB support still need to implement the
+ * remaining functions in this section, all of which should return -1.
+ */
+int ukvm_gdb_supported(void);
+
+/*
  * Fills *reg with a stream of hexadecimal digits for each guest register
  * in GDB register order, where each register is in target endian order.
  * Returns 0 if success, -1 otherwise.
--- a/ukvm/ukvm_elf.c
+++ b/ukvm/ukvm_elf.c
@@ -196,11 +196,7 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
         prot = PROT_NONE;
         if (phdr[ph_i].p_flags & PF_R)
             prot |= PROT_READ;
-/* GDB clients can change memory, and software breakpoints work by replacing
- * instructions with int3's. */
-#ifndef UKVM_MODULE_GDB
         if (phdr[ph_i].p_flags & PF_W)
-#endif
             prot |= PROT_WRITE;
         if (phdr[ph_i].p_flags & PF_X)
             prot |= PROT_EXEC;
--- a/ukvm/ukvm_elf.c
+++ b/ukvm/ukvm_elf.c
@@ -196,7 +196,11 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
         prot = PROT_NONE;
         if (phdr[ph_i].p_flags & PF_R)
             prot |= PROT_READ;
+/* GDB clients can change memory, and software breakpoints work by replacing
+ * instructions with int3's. */
+#ifndef UKVM_MODULE_GDB
         if (phdr[ph_i].p_flags & PF_W)
+#endif
             prot |= PROT_WRITE;
         if (phdr[ph_i].p_flags & PF_X)
             prot |= PROT_EXEC;
--- a/ukvm/ukvm_gdb.h
+++ b/ukvm/ukvm_gdb.h
@@ -1,0 +1,40 @@
+/* 
+ * Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file
+ *
+ * This file is part of ukvm, a unikernel monitor.
+ *
+ * Permission to use, copy, modify, and/or distribute this software
+ * for any purpose with or without fee is hereby granted, provided
+ * that the above copyright notice and this permission notice appear
+ * in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef UKVM_GDB_H
+#define UKVM_GDB_H
+
+/* GDB breakpoint/watchpoint types */
+#define GDB_BREAKPOINT_SW        0
+#define GDB_BREAKPOINT_HW        1
+#define GDB_WATCHPOINT_WRITE     2
+#define GDB_WATCHPOINT_READ      3
+#define GDB_WATCHPOINT_ACCESS    4
+
+#define GDB_SIGNAL_FIRST         0
+#define GDB_SIGNAL_QUIT          3
+#define GDB_SIGNAL_KILL          9
+#define GDB_SIGNAL_TRAP          5
+#define GDB_SIGNAL_SEGV          11
+#define GDB_SIGNAL_TERM          15
+#define GDB_SIGNAL_IO            23
+#define GDB_SIGNAL_DEFAULT       144
+
+#endif /* #ifndef UKVM_GDB_H */
--- a/ukvm/ukvm_gdb.h
+++ b/ukvm/ukvm_gdb.h
@@ -22,11 +22,16 @@
 #define UKVM_GDB_H
 
 /* GDB breakpoint/watchpoint types */
-#define GDB_BREAKPOINT_SW        0
-#define GDB_BREAKPOINT_HW        1
-#define GDB_WATCHPOINT_WRITE     2
-#define GDB_WATCHPOINT_READ      3
-#define GDB_WATCHPOINT_ACCESS    4
+typedef enum _gdb_breakpoint_type {
+    /* Do not change these. The values have to match on the GDB client
+     * side. */
+    GDB_BREAKPOINT_SW = 0,
+    GDB_BREAKPOINT_HW,
+    GDB_WATCHPOINT_WRITE,
+    GDB_WATCHPOINT_READ,
+    GDB_WATCHPOINT_ACCESS,
+    GDB_BREAKPOINT_MAX
+} gdb_breakpoint_type;
 
 #define GDB_SIGNAL_FIRST         0
 #define GDB_SIGNAL_QUIT          3
--- a/ukvm/ukvm_gdb_freebsd_x86_64.c
+++ b/ukvm/ukvm_gdb_freebsd_x86_64.c
@@ -1,0 +1,80 @@
+/* 
+ * Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file
+ *
+ * This file is part of ukvm, a unikernel monitor.
+ *
+ * Permission to use, copy, modify, and/or distribute this software
+ * for any purpose with or without fee is hereby granted, provided
+ * that the above copyright notice and this permission notice appear
+ * in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * ukvm_gdb_freebsd_x86_64.c: glue between the GDB server (at ukvm_modules_gdb.c)
+ * and FreeBSD's vmm.
+ * XXX: this file is linked in even if we don't use the GDB module.
+ */
+
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#include <machine/vmm.h>
+#include <sys/param.h>
+#include <sys/cpuset.h>
+#include <machine/vmm_dev.h>
+
+#include "ukvm.h"
+#include "ukvm_hv_freebsd.h"
+#include "ukvm_gdb_x86_64.h"
+#include "ukvm_gdb.h"
+
+int ukvm_gdb_read_registers(struct ukvm_hv *hv,
+                            uint8_t *registers,
+                            uint64_t *len)
+{
+    return -1;
+}
+
+int ukvm_gdb_write_registers(struct ukvm_hv *hv,
+                             uint8_t *registers,
+                             uint64_t len)
+{
+    return -1;
+}
+
+int ukvm_gdb_enable_ss(struct ukvm_hv *hv)
+{
+    return -1;
+}
+
+int ukvm_gdb_disable_ss(struct ukvm_hv *hv)
+{
+    return -1;
+}
+
+int ukvm_gdb_read_last_signal(struct ukvm_hv *hv, int *signal)
+{
+    return -1;
+}
+
+int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
+                            uint64_t addr, uint32_t len)
+{
+    return -1;
+}
+
+int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, uint32_t type,
+                               uint64_t addr, uint32_t len)
+{
+    return -1;
+}
--- a/ukvm/ukvm_gdb_freebsd_x86_64.c
+++ b/ukvm/ukvm_gdb_freebsd_x86_64.c
@@ -21,7 +21,6 @@
 /*
  * ukvm_gdb_freebsd_x86_64.c: glue between the GDB server (at ukvm_modules_gdb.c)
  * and FreeBSD's vmm.
- * XXX: this file is linked in even if we don't use the GDB module.
  */
 
 #include <sys/ioctl.h>
@@ -40,14 +39,14 @@
 
 int ukvm_gdb_read_registers(struct ukvm_hv *hv,
                             uint8_t *registers,
-                            uint64_t *len)
+                            size_t *len)
 {
     return -1;
 }
 
 int ukvm_gdb_write_registers(struct ukvm_hv *hv,
                              uint8_t *registers,
-                             uint64_t len)
+                             size_t len)
 {
     return -1;
 }
@@ -68,13 +67,13 @@ int ukvm_gdb_read_last_signal(struct ukvm_hv *hv, int *signal)
 }
 
 int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
-                            uint64_t addr, uint32_t len)
+                            ukvm_gpa_t addr, size_t len)
 {
     return -1;
 }
 
 int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, uint32_t type,
-                               uint64_t addr, uint32_t len)
+                               ukvm_gpa_t addr, size_t len)
 {
     return -1;
 }
--- a/ukvm/ukvm_gdb_freebsd_x86_64.c
+++ b/ukvm/ukvm_gdb_freebsd_x86_64.c
@@ -37,6 +37,11 @@
 #include "ukvm_gdb_x86_64.h"
 #include "ukvm_gdb.h"
 
+int ukvm_gdb_supported(void)
+{
+    return -1;
+}
+
 int ukvm_gdb_read_registers(struct ukvm_hv *hv,
                             uint8_t *registers,
                             size_t *len)
--- a/ukvm/ukvm_gdb_kvm_x86_64.c
+++ b/ukvm/ukvm_gdb_kvm_x86_64.c
@@ -52,7 +52,7 @@
 #include "queue.h"
 
 struct breakpoint_t {
-    uint32_t type;
+    gdb_breakpoint_type type;
     ukvm_gpa_t addr;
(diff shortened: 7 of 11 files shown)


ukvm/gdb_signals.h
```diff
@@ -0,0 +1,65 @@
+/* Target signal numbers for GDB and the GDB remote protocol.
```
mato: This file and `gdb_signals.def` can't go in verbatim as it has GPL headers on it.

Can you rewrite the file and/or just add the specific `#defines` we need in a suitable place (e.g. `ukvm_module_gdb.c`)?

ricarkol:
Thanks mato, updated the branch with the suggested changes.

mato:
(approved but don't merge yet, more comments forthcoming)

ukvm/ukvm.h
```diff
@@ -142,4 +142,18 @@ extern struct ukvm_module ukvm_module_blk;
 extern struct ukvm_module ukvm_module_net;
 extern struct ukvm_module ukvm_module_gdb;
 
+/*
+ * GDB specific functions to be implemented on all backends for all
+ * architectures.
+ */
```
mato: Could you document these functions similarly to how the other public functions are documented in this file?

Also, please use `ukvm_gpa_t` for addresses (as opposed to `uint64_t`) and `size_t` for the various `len` parameters which are a byte size (as opposed to `uint32_t` etc).

ukvm/ukvm_elf.c
```diff
@@ -196,7 +196,11 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
         prot = PROT_NONE;
         if (phdr[ph_i].p_flags & PF_R)
             prot |= PROT_READ;
+/* GDB clients can change memory, and software breakpoints work by replacing
+ * instructions with int3's. */
+#ifndef UKVM_MODULE_GDB
         if (phdr[ph_i].p_flags & PF_W)
+#endif
```
mato: There's a better way to do this. Rather than cluttering the code here, drop this change and add the following to `setup()` in the GDB module just before the call to `gdb_stub_start()`:
````
    if (mprotect(hv->mem, hv->mem_size, PROT_READ | PROT_WRITE | PROT_EXEC) == -1)
        err(1, "GDB: Cannot remove guest memory protection");
````
For reference, calling `mprotect()` more than once on the same or enclosing region is explicitly allowed, at least on Linux, precisely for this purpose.

File: ukvm/ukvm_gdb_freebsd_x86_64.c [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * ukvm_gdb_freebsd_x86_64.c: glue between the GDB server (at ukvm_modules_gdb.c)
+ * and FreeBSD's vmm.
+ * XXX: this file is linked in even if we don't use the GDB module.
```
mato: One way to fix this would be to `#include` the implementation (this file, KVM equivalent, etc.) using the appropriate chain of `#ifdefs` in `ukvm_module_gdb.c`.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,435 @@
+/* 
+ * Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file
+ *
+ * This file is part of ukvm, a unikernel monitor.
+ *
```
mato: Nit: To keep the license headers standardised, can you please start the file with the standard header used everywhere else and move the "based on XXXX and its licenses" to a separate comment following the header.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,435 @@
...
+
+/* Stepping is disabled by default. */
+static uint32_t stepping = 0;
+/* This is the trap instruction used for software breakpoints. */
+static const uint8_t int3 = 0xcc;
+
+static int kvm_arch_insert_sw_breakpoint(struct ukvm_hv *hv, struct breakpoint_t *bp)
+{
+    bp->saved_insn = *((uint8_t *)hv->mem + bp->addr);
+    /* It was already checked that addr is good. */
```
mato: Where?

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,435 @@
...
+    *((uint8_t *)hv->mem + bp->addr) = int3;
+    return 0;
+}
+
+static int kvm_arch_remove_sw_breakpoint(struct ukvm_hv *hv, struct breakpoint_t *bp)
+{
+    assert((uint8_t) *((uint8_t *)hv->mem + bp->addr) == int3);
+    /* It was already checked that addr is good. */
```
mato: Where?

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,435 @@
...
+    int n = 0;
+
+    if (stepping)
+        dbg.control = KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_SINGLESTEP;
+
+    if (!SLIST_EMPTY(&sw_breakpoints)) {
+        dbg.control |= KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_SW_BP;
+    } else {
+        dbg.control |= KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_HW_BP;
+        /* Not documented magic numbers from QEMU */
```
mato: Uh-huh. Maybe this is DRx and the magic numbers can be found in the Intel SDM?

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,435 @@
...
+#include "ukvm.h"
+#include "ukvm_hv_kvm.h"
+#include "ukvm_gdb_x86_64.h"
+#include "ukvm_gdb.h"
+#include "queue.h"
+
+struct breakpoint_t {
+    uint32_t type;
+    uint64_t addr;
+    uint32_t len;
```
mato: What is `len`? The length of the instruction at the breakpoint address?

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,435 @@
...
+SLIST_HEAD(breakpoints_head, breakpoint_t);
+static struct breakpoints_head sw_breakpoints;
+static struct breakpoints_head hw_breakpoints;
+
+/* Not really documented. But, QEMU uses a hardcoded max of 4. */
+#define MAX_HW_BREAKPOINTS             4
+static uint32_t nr_hw_breakpoints = 0;
+
+/* Stepping is disabled by default. */
+static uint32_t stepping = 0;
```
mato: -> `static bool`

ukvm/ukvm_module_gdb.c
```diff
@@ -3,6 +3,9 @@
  *
  * This file is part of ukvm, a unikernel monitor.
  *
+ * Based on binutils-gdb/gdb/stubs/i386-stub.c, which is:
+ * Not copyrighted.
+ *
```
mato: Again, please keep the standard license header and move this into the next comment.

ukvm/ukvm_module_gdb.c
```diff
@@ -132,195 +41,164 @@
...
+        return (ch - 'A' + 10);
+    return -1;
+}
 
-static void wait_for_connect(int portn)
+/*
+ * Converts the memory pointed to by mem into an hex string in buf.
+ * Returns a pointer to the last char put in buf (null).
+ */
+static char *mem2hex(char *mem, char *buf, int count)
```
mato: Nit: Can the types here be `unsigned char` and `size_t`?

ukvm/ukvm_module_gdb.c
```diff
@@ -132,195 +41,164 @@
...
 {
-    struct sockaddr_in sockaddr;
-    socklen_t sockaddr_len;
-    struct protoent *protoent;
-    int r;
-    int opt;
+    int i;
+    uint8_t ch;
```
mato: ... which would also make these `size_t` and `unsigned char`.

ukvm/ukvm_module_gdb.c
```diff
@@ -132,195 +41,164 @@
...
+    *buf = 0;
+    return buf;
+}
 
-    printf("GDB trying to get a connection at port %d\n", portn);
+/*
+ * Converts the hex string in buf into binary in mem.
+ * Returns a pointer to the character AFTER the last byte written.
+ */
+static char *hex2mem(char *buf, char *mem, int count)
```
mato: Nit: Types as for `mem2hex()`?

ukvm/ukvm_module_gdb.c
```diff
@@ -132,195 +41,164 @@
...
+    struct sockaddr_in server_addr, client_addr;
+    struct protoent *protoent;
+    struct in_addr ip_addr;
+    socklen_t len;
+    int opt;
 
-    /* Allow rapid reuse of this port */
+    listen_socket_fd = socket(AF_INET, SOCK_STREAM, 0);
+    if (listen_socket_fd == -1) {
+        perror("GDB: Could not create socket");
```
mato: Can you use the `err.h` family of functions for consistency?

ukvm/ukvm_module_gdb.c
```diff
@@ -132,195 +41,164 @@
...
+     
+    if (bind(listen_socket_fd, (struct sockaddr *)&server_addr,
+             sizeof(server_addr)) < 0) {
+        perror("GDB: bind failed");
+        return -1;
+    }
+     
+    listen(listen_socket_fd , 0);
+     
+    puts("GDB: Waiting for a debugger...");
```
mato: Use `warnx()` so that this is prefixed with the `ukvm` binary and goes to standard error, not standard output (which may be redirecting the guest console). Also, the message should mention the port it's expecting the connection on. Maybe suggest to the user the command to use? (`target remote :port`)

kernel/queue.h
```diff
@@ -80,7 +80,11 @@
  * For details on the use of these macros, see the queue(3) manual page.
  */
 
-#include "kernel.h"
+/* XXX: Consider moving this to some common libraries directory for solo5 and
+ *      ukvm. */
+#ifndef UKVM_HOST
```
mato: There's also a `__SOLO5_KERNEL__` (see kernel/lib.c), maybe use that instead?

ukvm/ukvm.h
ricarkol: Added and changed types to `ukvm_gpa_t` and `size_t`.

ukvm/ukvm_elf.c
ricarkol: Added the mprotect on the gdb side.

ukvm/ukvm_gdb_freebsd_x86_64.c
ricarkol: Done.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Moved.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Yes, it's the length of the instruction.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: changed.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: In ukvm_module_gdb.c, but added a check here as well.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: In ukvm_module_gdb.c, but added a check here as well.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Yes, they were in the Intel SDM. Added some comments describing them.

ukvm/ukvm_module_gdb.c
ricarkol: Moved.

ukvm/ukvm_module_gdb.c
ricarkol: Changed.

ukvm/ukvm_module_gdb.c
ricarkol: Also changed.

ukvm/ukvm_module_gdb.c
ricarkol: Changed

ukvm/ukvm_module_gdb.c
ricarkol: Using warnx and errx now.

ukvm/ukvm_module_gdb.c
ricarkol: Changed and added a suggested command for attaching to this gdb server.

kernel/queue.h
ricarkol: Changed.

ricarkol:
Thanks for the review mato. Addressed all comments and pushed a new commit.

ukvm/ukvm_module_gdb.c
```diff
@@ -341,30 +401,32 @@ static void gdb_handle_exception(struct ukvm_hv *hv, int sigval)
...
                 (result > hv->mem_size)) {
-                send_packet(GDB_ERROR_MSG);
+		/* Don't panic about this, just return error so the debugger
+                 * tries again. */
+                send_error_msg();
             } else {
-                hex2mem(obuf, (char *)hv->mem + addr, len);
-                send_packet("");
+                hex2mem(obuf, hv->mem + addr, len);
+                send_okay_msg();
```
ricarkol: This was actually a fix. We were sending not supported instead of OK.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -63,60 +65,83 @@ SLIST_HEAD(breakpoints_head, breakpoint_t);
...
     };
     int n = 0;
 
     if (stepping)
         dbg.control = KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_SINGLESTEP;
 
-    if (!SLIST_EMPTY(&sw_breakpoints)) {
+    if (!SLIST_EMPTY(&sw_breakpoints))
         dbg.control |= KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_SW_BP;
-    } else {
```
ricarkol: This was a bug, we can have sw and hw at the same time. Fixed in the last commit.

tests/test_hello/GNUmakefile
```diff
@@ -19,6 +19,6 @@
 UKVM_TARGETS=test_hello.ukvm ukvm-bin
 VIRTIO_TARGETS=test_hello.virtio
 MUEN_TARGETS=test_hello.muen
-UKVM_MODULES=
+UKVM_MODULES=gdb # Just check that it builds.
```
mato: `ukvm/GNUmakefile` already does the "do all modules build" check, so this is unnecessary.

ukvm/ukvm.h
```diff
@@ -142,4 +142,55 @@ extern struct ukvm_module ukvm_module_blk;
...
+
+/*
+ * Fills *registers with a stream of hexadecimal digits for each register in
```
mato: `*registers` -> `*reg`. Also, this should say "for each guest register" or something like that.

ukvm/ukvm.h
```diff
@@ -142,4 +142,55 @@ extern struct ukvm_module ukvm_module_blk;
...
+ * GDB register order, where each register is in target endian order.
+ * Returns 0 if success, -1 otherwise.
+ */
+int ukvm_gdb_read_registers(struct ukvm_hv *hv, uint8_t *reg, size_t *len);
+
+/*
+ * Writes the shadow registers from a stream of hexadecimal digits for each register in
```
mato: "shadow registers" -> "guest registers"? Also, "for each register in *reg"?

ukvm/ukvm.h
```diff
@@ -142,4 +142,55 @@ extern struct ukvm_module ukvm_module_blk;
...
+ * GDB register order, where each register is in target endian order.
+ * Returns 0 if success, -1 otherwise.
+ */
+int ukvm_gdb_write_registers(struct ukvm_hv *hv, uint8_t *reg, size_t len);
+
+/*
+ * Enagle single stepping. Returns 0 if success, -1 otherwise.
```
mato: Typo

ukvm/ukvm.h
```diff
@@ -142,4 +142,55 @@ extern struct ukvm_module ukvm_module_blk;
...
+int ukvm_gdb_disable_ss(struct ukvm_hv *hv);
+
+/*
+ * Reads the current KVM exit code and maps it to a GDB signal value.
+ * Returns 0 if success, -1 otherwise.
+ */
+int ukvm_gdb_read_last_signal(struct ukvm_hv *hv, int *signal);
+
+/*
+ * Add a breakpoint of type software or hardware, at address addr.  len is
```
mato: Unclear what are the valid values for `type` here, mention that these are defined in `ukvm_gdb.h`?

ukvm/ukvm.h
```diff
@@ -142,4 +142,55 @@ extern struct ukvm_module ukvm_module_blk;
...
+ * typically the size of the breakpoint in bytes that should be inserted
+ * Returns 0 if success, -1 otherwise.
+ */
+int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, uint32_t type,
+                            ukvm_gpa_t addr, size_t len);
+
+/*
+ * Remove a breakpoint of type software or hardware, at address addr.  len is
```
mato: As above.

ukvm/ukvm_gdb.h
```diff
@@ -0,0 +1,40 @@
...
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef UKVM_GDB_H
+#define UKVM_GDB_H
+
+/* GDB breakpoint/watchpoint types */
+#define GDB_BREAKPOINT_SW        0
+#define GDB_BREAKPOINT_HW        1
+#define GDB_WATCHPOINT_WRITE     2
```
mato: Are the `GDB_WATCHPOINT_XXX` values supported? If not then we should probably remove them.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+#include <linux/kvm_para.h>
+
+#include "ukvm.h"
+#include "ukvm_hv_kvm.h"
+#include "ukvm_gdb_x86_64.h"
+#include "ukvm_gdb.h"
+#include "queue.h"
+
+struct breakpoint_t {
+    uint32_t type;
```
mato: ... valid values?

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    return 0;
+}
+
+static struct breakpoint_t *bp_list_find(uint32_t type, ukvm_gpa_t addr, size_t len)
+{
+    struct breakpoint_t *bp;
+
+    if (type == GDB_BREAKPOINT_SW) {
```
mato: This if/else chain should reject invalid values for `type`.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+    bp = malloc(sizeof (struct breakpoint_t));
+    if (bp == NULL)
+        return NULL;
+
+    bp->addr = addr;
+    bp->type = type;
+    bp->len = len;
+    bp->refcount = 1;
+
+    if (type == GDB_BREAKPOINT_SW) {
```
mato: This function should reject invalid values for `type`.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+                return bp;
+        }
+    }
+
+    return NULL;
+}
+
+/*
+ * Adds a new breakpoint to the list of breakpoints 
+ * Returns the found or created breakpoint, NULL if otherwise.
```
mato: The comment seems incorrect, how does the function return a "found but not created" breakpoint?

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+
+    bp = bp_list_find(type, addr, len);
+    if (!bp)
+        return -1;
+
+    bp->refcount--;
+    if (bp->refcount > 0)
+        return 0;
+
+    if (type == GDB_BREAKPOINT_SW) {
```
mato: Reject invalid values for `type`.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+        SLIST_REMOVE(&hw_breakpoints, bp, breakpoint_t, entries);
+        nr_hw_breakpoints--;
+    }
+
+    free(bp);
+
+    return 0;
+}
+
+/*
```
mato: No need to have duplicate documentation here (and elsewhere below) for functions already documented in `ukvm.h`.

ukvm/ukvm_gdb_kvm_x86_64.c
```diff
@@ -0,0 +1,460 @@
...
+        return 0;
+
+    bp = bp_list_insert(type, addr, len);
+    if (bp == NULL)
+        return -1;
+
+    if (type == GDB_BREAKPOINT_SW)
+        kvm_arch_insert_sw_breakpoint(hv, bp);
+
+    if (ukvm_gdb_update_guest_debug(hv) < 0)
```
mato: I'd prefer to use `== -1` rather than `< 0` in checks like these and more below.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
-int remote_debug;
-/*  debug >  0 prints ill-formed commands in valid packets & checksum errors */
-
-static const char hexchars[] = "0123456789abcdef";
-
-/* Number of registers.  */
-#define NUMREGS        32
+    listen_socket_fd = socket(AF_INET, SOCK_STREAM, 0);
+    if (listen_socket_fd == -1) {
+        errx(1, "Could not create socket");
```
mato: -> `err()` since `errno` is valid here.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
+        return -1;
+    }
+    
+    opt = 1;
+    if (setsockopt(listen_socket_fd, SOL_SOCKET, SO_REUSEADDR, &opt,
+                   sizeof(opt)) < 0)
+        errx(1, "setsockopt(SO_REUSEADDR) failed");
```
mato: -> `err()` since `errno` is valid here.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
+
+    server_addr.sin_family = AF_INET;
+    server_addr.sin_addr.s_addr = htonl(INADDR_ANY);
+    server_addr.sin_port = htons(portno);
+     
+    if (bind(listen_socket_fd, (struct sockaddr *)&server_addr,
+             sizeof(server_addr)) < 0) {
+        errx(1, "bind failed");
```
mato: -> `err()` since `errno` is valid here.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
+        return -1;
+    }
+     
+    listen(listen_socket_fd , 0);
```
mato: Can fail, check status.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
+     
+    warnx("Waiting for a debugger. Connect to it like this:");
+    warnx("\tgdb --ex=\"target remote localhost:%d\" KERNEL", portno);
+     
+    len = sizeof(client_addr);
+    socket_fd = accept(listen_socket_fd, (struct sockaddr *)&client_addr, &len);
+    if (socket_fd < 0) {
+        errx(1, "accept failed");
```
mato: -> `err()`

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
-    RAX, RBX, RCX, RDX,
-    RSI, RDI, RBP, RSP,
-    R8, R9, R10, R11,
-    R12, R13, R14, R15,
-    RIP, EFLAGS, CS, SS,
-    DS, ES, FS, GS
-};
+    protoent = getprotobyname("tcp");
+    if (!protoent) {
+        errx(1, "getprotobyname (\"tcp\") failed");
```
mato: -> `err()`

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
+    }
 
-/*
- * these should not be static cuz they can be used outside this module
- */
-long registers[NUMREGS];
+    opt = 1;
+    if (setsockopt(socket_fd, protoent->p_proto, TCP_NODELAY, &opt,
+                   sizeof(opt)) < 0)
+        errx(1, "setsockopt(TCP_NODELAY) failed");
```
mato: -> `err()`

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
 {
-    if ((ch >= 'a') && (ch <= 'f'))
-        return (ch - 'a' + 10);
-    if ((ch >= '0') && (ch <= '9'))
-        return (ch - '0');
-    if ((ch >= 'A') && (ch <= 'F'))
-        return (ch - 'A' + 10);
-    return -1;
+    /* TCP is already buffering, so no need to buffer here as well. */
+    send(socket_fd, &ch, 1, 0);
```
mato: Should check status from `send()` here. `assert()` is probably fine for now, unless you want to refactor the code more...

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
 }
 
+static unsigned char recv_char(void)
+{
+    unsigned char ch;
+    recv(socket_fd, &ch, 1, 0);
```
mato: Check status from `recv()` here. `assert()` is probably fine for now, unless you want to refactor the code more...

ukvm/ukvm_module_gdb.c
```diff
@@ -361,278 +256,252 @@ unsigned char *getpacket(void)
...
+    send_char(hexchars[checksum % 16]);
 }
 
-
-int gdb_is_pc_breakpointing(uint64_t addr)
+/*
+ * Send a packet and wait for a successful ACK of '+' from the debugger.
+ * An ACK of '-' means that we have to resend.
+ */
+static void send_packet(unsigned char *buffer)
```
mato: Any particular reason why `buffer` is an `unsigned char`? This seems to result in a bunch of unnecessary casting all over the place later on, also applies to `packet` and `obuf` in `gdb_handle_exception()` and `send_response()`.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
+    if ((ch >= 'A') && (ch <= 'F'))
+        return (ch - 'A' + 10);
+    return -1;
 }
 
-
-static char buf[4096], *bufptr = buf;
-static void flush_debug_buffer(void)
+/*
+ * Converts the memory pointed to by mem into an hex string in buf.
```
mato: `(count)` bytes of memory?

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,206 @@
...
-        flush_debug_buffer();
-    *bufptr++ = ch;
+    *buf = 0;
+    return buf;
 }
 
-
-int getDebugChar(void)
+/*
+ * Converts the hex string in buf into binary in mem.
```
mato: If `buf` is a C string why does this need `count`?

ukvm/ukvm_module_gdb.c
```diff
@@ -332,25 +230,22 @@ unsigned char *getpacket(void)
...
             xmitcsum += hex(ch);
 
             if (checksum != xmitcsum) {
-                if (remote_debug) {
-                    fprintf(stderr,
-                            "bad checksum.  My count = 0x%x, sent=0x%x. buf=%s\n",
-                            checksum, xmitcsum, buffer);
-                }
-                putDebugChar('-');        /* failed checksum */
+                errx(1, "bad checksum.  My count = 0x%x, sent=0x%x. buf=%s",
```
mato: Maybe a more descriptive error message? "Failed checksum from GDB"? Also, you then send a response so should this be a `warnx` instead?

ukvm/ukvm_module_gdb.c
```diff
@@ -361,278 +256,252 @@ unsigned char *getpacket(void)
...
+        case 'G': {
+            /* Write general registers */
+            len = BUFMAX;
+            /* Call read_registers just to get len (not very efficient). */
+            if (ukvm_gdb_read_registers(hv, registers, &len) < 0) {
+                send_error_msg();
+                break;
+            }
+            /* Packet looks like 'Gxxxxx', so we have to skip the first char */
+            hex2mem(packet + 1, registers, len);
```
mato: Should `len` be `len -1`?

ricarkol: No, `len` does not include the "header" (the `G`).

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Removed the duplicated documentation.

ukvm/ukvm.h
ricarkol: Added a `gdb_breakpoint_type` enum.

tests/test_hello/GNUmakefile
ricarkol: Removed.

ukvm/ukvm.h
ricarkol: Changed.

ukvm/ukvm.h
ricarkol: Changed

ukvm/ukvm_gdb.h
ricarkol: They are supported.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: added an enum

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Changed all instances.

ukvm/ukvm_module_gdb.c
ricarkol: Changed.

ukvm/ukvm_module_gdb.c
ricarkol: In all of these cases `strlen(buf) == count`, however this function could be used for other commands where the data sent from GDB is made of multiple smaller commands.

ukvm/ukvm_module_gdb.c
ricarkol: changed

ukvm/ukvm_module_gdb.c
ricarkol: added error handling

ukvm/ukvm_module_gdb.c
ricarkol: Added an assert for now.

ukvm/ukvm_module_gdb.c
ricarkol: Added an assert for now.

ukvm/ukvm_module_gdb.c
ricarkol: Changed.

ukvm/ukvm_module_gdb.c
ricarkol: This was due to the previous commit that tried to change all `char *`'s to `unsigned char *``s. I thought you were suggesting using `unsigned char *`'s for all C strings. This new version consistently uses `char *` for C strings, and `unsigned char *`'s for memory. Look at `hex2mem` as an example.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Everything that's not a software breakpoint is treated as a hardware breakpoint. The reason is that we implement watchpoints as hardware breakpoints. Changed the if/else to a switch case for more clarity. Also added some comments about that.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Changed.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Asserting on invalid types. We should never get invalid types from GDB. And we handle all possible types.

ukvm/ukvm_gdb_kvm_x86_64.c
ricarkol: Asserting on invalid types. We should never get invalid types from GDB. And we handle all possible types.

ricarkol:
Thanks for the review @mato . Committed a change with all the requested changes.

ukvm/ukvm_module_gdb.c
```diff
@@ -19,308 +19,223 @@
...
-long registers[NUMREGS];
+    opt = 1;
+    if (setsockopt(socket_fd, protoent->p_proto, TCP_NODELAY, &opt,
+                   sizeof(opt)) == -1)
+        err(1, "setsockopt(TCP_NODELAY) failed");
 
-/***************************  ASSEMBLY CODE MACROS *************************/
-/*                                                                            */
+    ip_addr.s_addr = client_addr.sin_addr.s_addr;
+    warnx("Connection from debugger at %s\n", inet_ntoa(ip_addr));
```
mato: Minor nit (just noticed now while testing), there's an extra `\n` in this call to `warnx()`.

ricarkol: Left that one on purpose because the output felt a bit "crowded". Want me to remove it?

mato:
Did some testing, found the following problems:

1. Shutdown (detach or quit) is racy. Half the time when doing `quit` on the GDB side I get "ukvm_module_gdb.c:197: send_char: Assertion \`ret == 1' failed.`". So it looks like the error handling and/or logic dealing with shutdown needs a bit more work.

2. There's something odd going on with memory accesses:

````
$ gdb -q --ex="target remote localhost:1234" ./test_hello.ukvm
Reading symbols from ./test_hello.ukvm...done.
Remote debugging using localhost:1234
_start (arg=0x5000) at ukvm/kernel.c:24
24	{
(gdb) b process_bootinfo 
Breakpoint 1 at 0x1000b0: file ukvm/platform.c, line 30.
(gdb) c
Continuing.

Breakpoint 1, process_bootinfo (arg=arg@entry=0x5000) at ukvm/platform.c:30
30	    cmdline = bi->cmdline;
(gdb) bt
#0  process_bootinfo (arg=arg@entry=0x5000) at ukvm/platform.c:30
#1  0x0000000000100195 in platform_init (arg=arg@entry=0x5000)
    at ukvm/platform_init.c:25
#2  0x000000000010001b in _start (arg=0x5000) at ukvm/kernel.c:30
(gdb) p *bi
Cannot access memory at address 0x5014
(gdb) x /16x 0x5000
0x5000:	0x20000000	0x00000000	0x0010e000	0x00000000
0x5010:	0x00006000	0x00000000	0xb8590de0	0x00000000
0x5020:	0x00000000	0x00000000	0x00000000	0x00000000
0x5030:	0x00000000	0x00000000	0x00000000	0x00000000
(gdb) x /1c 0x5014
0x5014:	0 '\000'
(gdb) p *bi
Cannot access memory at address 0x5014
(gdb) p bi->cmdline
$1 = 0x6000 "foo bar"
````

Note the "Cannot access memory at address 0x5014".

File: ukvm/ukvm_module_gdb.c (hunk 48 above)
Comment (mato) [inline ukvm/ukvm_module_gdb.c]: Yes please, if you don't mind.

**Comment by ricarkol** [comment]
Thanks for the testing. Will investigate this. There shouldn't be anything special about 0x5014. There has to be some command option coming from the gdb client that's not being managed correctly (or rejected).

**Comment by mato** [comment]
@ricarkol

> This was due to the previous commit that tried to change all `char *`'s to `unsigned char *``s. I thought you were suggesting using `unsigned char *`'s for all C strings. This new version consistently uses `char *` for C strings, and `unsigned char *`'s for memory. Look at `hex2mem` as an example.

This new version is good. And what I meant was what you write above :-)
Thanks!

**Comment by mato** [comment]
On a backend where GDB is not yet implemented behaviour is confusing; e.g. on FreeBSD I can start ukvm with `--gdb` and only upon connecting to it do I get an obscure error...

````
target remote localhost:1234
Remote debugging using localhost:1234
Couldn't establish connection to remote target
Remote failure reply: E01
````

...and an assertion failure in `recv_char()` on the ukvm side, presumably due to GDB closing the socket.

Given that it will take a while for all backends to gain debugger support. I suggest adding an additional function to the top-level `ukvm.h` API:

````
/*
 * Check if backend implements GDB support. Returns 0 if supported, -1 if not.
 * Note backends not implementing GDB support still need to implement the remaining functions in this
 * section, all of which should return -1.
 */
int ukvm_gdb_supported(void);
````

The GDB module should check for this in `setup()` and respond with a meaningful error ("GDB support not implemented on this backend/architecture").

@ricarkol What do you think?

**Comment by ricarkol** [comment]
Thanks again mato. Fixed both issues: correctly parsing the `m` command (read memory), and better error handling on `send_char` and `get_char`.

File: ukvm/ukvm_module_gdb.c [hunk 49]
```diff
@@ -249,17 +260,23 @@ static char *recv_packet(void)
...
             if (checksum != xmitcsum) {
                 warnx("Failed checksum from GDB. "
                       "My count = 0x%x, sent=0x%x. buf=%s",
                       checksum, xmitcsum, buffer);
-                send_char('-');        /* failed checksum */
+                if (send_char('-') == -1)       /* failed checksum */
+                    err(1, "GDB: Could not send an ACK to the debugger.");
             } else {
-                send_char('+');        /* successful transfer */
+                if (send_char('+') == -1)       /* successful transfer */
```
mato: That should probably read "unsuccessful transfer"?

mato:
Thanks @ricarkol, the latest version looks good to me. Have tested some memory accesses and also behaviour on an unsupported target (FreeBSD) and all works as expected. I'll leave it to you to merge when ready.

The next step would then be to discuss with the Mirage folks (probably in a PR) what the best way to integrate "enabling GDB in the ukvm built for a unikernel" is. Also, please update the documentation once that is sorted since the current docs are out of date.

ricarkol:
Thanks mato.

Yes, the mirage change is the next item. Will update solo5 README after that.

hannesm:
yes, this is merged... but let me raise some concern:  until now (block/net), command line arguments are prefixed with the module name (`--block` / `--net` / `--net-mac`) -- which I find very intuitive.

This module now exposes `--gdb` and `--port` -- could we rename `--port` to `--gdb-port`?

mato:
@hannesm Makes sense to me. @ricarkol Any objections to renaming `--port` to `--gdb-port`?

hannesm:
@mato I edited my comment and added some discussion about `strtol` vs `sscanf`.

mato:
@hannesm The `10` in `strtol()` is the base, not a length. Having said that, `--port` parameter parsing should be corrected to either handle all the various errors that `strtol()` can return (there are quite a few!) or, possibly easier, just check that the resulting value of `port` is not garbage (ie. within `0` ... `65535`) and abort with `err()` if it is.

hannesm:
thx, I opened #208 awaiting @ricarkol @mato review for this

PR #199, opened 2017-06-06
ukvm_elf: Do not mprotect an elf segment when it's writable.

ricarkol:
Segments can be `PF_R + PF_W + PF_X`. We should not disable writes for those.

Just in case, we only care about `PF_R + PF_X` (executable) segments. And `PR_R`, `PF_W`, and `PF_X` are the only available flags in the header.

### DIFF
--- a/ukvm/ukvm_elf.c
+++ b/ukvm/ukvm_elf.c
@@ -160,6 +160,7 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
         uint64_t paddr = phdr[ph_i].p_paddr;
         uint64_t align = phdr[ph_i].p_align;
         uint64_t result;
+        int prot;
 
         if (phdr[ph_i].p_type != PT_LOAD)
             continue;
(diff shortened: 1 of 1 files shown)


mato:
Can we do something like this instead?

````
--- a/ukvm/ukvm_elf.c
+++ b/ukvm/ukvm_elf.c
@@ -192,11 +192,18 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
             goto out_invalid;
         memset(daddr + filesz, 0, memsz - filesz);
 
-        /* Write-protect the executable segment */
-        if (phdr[ph_i].p_flags & PF_X) {
-            if (mprotect(daddr, _end - paddr, PROT_EXEC | PROT_READ) == -1)
-                goto out_error;
-        }
+        int prot = PROT_NONE;
+        if (phdr[ph_i].p_flags & PF_R)
+            prot |= PROT_READ;
+        if (phdr[ph_i].p_flags & PF_W)
+            prot |= PROT_WRITE;
+        if (phdr[ph_i].p_flags & PF_X)
+            prot |= PROT_EXEC;
+        if (prot & PROT_WRITE && prot & PROT_EXEC)
+            warnx("%s: Warning: phdr[%u] requests WRITE and EXEC permissions",
+                    file, ph_i);
+        if (mprotect(daddr, _end - paddr, prot) == -1)
+            goto out_error;
     }
 
     free (phdr);
````

**Comment by ricarkol** [comment]
Thanks for the review mato. Updated the change accordingly.

I saw these xwr segments in includeos elfs. I wonder if C++ needs them (for some reason).

File: ukvm/ukvm_elf.c [hunk 1]
```diff
@@ -193,10 +194,18 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
         memset(daddr + filesz, 0, memsz - filesz);
 
         /* Write-protect the executable segment */
```
mato: Can you drop the comment entirely since it's now out of date? Thanks

mato:
> I saw these xwr segments in includeos elfs. I wonder if C++ needs them (for some reason). 

More likely to be an artefact of the IncludeOS build system or link process. Anyhow, this is why I added the `warnx()` rather than refusing such segments out right.

ricarkol:
Removed the comment. Thanks!

mato:
Thanks @ricarkol. Have run the tests on FreeBSD just to check that the `mprotect()` change doesn't break anything and all is fine, so merging.

Issue #200, opened 2017-06-13
[tracking] Minimal refresh of Solo5 public APIs


mato:
The Solo5 public APIs (see [solo5.h](kernel/solo5.h)) have evolved over the course of the project and suffer from various issues, including but not limited to:

- Type confusion (eg. use of `int` rather than `size_t`)
- Lack of `const` types for read-only parameters
- Under-specified failure cases
- Difference of opinion between target implementations (eg. virtio vs ukvm multi-sector block operations)

With the various different ports and targets progressing we need to do a minimal refresh of the APIs sooner rather than later to both ease the work on new ports/targets and avoid technical debt. This will require breaking changes to the APIs, however the intent is to keep the changes minimal as opposed to doing a full redesign of the APIs at this point.

This issue tracks the work in progress and PRs (including downstream changes in MirageOS). The changes will require new OPAM releases of Solo5 and MirageOS mid-layers the progress of which will also be tracked here.

mato:
Solo5 changes are in #201. Work in progress branches for MirageOS mid-layers are:

- https://github.com/mato/ocaml-freestanding/tree/solo5-api-refresh
- https://github.com/mato/mirage-solo5/tree/solo5-api-refresh
- https://github.com/mato/mirage-block-solo5/tree/solo5-api-refresh
- https://github.com/mato/mirage-bootvar-solo5/tree/solo5-api-refresh
- https://github.com/mato/mirage-console-solo5/tree/solo5-api-refresh
- https://github.com/mato/mirage-net-solo5/tree/solo5-api-refresh

These branches may be rebased, I will create PRs for them later.

/cc @hannesm This should address your `Console` concerns.

mato:
This is completed on the Solo5 side by #245, closing.

PR #201, opened 2017-06-13
Minimal refresh of Solo5 public APIs

mato:
This is a minimal refresh of the Solo5 public APIs and their underlying backend implementations for ukvm, virtio and muen to fix the most egregious problems as described in #200.

Summary of high-level changes, see [solo5.h](kernel/solo5.h) for full details:
1. `solo5_app_main()`: Add `const` to `cmdline`, document as returning a status code.
2. `solo5_net_write_sync()`: Add `const` to `buf`, pass `size` as `size_t`, document and enforce failure cases.
3. `solo5_net_read_sync()`: Pass sizes as `size_t`, return `read_size` separately from error code, document and enforce failure cases.
4. `solo5_net_mac_str()`: Add `const` to return type.
5. `solo5_blk_write_sync()`: Add `const` to `buf`, use `size_t` for sizes, restrict to single-sector full writes, document and enforce failure cases.
6. `solo5_blk_read_sync()`: Use `size_t` for sizes, restrict to single-sector full reads, document and enforce failure cases.
7. `solo5_blk_sectors()`: Rename to `solo5_blk_capacity()` to lessen confusion with `solo5_blk_sector_size()`.
8. `solo5_console_write()`: Change return type to `void` and document as _not_ having explicit failure cases.
9. All other public functions: Clarify documentation.

No functional change intended, hopefully a lot of edge case bugs elided. Tests have been updated and are passing on Linux and FreeBSD for `ukvm` and `virtio`.

### DIFF
--- a/kernel/solo5.h
+++ b/kernel/solo5.h
@@ -31,69 +31,96 @@
 /*
  * Application entry point.
  *
- * FIXME: These APIs are to be considered unstable and have multiple known
- * issues (type confusion, missing const-ness, in/out parameters, inconsistent
- * return values). The comments in this file document the current
- * implementation.
+ * The value returned from this function is the application's exit status.
+ *
+ * FIXME: These APIs are to be considered unstable.  The comments in this file
+ * document the current implementation.
  */
-int solo5_app_main(char *cmdline);
+int solo5_app_main(const char *cmdline);
 
 /*
  * Network I/O.
+ *
+ * Currently only a single network device is supported. This device mimics an
+ * Ethernet v2 network interface with an MTU of 1500 bytes. Therefore, buffers
+ * passed to solo5_net_read_sync() must be at least 1514 bytes in length
+ * (including space for the Ethernet data link layer).
  */
 
 /*
- * Send a single network packet (*data) of size (n). Returns -1 on failure, 0
- * on success.
+ * Transmit a single network packet of (size) bytes from (*buf). (size) must be
+ * less than or equal to 1514 bytes.
+ *
+ * On failure:
+ *     Returns -1. This includes transient error conditions (e.g. short write)
+ *     and invalid arguments (e.g. size too big).
+ * On success:
+ *     Returns 0.
  */
-int solo5_net_write_sync(uint8_t *data, int n);
+int solo5_net_write_sync(const uint8_t *buf, size_t size);
 
 /*
- * Receive a single network packet into (*data) up to a maximum size of (*n).
- * Returns 0 on success and size of received packet (*n), -1 on failure or if
- * no packets are available.
+ * Receive a single network packet into (*buf), up to a maximum of (buf_size)
+ * bytes. (buf_size) must be at least 1514 bytes.
+ *
+ * On failure:
+ *     Returns -1. This includes invalid argument errors (e.g. buf_size too
+ *     small).
+ * If no packets are available:
+ *     Returns 0 and a (*read_size) of 0.
+ * On success:
+ *     Returns 0 and the length of the received packet in bytes in (*read_size).
  */
-int solo5_net_read_sync(uint8_t *data, int *n);
+int solo5_net_read_sync(uint8_t *restrict buf, size_t buf_size, size_t
+        *restrict read_size);
 
 /*
  * Returns a pointer to the network MAC address, formatted as a C string
  * XX:XX:XX:XX:XX:XX.
  */
-char *solo5_net_mac_str(void);
+const char *solo5_net_mac_str(void);
 
 /*
  * Block I/O.
- */
-
-/*
- * Writes (n) bytes from (*data) starting at the sector (sec) to the block
- * device. (n) does not need to be a multiple of the sector size. Returns 0 on
- * success, -1 on error.
- */
-int solo5_blk_write_sync(uint64_t sec, uint8_t *data, int n);
-
-/*
- * Reads (*n) bytes starting at the sector (sec) from the block device into
- * (*data). (n) does not need to be a multiple of the sector size. Returns 0 on
- * success, -1 on error.
  *
- * XXX: (*n) is not set as an output parameter. ukvm assert()s that exactly
- * (*n) bytes were written to the device.
+ * Currently only a single sector-addressable block device is supported, and
+ * only single-sector reads and writes are supported.
  */
-int solo5_blk_read_sync(uint64_t sec, uint8_t *data, int *n);
 
 /*
- * Returns the block device sector size.
+ * Writes (size) bytes from (*buf) at the sector (offset) to the block device.
+ * (size) must be equal to the sector size.
+ *
+ * On failure:
+ *     Returns -1. This includes invalid argument errors and incomplete writes.
+ * On success:
+ *     Returns 0.
+ */
+int solo5_blk_write_sync(uint64_t offset, const uint8_t *buf, size_t size);
+
+/*
+ * Reads (size) bytes at the sector (offset) from the block device into (*buf).
+ * (size) must be equal to the sector size.
+ *
+ * On failure:
+ *     Returns -1. This includes invalid argument errors and incomplete reads.
+ * On success:
+ *     Returns 0.
+ */
+int solo5_blk_read_sync(uint64_t offset, uint8_t *buf, size_t size);
+
+/*
+ * Returns the block device sector size in bytes.
  */
 int solo5_blk_sector_size(void);
 
 /*
- * Returns the size of the block device in sectors.
+ * Returns the capacity of the block device in sectors.
  */
-uint64_t solo5_blk_sectors(void);
+uint64_t solo5_blk_capacity(void);
 
 /*
- * Returns true if the block device is writable.
+ * Returns 1 if the block device is writable, otherwise 0.
  */
 int solo5_blk_rw(void);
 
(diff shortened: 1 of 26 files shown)


mato:
@Kensan I've not tested this on Muen, you might want to double-check my changes.

Kensan:
@mato Will give it a spin as soon as I can.

Kensan:
@mato I tested the changes on Muen and the muen-net implementation requires some adaptation, see b424a29. With that **test_ping_serve** works again.

kernel/solo5.h
```diff
@@ -31,69 +31,96 @@
...
+ * Writes (size) bytes from (*buf) at the sector (offset) to the block device.
+ * (size) must be equal to the sector size.
+ *
+ * On failure:
+ *     Returns -1. This includes invalid argument errors and incomplete writes.
+ * On success:
+ *     Returns 0.
  */
-int solo5_blk_write_sync(uint64_t sec, uint8_t *data, int n);
+int solo5_blk_write_sync(uint64_t offset, const uint8_t *buf, size_t size);
```
ricarkol: I can't remember the number of times I had to check if offset is in bytes or sectors. Could offset be something like `sector_t offset` or `sector_t sector`.

Also, size should be number of sectors. we might never support writing less than a sector. As that would require reading it first.

kernel/solo5.h
```diff
@@ -136,8 +165,13 @@ uint64_t solo5_clock_monotonic(void);
...
+ * Blocks until monotonic time reaches until_nsecs or network I/O is possible,
+ * whichever is sooner.
+ *
+ * If network I/O is possible:
+ *     Returns 1.
+ * If monotonic time has reached until_nsecs and network I/O is not possible:
+ *     Returns 0.
  *
  * TODO: Extend this interface to select which I/O events are of interest.
  */
```
ricarkol: To be more like the POSIX poll(), the time should be a delta and not a future timestamp. Plus, that would remove this time reading in the mirage event loop:

```
        let timeout =
          match Time.select_next () with
          |None -> Time.Monotonic.(time () + of_nanoseconds 86_400_000_000_000L) (* one day = 24 * 60 * 60 s *)
          |Some tm -> tm
        in
        if poll timeout then begin
```

https://github.com/mirage/mirage-solo5/blob/master/lib/main.ml#L68

On the other hand, that's a risky change, so up to you.

ukvm/ukvm_module_blk.c
```diff
@@ -79,7 +96,17 @@ static void hypercall_blkread(struct ukvm_hv *hv, ukvm_gpa_t gpa)
     ssize_t ret;
     off_t pos, end;
 
+#if 0
     assert(rd->len <= SSIZE_MAX);
+#else
+    /*
+     * XXX: Artificially limit to single sector writes for now.
```
hannesm: `writes` -> `reads`

tests/test_blk/test_blk.c
```diff
@@ -59,18 +55,17 @@ int check_sector_write(uint64_t sector)
...
 
     puts("\n**** Solo5 standalone test_blk ****\n\n");
 
     /*
      * Write and read/check one tenth of the disk.
      */
-    nsectors = solo5_blk_sectors();
-    for (i = 0; i <= nsectors; i += 10) {
+    nsectors = solo5_blk_capacity();
+    for (i = 0; i < nsectors; i += 10) {
```
hannesm: here, `<=` is changed to `<`, is this intentional?

kernel/solo5.h
```diff
@@ -102,15 +129,17 @@ int solo5_blk_rw(void);
...
 
 /*
- * Writes (n) bytes from (*buf) to the console. Returns number of bytes
- * actually written.
+ * Writes (size) characters from (*buf) to the console.
+ *
+ * Always succeeds, however the implementation may truncate messages (e.g. due
```
hannesm: this is not entirely true: if `write` returns `< 0`, an assertion is raised (in the ukvm implementation): https://github.com/mato/solo5/blob/13e7a06cfcb24452d0e8113057cf4719f7ca2915/ukvm/ukvm_core.c#L103

kernel/solo5.h
```diff
@@ -136,8 +165,13 @@ uint64_t solo5_clock_monotonic(void);
 uint64_t solo5_clock_wall(void);
```
hannesm: would be great to specify error behaviour for these as well... at least the ukvm implementation asserts that `clock_gettime` returns non-zero

kernel/solo5.h
mato: I believe that the use of "sectors" in this API comes from @djwillia's original MirageOS porting work; MirageOS uses sectors in it's block device [signatures](https://github.com/mirage/mirage-block/blob/master/src/mirage_block.mli#L68).

The current situation using a mix of sectors (for `offset`) and bytes (for `size`) is indeed confusing, however I'm not sure which of the two options is better in the long run:

1. Use byte offsets and sizes, and reject offsets and sizes that are not a multiple of the sector size.
2. Use sector offsets and sizes.

IMO in both cases there's no need for an explicit `sector_t`, `uint64_t` is fine.

In the long run I'd like to abstract block and network I/O over a common API, in which case we'd have to use bytes for offsets and sizes anyway. So, I'd prefer the first option.

Random sprinkling of Mirage folks for other opinions: @djs55 @gjaldon @samoht What do you think?

djs55: The MirageOS OCaml code uses sectors to avoid having to add a runtime error -- we're keen on eliminating runtime errors where we can. If solo5 becomes more like Unix and uses byte offsets then it's fine -- the OCaml code can perform the conversion as it already does for Unix.

kernel/solo5.h
mato: In my experience, the use of relative timeouts in POSIX poll() leads to unnecessary complications in the common case for an event loop, which is when I/O events are regularly arriving _and_ the application has some timers scheduled. Ignoring the Mirage implementation for now, consider the two options in this scenario:

_If relative timeouts are used_: The application _must_ recalculate the relative _timeout_ passed to poll() in each iteration of the loop. Given that poll() does not return[1] the remaining time, this involves a call to get the current time.

_If absolute timeouts are used_: The application can keep calling poll(_t_) in each iteration of the loop, where _t_ is the nearest scheduled timer, until poll() indicates that the time _t_ has been reached[2], at which point the application selects the next scheduled timer.

[1] If poll() is implemented in the backend using select() then some implementations _do_ return the time not slept. However, POSIX permits either behaviour so in practice this cannot be relied upon.

[2] Our API does not currently clearly indicate if the timeout has expired. This is a bug. Here is a strawman proposal for a fixed interface (without any ability to select what I/O is interesting since we don't have any infrastructure for that yet):

````
/*
 * Wait until network packets are available or monotonic time reaches (deadline), whichever is sooner.
 *
 * On success:
 *    Returns 0, and (*io) is true if network packets are available, (*timed_out) is true if (deadline) has passed.
 * On failure:
 *    Returns -1.
 */
int solo5_wait(uint64_t deadline, bool *io, bool *timed_out);
````

In the above proposal I've deliberately renamed the call to `solo5_wait()` to disassociate it from the POSIX poll(); I've had several conversations with people who were confused by the "poll" in the name.

Note that the change in semantics (reporting whether or not timeout has expired) requires a bit more work from the (ukvm) backends. That's fine and can be optimized later using epoll()/timerfd_create() on Linux and kqueue() on FreeBSD.

In summary, I'd prefer that we stick with absolute timeouts for poll() and implement the above proposal or similar as part of this change.

Aside: I went hunting for some modern "wait for events" APIs while thinking about this, those [implemented](https://github.com/fuchsia-mirror/magenta/blob/master/docs/syscalls/waitset_wait.md) by Magenta (the kernel used in Google Fuchsia) are a useful source of future inspiration and incidentally exclusively use absolute timeouts.

tests/test_blk/test_blk.c
mato: Yes, the sector offset is 0-based, so valid values are (0 ... solo5_blk_capacity() - 1). Interestingly there are "bugs" in the current backends (e.g. virtio/bhyve and ukvm) where writing past the end of the "device" would succeed if the underlying device was backed by a file which is the case in all of our tests.

kernel/solo5.h
mato: Thanks for pointing that out. The backend code could be improved to ignore _some_ errors that are obviously a sign of run-time resource exhaustion (`ENOSPC` comes to mind). For anything else it should probably still assert. In either case I don't see any point in returning a failure status to the guest; what could it reasonably do to "recover"?

kernel/solo5.h
mato: `clock_gettime()` can fail due to (on Linux):
- `EFAULT` tp points outside the accessible address space.
- `EINVAL` The clk_id specified is not supported on this system.

We should never get `EFAULT` (since we've already verified the buffer passed by the guest). If we do then we want to terminate the guest with extreme prejudice anyway. We will never get `EINVAL` since `CLOCK_MONOTONIC` is supported on any sane system.

mato: ... in other words `solo5_clock_xxx()` can never fail. If they do for some reason then that is legitimately an unrecoverable (and fatal) error.

kernel/solo5.h
mato: @djs55 Doesn't passing in a buffer that is not a multiple of the sector size count as a runtime error? The documentation in mirage_block.mli is silent about this other than insisting on the buffers being a multiple of the sector size. Likewise, what if an invalid sector is passed (beyond the end of the device)?

djs55: @mato yes that's right-- both of those cases will fail at runtime (possibly killing the app, not sure about that). We ought to define that behaviour a bit better! So perhaps it's more accurate to say that a runtime error exists (unfortunately) but there is one fewer case that can trigger it :) Maybe (big maybe) it would be possible to design the rest of errors out with clever use of types without being too annoying.

kernel/solo5.h
ricarkol: Agree, going for `solo5_wait()` instead of `solo5_poll()` might be the easiest way to avoid confusion.

kernel/solo5.h
hannesm: I'm with you that there's no way to recover from most `write` errors.  I'd rephrase `Always succeeds` to reflect the actual behaviour  (although the virtio backend using `outb` can never fail AFAICS).  Do you expect that compilations always have assertions enabled (and leading to fatal errors)?  Or should there be a call to `exit()` instead of `assert` (sorry if I missed some note which explains how to use assertions in Solo5)?

kernel/solo5.h
hannesm: That is good to hear - related to above, maybe it makes sense to explicitly document in `solo5.h` that fatal errors may occur in each call, which lead to an assertion failure (or `exit`)!?

mato:
Superseded by #245 which should address (or is in the process of addressing) the issues raised here.

PR #204, opened 2017-07-08
fix case fall through for gcc 7

4gh:
gcc 7 warns on switch case fall through.

Due to -Werror this will abort the compilation.

adding /* FALLTHRU */ as a comment stops the warning.

see: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

### DIFF
--- a/kernel/ee_printf.c
+++ b/kernel/ee_printf.c
@@ -637,6 +637,7 @@ repeat:
 
         case 'A':
             flags |= UPPERCASE;
+            /* FALLTHRU */
 
         case 'a':
             if (qualifier == 'l')
@@ -654,6 +655,7 @@ repeat:
 
         case 'X':
             flags |= UPPERCASE;
+            /* FALLTHRU */
 
         case 'x':
             base = 16;


ricarkol:
Thanks for the fix @4gh 

clang 4.0 doesn't seem to like "fallthroughs" either. We use clang for freebsd. I guess clang has some other annotation like `FALLTHRU`, but can't find it.

What about something like this that would work on both cases:

```
        case 'X':
            flags |= UPPERCASE;
            goto fallthrough_x;

        case 'x':
fallthrough_x:
            base = 16;
            break;
```

mato:
@ricarkol That syntax is pretty horrible. Is this actually a problem on Clang? I'm not seeing any failed builds on FreeBSD due to fallthroughs.

I went and hunted for a compiler-agnostic way of specifying the intent to fallthrough but there doesn't appear to be such a thing. Given the few cases involved it might be easiest to just rewrite the offending case statements ie. duplicate the code.

In any case I think this change is fine to merge as is.

ricarkol:
@mato I had a mess of compilers on my side and the warning was actually coming from gcc and not clang. I can't reproduce it with clang-4 either. Sorry for the confusion.

PR #207, opened 2017-07-26
fix strncmp in handle_cmdarg: --net-mac= has 10 chars

hannesm:


mato:
Good catch. We really should have more tests...

hannesm:
I read through the command line processing done in solo5, and found this one issue :)  apart from that I think it is safe

PR #208, opened 2017-07-26
ukvm gdb: rename --port to --gdb-port, handle strtol errors

hannesm:
see discussion in #198


mato:
@hannesm You have a typo in your code, please fix.

ricarkol:
Thanks for the fix @hannesm

PR #209, opened 2017-07-26
ARM64 support

Weichen81:
These commits are used  to enable the ARM64 support on ukvm.
the issue is：
https://github.com/Solo5/solo5/issues/151

The PR is the continuity of  PR https://github.com/Solo5/solo5/pull/193 which was closed by
mistake and could not be reopened.

@mato @djwillia @ricarkol


mato:
@Weichen81 Thanks, this is looking good. I think we're almost ready to merge and we can do further work in later PRs. Thank you for your patience with my reviews :-)

One question. I notice that compared to #193, you have changed the way the MMIO region for hypercall dispatch is set up. I like the arrangement where the MMIO region is (for now) located at 4GB. However: can you confirm that the KVM behaviour of reporting `KVM_EXIT_MMIO` for writes _outside of a defined memory slot_ is intentional? The changes I made in https://github.com/Solo5/solo5/pull/193#issuecomment-316081985 assumed we needed the extra readonly memory slot to back the MMIO region.

Weichen81:
@mato 

Thanks for reviewing this PR so quickly :)

>can you confirm that the KVM behaviour of reporting KVM_EXIT_MMIO for writes outside of a defined memory slot is intentional?

Yes, this is the intentional design of Linux-ARM/KVM. When guest accessed an address outside of the
RAM space. It will be considered as the guest tried to access I/O memory. And then the KVM try to select an I/O emulator to handle this I/O access. If this I/O emulator exists in Kernel (The vgic, ARM's virtual
interrupt controller for guest is implemented as a kernel module), this I/O access will be emulated in
kernel and not report KVM_EXIT_MMIO to userspace. Otherwise, this I/O access will be reported to
userspace by KVM_EXIT_MMIO.

>The changes I made in #193 (comment) assumed we needed the extra readonly memory slot to back the MMIO region.

The memory slot is only used for RAM. The IO memory doesn't need memory slot. The writeable/readonly attribute of MMIO is controlled by I/O emulator.

Regards.

Weichen81:
@mato 
I forgot one thing. Here we mentioned addresses are guest physical address (Intermediate Physical Address - IPA). Because the guest is using virtual address on AArch64. If the guest access a VA above 5GB (We only mapped VA = PA for 4GB RAM + 1GB MMIO, but the whole VA space size could be (1<<48). It's impossible for us to map the whole VA to PA), the guest will get a Data Abort Exception inside VM, will not trap to host. Currently, the guest has handled this kind exceptions and do PANIC.

mato:
@Weichen81:

> Yes, this is the intentional design of Linux-ARM/KVM. When guest accessed an address outside of the
RAM space. It will be considered as the guest tried to access I/O memory. And then the KVM try to select an I/O emulator to handle this I/O access. If this I/O emulator exists in Kernel (The vgic, ARM's virtual
interrupt controller for guest is implemented as a kernel module), this I/O access will be emulated in
kernel and not report KVM_EXIT_MMIO to userspace. Otherwise, this I/O access will be reported to
userspace by KVM_EXIT_MMIO.

Thanks for confirming this.

> I forgot one thing. Here we mentioned addresses are guest physical address (Intermediate Physical Address - IPA). Because the guest is using virtual address on AArch64. If the guest access a VA above 5GB (We only mapped VA = PA for 4GB RAM + 1GB MMIO, but the whole VA space size could be (1<<48). It's impossible for us to map the whole VA to PA), the guest will get a Data Abort Exception inside VM, will not trap to host. Currently, the guest has handled this kind exceptions and do PANIC.

That seems fine, if I understand you correctly it's equivalent to the behaviour on x86_64 where the guest will get a Page Fault exception if it attempts to access memory outside of the range mapped in the guest page tables, in which case this also results in a PANIC.

I noticed you didn't cherry-pick my change fixing `test_time` (https://github.com/mato/solo5/commit/402e296ec7a10aa20caa45c47063b1cd53ee4a9c). Are you OK with me pushing this to your branch so that I can merge it as part of this PR?

Weichen81:
Hi mato,

It's OK. I forgot to pickup that patch, feel free to push it to this branch ;)

Thank you!

mato:
We will also need to enable CI for Solo5 (probably with the surf-build setup) and the rest of MirageOS. I'll discuss the best way to set this up with @avsm next week.

agraf:
The readme still mentions ukvm as x86_64 only - that should probably get fixed :)

Weichen81:
@agraf Thanks for your reminder. I will update the readme :)

PR #210, opened 2017-07-27
Fix --gdb-port handling

mato:
Followup to #208. In this case I actually tested the changes. /cc @hannesm 

Two problems in the original change to --gdb-port:

1. Can't use `strncmp()` to test for `"--gdb"` since this is true for `"--gdb-port=X"` also.
2. `strtol()` error handling is horrible and will happily return `0` with `errno=0` if no digits were matched. Use `sscanf()` instead.


hannesm:
thanks, and sorry for breaking this

PR #211, opened 2017-07-29
add command line argument for memory size

hannesm:
I have tested this locally, and (when using MirageOS), values `>=12 MB` are reasonable (with 8 MB the OCaml runtime won't start up, 10 MB results in a out of memory before execution of main in the hello example from mirage-skeleton/tutorial).  Solo5/UKVM uses 2MB steps (see assertion in ukvm_cpu_x86_64: `mem_size & (X86_GUEST_PAGE_SIZE - 1)) == 0` -- where `X86_GUEST_PAGE_SIZE = 2 << 20`), while this `--mem` option allows 1MB granularity.

(It would make more sense to allow for 2MB steps, but it is not very intuitive to do so -- this is why 1MB steps are allowed in this PR, but an odd value results in an assertion failure.)

Previously, each virtual machine mmap()ed 512MB memory (plus overhead in the host system, which seems to be constant 8MB).  With this change, ~50x reduction can be achieved for the simplest kernels.

### DIFF
--- a/ukvm/ukvm.h
+++ b/ukvm/ukvm.h
@@ -83,7 +83,7 @@ struct ukvm_hv *ukvm_hv_init(size_t mem_size);
 
 /*
  * Computes the memory size to use for this monitor, based on the user-provided
- * value (rounding up if necessary).
+ * value (rounding down if necessary).
  */
 void ukvm_hv_mem_size(size_t *mem_size);
 
--- a/ukvm/ukvm.h
+++ b/ukvm/ukvm.h
@@ -82,6 +82,12 @@ inline void *ukvm_checked_gpa_p(struct ukvm_hv *hv, ukvm_gpa_t gpa, size_t sz,
 struct ukvm_hv *ukvm_hv_init(size_t mem_size);
 
 /*
+ * Computes the memory size to use for this monitor, based on the user-provided
+ * value (rounding up if necessary).
+ */
+void ukvm_hv_mem_size(size_t *mem_size);
+
+/*
  * Initialise VCPU state with (gpa_ep) as the entry point and (gpa_kend) as the
  * last byte of memory used by the guest binary. In (*cmdline), returns a
  * buffer with UKVM_CMDLINE_SIZE bytes of space for the guest command line.
--- a/ukvm/ukvm_cpu_aarch64.c
+++ b/ukvm/ukvm_cpu_aarch64.c
@@ -108,15 +108,14 @@ void aarch64_setup_memory_mapping(uint8_t *va_addr, uint64_t ram_size,
 
 void aarch64_mem_size(size_t *mem_size) {
   size_t mem;
-  if (*mem_size < PMD_SIZE * 4)
-    warnx("adjusting memory to minimum of %zu", PMD_SIZE * 4);
-  mem = PMD_SIZE * 4;
-  mem = (mem + (PMD_SIZE - 1)) / PMD_SIZE;
-  mem *= PMD_SIZE;
+  mem = (*mem_size / PMD_SIZE) * PMD_SIZE;
   if (mem > *mem_size)
-    warnx("adjusting memory to %zu", mem);
+    err(1, "won't increase your memory from %zu bytes to %zu bytes",
+        *mem_size, mem);
+  if (mem < *mem_size)
+    warnx("adjusting memory to %zu bytes", mem);
   if (mem > AARCH64_MMIO_BASE)
-    err(1, "The guest memory %zu exceeds the max size %zu\n",
+    err(1, "guest memory size %zu bytes exceeds the max size %ul bytes",
         mem, AARCH64_MMIO_BASE);
   *mem_size = mem;
 }
--- a/ukvm/ukvm_cpu_aarch64.c
+++ b/ukvm/ukvm_cpu_aarch64.c
@@ -105,3 +105,18 @@ void aarch64_setup_memory_mapping(uint8_t *va_addr, uint64_t ram_size,
         phy_space_size -= map_size;
     }
 }
+
+void aarch64_mem_size(size_t *mem_size) {
+  size_t mem;
+  if (*mem_size < PMD_SIZE * 4)
+    warnx("adjusting memory to minimum of %zu", PMD_SIZE * 4);
+  mem = PMD_SIZE * 4;
+  mem = (mem + (PMD_SIZE - 1)) / PMD_SIZE;
+  mem *= PMD_SIZE;
+  if (mem > *mem_size)
+    warnx("adjusting memory to %zu", mem);
+  if (mem > AARCH64_MMIO_BASE)
+    err(1, "The guest memory %zu exceeds the max size %zu\n",
+        mem, AARCH64_MMIO_BASE);
+  *mem_size = mem;
+}
--- a/ukvm/ukvm_cpu_aarch64.c
+++ b/ukvm/ukvm_cpu_aarch64.c
@@ -113,7 +113,7 @@ void aarch64_mem_size(size_t *mem_size) {
     if (mem < *mem_size)
         warnx("adjusting memory to %zu bytes", mem);
     if (mem > AARCH64_MMIO_BASE)
-        err(1, "guest memory size %zu bytes exceeds the max size %ul bytes",
+        err(1, "guest memory size %zu bytes exceeds the max size %lu bytes",
             mem, AARCH64_MMIO_BASE);
     *mem_size = mem;
 }
--- a/ukvm/ukvm_cpu_aarch64.c
+++ b/ukvm/ukvm_cpu_aarch64.c
@@ -107,13 +107,13 @@ void aarch64_setup_memory_mapping(uint8_t *va_addr, uint64_t ram_size,
 }
 
 void aarch64_mem_size(size_t *mem_size) {
-  size_t mem;
-  mem = (*mem_size / PMD_SIZE) * PMD_SIZE;
-  assert (mem <= *mem_size);
-  if (mem < *mem_size)
-    warnx("adjusting memory to %zu bytes", mem);
-  if (mem > AARCH64_MMIO_BASE)
-    err(1, "guest memory size %zu bytes exceeds the max size %ul bytes",
-        mem, AARCH64_MMIO_BASE);
-  *mem_size = mem;
+    size_t mem;
+    mem = (*mem_size / PMD_SIZE) * PMD_SIZE;
+    assert (mem <= *mem_size);
+    if (mem < *mem_size)
+        warnx("adjusting memory to %zu bytes", mem);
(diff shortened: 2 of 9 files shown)


mato:
I'm not too fond of exposing `--mem=` directly to the user in this state -- as you write yourself, the set of working values is somewhat magic and non-working values result in an assertion which is not helpful.

What is the use case? Enforcing resource allocations on guests? Could you achieve the same thing (for now) with, say, `ulimit -d` on the `ukvm` process?

As to how to make a user-settable `--mem` work sanely, I'll think about what needs to be done internally to at least get sensible error handling working.

hannesm:
Thanks for looking into this, @mato.

My motivation is to lower the footprint on a host with lots of solo5 unikernels (and also to triage potential memory leaks in guest systems easier by limiting them to less memory) :)

Since the patch is very small and local, there's no hurry to upstream it -- I thought others may be interested.

`ulimit -d` (or `limits -d`) is a fine solution, but from the outside instead of from the inside.

mato:
I thought about this some time ago and then forgot to write down my notes for the next steps, so here goes:

I think the best way to handle a user-specified memory size without requiring her to understand the arch/backend constraints on the provided value is to have the arch/backend round the user-provided value up to its nearest acceptable size, logging a warning if the value has changed from that originally specified.

The implementation details are left as an excersize for the reader :-)

hannesm:
is `nearest acceptable size` rounding up or down?  I have arguments for both directions, but intuitively would round up (unless there are strong opinions for rounding down).

mato:
On Thursday, 31.08.2017 at 04:34, Hannes Mehnert wrote:
> is `nearest acceptable size` rounding up or down?  I have arguments for both directions, but intuitively would round up (unless there are strong opinions for rounding down).

Up IMO. Obviously if you hit an arch-defined limit (4G AFAIR for arm and
x86) then it should just fail with an error. 

We can also define some sensible minimum (8MB?)

hannesm:
@mato I gave it a try, if you have some time, please take a look.  as mentioned in the commit message, I'm not entirely sure about the aarch64 path -- it seems to mmap the memory, and check that it is < 4GB (but only after mmapping, similar to x86), but not more than that.

hannesm:
(btw I revised this 4 weeks ago with the suggestions from @mato, would be happy to hear about feedback)

mato:
@hannesm Oh, very sorry, I must have missed this in my INBOX. Marking as "TODO" now.

hannesm:
turns out, I prefer the undercommit much more than an overcommit.  the latest commit implements this: you'll get (n/2MB)*2MB memory (and get a warning if changed).  the case which should never happen `(mem > mem_size)` errors out explicitly (better safe than sorry).  hardcoded minimal value is no longer present.

ukvm/ukvm.h
```diff
@@ -82,6 +82,12 @@ inline void *ukvm_checked_gpa_p(struct ukvm_hv *hv, ukvm_gpa_t gpa, size_t sz,
 struct ukvm_hv *ukvm_hv_init(size_t mem_size);
 
 /*
+ * Computes the memory size to use for this monitor, based on the user-provided
+ * value (rounding up if necessary).
```
mato: Comment needs updating for the change to undercommit.

ukvm/ukvm_cpu_aarch64.c
```diff
@@ -105,3 +105,17 @@ void aarch64_setup_memory_mapping(uint8_t *va_addr, uint64_t ram_size,
         phy_space_size -= map_size;
     }
 }
+
+void aarch64_mem_size(size_t *mem_size) {
+  size_t mem;
+  mem = (*mem_size / PMD_SIZE) * PMD_SIZE;
+  if (mem > *mem_size)
```
mato: Why is this case necessary?

ukvm/ukvm_cpu_x86_64.c
```diff
@@ -23,13 +23,28 @@
...
 #include <stddef.h>
 #include <stdint.h>
 #include <string.h>
 
 #include "ukvm_cpu_x86_64.h"
 
+void ukvm_x86_mem_size(size_t *mem_size) {
+  size_t mem;
+  mem = (*mem_size / X86_GUEST_PAGE_SIZE) * X86_GUEST_PAGE_SIZE;
+  if (mem > *mem_size)
```
mato: Why is this case necessary?

mato:
@hannesm:

>  the case which should never happen (mem > mem_size) errors out explicitly (better safe than sorry).

Oh, I see now. You can disregard my comments about that case then, though if it's truly a "should never happen" then I would use `assert()`.

hannesm:
@mato thx for your review, I added some commits

mato:
@hannesm Thanks, LGTM.

@Weichen81 Unfortunately I can't test on ARM64 at the moment -- would you be able to test that this (`--mem`) option works correctly there and/or quickly review that the code in `aarch64_mem_size()` looks reasonable?

ukvm/ukvm_cpu_aarch64.c
```diff
@@ -105,3 +105,15 @@ void aarch64_setup_memory_mapping(uint8_t *va_addr, uint64_t ram_size,
...
+
+void aarch64_mem_size(size_t *mem_size) {
+    size_t mem;
+    mem = (*mem_size / PMD_SIZE) * PMD_SIZE;
+    assert (mem <= *mem_size);
+    if (mem < *mem_size)
+        warnx("adjusting memory to %zu bytes", mem);
+    if (mem > AARCH64_MMIO_BASE)
+        err(1, "guest memory size %zu bytes exceeds the max size %ul bytes",
+            mem, AARCH64_MMIO_BASE);
```
Weichen81: Here will get a compiler error because AARCH64_MMIO_BASE unsigned long, the the format excepts %ul.
I think this would be %ul?

ukvm/ukvm_cpu_aarch64.c
```diff
@@ -105,3 +105,15 @@ void aarch64_setup_memory_mapping(uint8_t *va_addr, uint64_t ram_size,
         phy_space_size -= map_size;
     }
 }
+
+void aarch64_mem_size(size_t *mem_size) {
+    size_t mem;
+    mem = (*mem_size / PMD_SIZE) * PMD_SIZE;
```
Weichen81: Sorry, here seems use round_down to calculate memory size. From the conversation history I have found a concern about round_up/round_down. IMO, if use round_down, if the application need at less 11MB memory to run, but we'll allocate only 10MB for it. I think the application could not work properly.

Weichen81:
@hannesm @mato 

I have tested this PR. After fixing the compiler error I comment above, the code works properly on my ARM64 server:
 ./ukvm-bin --mem=11 ./test_hello.ukvm
ukvm-bin: adjusting memory to 10485760 bytes
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Memory map: 10 MB addressable:
Solo5:       text @ (0x100000 - 0x106fff)
Solo5:     rodata @ (0x107000 - 0x107fff)
Solo5:       data @ (0x108000 - 0x109fff)
Solo5:       heap >= 0x10a000 < stack < 0xa00000

**** Solo5 standalone test_hello ****

Hello, World
Command line is: ''

ukvm/ukvm_cpu_aarch64.c
mato: You're right about AARCH64_MMIO_BASE defined as `unsigned long`. Did you mean `%lu` for the format? `%ul` is not a legal format, perhaps that's a typo @hannesm?

ukvm/ukvm_cpu_aarch64.c
mato: I don't think it matters too much, since this is the _maximum_ memory size usable by the guest, so I'd not expect people to try and trim it down to _exact_ figures. In any case, it'll only ever get rounded down by 1MB and you will get a warning. We can always revisit this later.

mato:
@Weichen81 Thanks for testing!

ukvm/ukvm_cpu_aarch64.c
Weichen81: Ahh, Yes, it should be %lu :)

ukvm/ukvm_cpu_aarch64.c
Weichen81: Ok, it's not a big problem.

Weichen81:
@mato You're welcome : )

hannesm:
thank you for the additional review!  while checking the other code path (x86_64), I discovered that there was a `%ul` as well, which is now a `%u` (`X86_GUEST_PAGE_SIZE` is an `int`).

## Rounding down vs rounding up

While rounding up (and automatically setting up a minimum) may look good on first sight, if you want to deploy solo5 in an automated fashion (which I do in https://hannes.nqsb.io/Posts/VMM), and you stick with rounding up, you need to reason about the rounded up value (which may be backend-specific, and if you additionally adjust to a minimum value, it is pretty complicated to compute).  Rounding down is much simpler to cope with from the operations perspective (where you need to distribute the actual physical resources).  The failure (not sufficient memory if you assigned `--mem=11`, rounded down to 10, but the application needs 11) includes that the unikernel won't boot up, but the first line of `ukvm-bin` output is a warning `adjusting memory to %zu bytes` - should be pretty easy to debug.

NB: The given `--mem` value does only include the maximum size of the guest system, the `ukvm-bin` monitor in the host system also allocates (constant AFAICS).

ukvm/ukvm_cpu_aarch64.c
hannesm: @Weichen81 I don't know AARCH64 details, this code is very similar to its `x86_64` counterpart, and allocates only in 2MB chunks.  If it is possible on AARCH64 to allocate smaller amounts (such as 512KB or 1MB), we should adjust this code to allow that.  if a unikernel needs 11MB, the only way atm is to allocate 12MB for it, due to the 2MB chunks.

Weichen81: @hannesm 2MB is the minimal granularity for memory allocation on AARCH64. Allocate 512KB or 1MB granularity  may create empty "hole" in mapping table. So you don't need to adjust  your current code for AARCH64.

Weichen81:
@hannesm
I didn't stick to round up :) I just wasn't clear about the reason to select round down. Thanks for your explanation, and now I understand. Your selection seems much reasonable,  using round up will import
more complicated operations, that is not what we want.
Thank you!

mato:
Folks, thanks for your reviews and patience. I think we have consensus, so am merging this.

hannesm:
@mato thanks for your review and input.  I'm now satisfied with the `--mem` argument :)

@Weichen81 thanks for review and your questions, writing down why it is rounding down is useful, and may serve questions asked at a later point.

PR #218, opened 2017-09-19
Added missing stdlib.h in ukvm_hv_kvm.c and ukvm_hv_kvm_x86_64.c

thomaco:
The two edited files in question call malloc(ukvm_hv_kvm.c) and calloc(ukvm_hv_kvm_x86_64.c), but does not include the standard library. As far as I know, this leads to one of two things; either the compilation aborts because of 'implicit declaration of function'(modern compilers) or the compiler assumes a function that returns an int. The first is safe, while the second can be highly dangerous.

This pull request fixes the problem by including stdlib.h in both files.

### DIFF
--- a/ukvm/ukvm_hv_kvm.c
+++ b/ukvm/ukvm_hv_kvm.c
@@ -29,7 +29,7 @@
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <string.h>
-
+#include <stdlib.h>
 #include <linux/kvm.h>
 
 #include "ukvm.h"
--- a/ukvm/ukvm_hv_kvm_x86_64.c
+++ b/ukvm/ukvm_hv_kvm_x86_64.c
@@ -31,7 +31,7 @@
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <string.h>
-
+#include <stdlib.h>
 #include <linux/kvm.h>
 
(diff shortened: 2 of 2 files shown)


ricarkol:
Thanks @thomaco 

@mato, @djwillia: will merge this as it looks safe. Also, this should be fine on Freebsd as well.

mato:
@thomaco Thanks, this looks fine. Not sure why it got missed and/or why the various CIs have not complained.

Which reminds me, it looks like the surf-build CIs did not run for this PR, I'll check on the Linux ones.

ukvm/ukvm_hv_kvm.c
```diff
@@ -29,7 +29,7 @@
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <string.h>
-
+#include <stdlib.h>
```
mato: Nit: This should be in the `#include` group above.

ukvm/ukvm_hv_kvm_x86_64.c
```diff
@@ -31,7 +31,7 @@
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <string.h>
-
+#include <stdlib.h>
```
mato: Nit: This should be in the `#include` group above.

PR #219, opened 2017-09-26
Add solo5_exec (for ukvm only)

ricarkol:
This change implements ukvm support for IncludeOS's [LiveUpdate](https://speakerdeck.com/amirmc/unikernels-where-are-they-now?slide=15) by implementing
a new proposed solo5 function called `solo5_exec`. This function gets a new unikernel as an ELF and executes it with `solo5_exec(ELF)`. Then, `ukvm` (not supported on muen nor virtio) loads this ELF in memory and restarts the execution.

IncludeOS is already pointing to Solo5:master, so this feature is not critical. But, ukvm is able to perform this operation in less than 1ms so this seems like the fastest known method of updating a running unikernel.

This was tested on x86_64 on Linux and Freebsd with two new tests: `test_exec` and `test_exec_failure`. A previous version of it was also tested with IncludeOS.


hannesm:
I haven't reviewed the code (sorry), but my opinion on such a feature is that it should be conditionally compiled in.  Providing an `exec` in a unikernel increases the attack surface tremendously (+16000 lines of C speak for themselves).  I'd be much more comfortable to not have such a hypercall within the deployed unikernel - at least for certain applications.

I'm also unsure about "live update" - shouldn't e.g. TCP connections etc. be migrated over to the new unikernel (likely this is out of scope for Solo5, but we should think within a broader perspective how to support such lifecycles - certainly this would need changes both in Solo5 and MirageOS)?

ricarkol:
Thanks @hannesm . Will work on moving the hypercall handler to a ukvm module. About the number of lines, it's mainly coming from a unikernel image that the tests are using, 16,159 lines from tests/test_exec/test_globals_ukvm_x86_64.h. Will work on generating it during setup with xxd (will disable the exec tests if xxd is not available on the system).

hannesm:
@ricarkol thanks! the +16000 lines C was an attempt of humour on my side, sorry for failing.  Nevertheless I'm a bit uneasy with having an `exec` hypercall in all my ukvm unikernels, and thanks for moving it into a module!

ricarkol:
@hannesm just moved the hypercall to a module. About auto-generating tests/test_exec/test_globals_ukvm_x86_64.h, it requires messing too much with our already brittle test scripts. Would rather not do it, but can do it if you and others (i.e., mato) want to.

mato:
This is an interesting feature. I've not reviewed the implementation in detail yet, since there are some key questions which I think we should answer before considering this for merging:

(Note that some of this touches on various discussions we've had either out-of-band or in other PRs/issues, so if something is not clear please ask for clarification.)

1. At a conceptual level, my view is that the presence of an `exec()` call significantly changes the trust relationship between the guest and the monitor, and the overall security model for a ukvm unikernel. Currently we have the option of providing guarantees about the _immutability_ of the guest code base, e.g. by enforcing _W^X_ and denying the guest the ability to map any new executable pages. By introducing `exec()` we lose that ability now and in the future (or at least make it "optional").

    At a higher level, while I understand the IncludeOS folks motivation for the live-update feature (needing to operate on existing cloud hypervisors independently of any higher-level orchestration stack that may or may not be provided), in the case of Solo5 I think this responsibility should be delegated to precisely such a higher-level orchestration stack running on the host, with ukvm only providing the bare minimum of a mechanism required to implement the functionality.

2. At an implementation level, we have discussed plans for further de-privileging the monitor (e.g. via seccomp) and/or implementing other targets in the future. It's important to consider the implications of providing `exec()` on these plans. To give an example, supporting `exec()` requires that the monitor process is _always_ able to call `mprotect()` to remap new pages as executable, which means that it _cannot_ drop these privileges after loading the initial ELF image and before starting to execute guest code since it would have no way of regaining them to perform the `exec()`.

It's plausible that including this functionality only as a module (as long as the core impact is minimal) is an acceptable compromise, as long as we can still provide the guarantees mentioned above when the module is not compiled in and it's clearly documented that asking for this module has these implications. However, at this point I'm not convinced that we won't paint ourselves into a corner by implementing this.

@ricarkol Minor implementation note: There is a set of magic `objdump` and `ld` runes that can be used to embed a binary at build time into the `test_exec` unikernel, I can dig that up if we decide to merge this.

ricarkol:
Thanks @mato.

This feature is mainly attractive because ukvm is able to do it "without downtime" (median of 590 microseconds from 30 samples). And it didn't really require much changes. The only two changes to the core of ukvm were: `ukvm_elf_load` now takes a buffer or a file, and FreeBSD:`ukvm_hv_vcpu_init` doesn't re-activate an already activated VM. Everything else is already in a module which implements `hypercall_exec`. This hypercall loads a new ELF in memory, calls `ukvm_hv_vcpu_init`, and returns to the exit loop.

I agree that this changes the threat model. Would it help if ukvm checked that the unikernel about to run is not malicious? right now, there is a big XXX in the middle of `hypercall_exec` pointing to where that check should be made. Also, FWIW, the proposed change checks for valid pointers and valid offsets. The behavior right now is that any funky pointer or offset kills the unikernel; no rollback or recovery is attempted. So, it seems that it should not be possible for the unikernel to craft a malicious ELF and access memory outside the unikernel memory.

About other ukvm backends, like seccomp or SGX enclaves, this hypercall would be a NOOP or an invalid call. I agree that it would be impossible or wouldn't make sense to implement exec there.

What about pushing this to master, but only compile and include the module for the two new automated tests? so we have automated tests forever and ever making sure that this feature doesn't rot, and maybe somebody finds a good use for it (*).

(*) @jmcadden is looking at porting v8 and libuv to IncludeOS, and it seems that startup times could be around 200ms (too large for serverless). Was thinking that we could do checkpoint/restore of the state right after initializing. And then realized that implementing restore could be conveniently done using `solo5_exec`.

mato:
> This feature is mainly attractive because ukvm is able to do it "without downtime" (median of 590 microseconds from 30 samples). And it didn't really require much changes. The only two changes to the core of ukvm were: ukvm_elf_load now takes a buffer or a file, and FreeBSD:ukvm_hv_vcpu_init doesn't re-activate an already activated VM. Everything else is already in a module which implements hypercall_exec. This hypercall loads a new ELF in memory, calls ukvm_hv_vcpu_init, and returns to the exit loop.

I understand the limited scope of the changes in the implementation at the _current_ point in time. What I'm concerned about are the wider consequences of these changes, specifically for a) further de-privileging the current ukvm implementation using `seccomp` and b) a future monitor targeting Linux userspace + `seccomp`. As far as I can see, for both of these cases (not just the userspace implementation!) the presence of a `solo5_exec()` hypercall makes it impossible to run the monitor with minimal privileges (e.g. disabling `mmap()` and friends).

Stepping back a bit; have you thought of other ways to implement a mechanism to provide comparable functionality where the updated unikernel can be provided by a higher-level component in the stack?

As a strawman example: We could have a hypercall to "exit and persist an opaque blob of state from guest memory", and a separate path to load that state at ukvm (re-)start time. That might be a way to provide this type of "just-in-time reload" functionality without the associated issues. Of course this also has implications on how any existing host resources would be re-acquired, but I think it's an angle worth exploring.

> Would it help if ukvm checked that the unikernel about to run is not malicious? [...] So, it seems that it should not be possible for the unikernel to craft a malicious ELF and access memory outside the unikernel memory.

It's not just about determining whether or not the ELF _file_ is malicious, it's about determining whether or not the replacement _code_ is trusted or not. In my mind at the level of ukvm/the monitor it should only be providing mechanism and not policy. Policy decisions should be left to a higher-level component.

> What about pushing this to master, but only compile and include the module for the two new automated tests? so we have automated tests forever and ever making sure that this feature doesn't rot, and maybe somebody finds a good use for it (*).

I'd rather have a clear user and consensus for this feature before merging it. Rarely-used-but-available features are a tried and true source of security bugs.

> (*) @jmcadden is looking at porting v8 and libuv to IncludeOS, and it seems that startup times could be around 200ms (too large for serverless). Was thinking that we could do checkpoint/restore of the state right after initializing. And then realized that implementing restore could be conveniently done using solo5_exec.

Ah, the "restore a checkpoint to get around slow runtime start-up" trick :-) This is probably best for an out-of-band discussion, but my general intuition is that once you get into the details of the actual implementation it turns out that the extra complexity is not worth it.

adamsteen:
Sorry if this is out of line, but thought I should mention it. The OpenBSD port of ukvm would prevent this hypercall from happening as it is, chroot’d, prevledge droop and pledged(stdio) before booting the kernel. Yes these could be removed, but wouldn’t think that is a good idea.

ricarkol:
@mato, @adamsteen: answers are below, but here is the plan. Will measure doing the replacement by restarting ukvm. If that time is good enough, will take a closer look at @hanness's vmmd. If the time is not good enough, then will move this PR for after we have userspace+seccomp support. If solo5_exec doesn't look good on top of userspace+seccomp, will drop this change. If it's good enough, will try to convince everybody again (and think of what to do about introducing policy decision into ukvm).

> Stepping back a bit; have you thought of other ways to implement a mechanism to provide comparable functionality where the updated unikernel can be provided by a higher-level component in the stack?

> As a strawman example: We could have a hypercall to "exit and persist an opaque blob of state from guest memory", and a separate path to load that state at ukvm (re-)start time. That might be a way to provide this type of "just-in-time reload" functionality without the associated issues. Of course this also has implications on how any existing host resources would be re-acquired, but I think it's an angle worth exploring.

Restarting ukvm is something I was trying to avoid as that will increase the replacement time by at least 5-10 ms. Will try to get a good approximation to the real number.

> Would it help if ukvm checked that the unikernel about to run is not malicious? [...] So, it seems that it should not be possible for the unikernel to craft a malicious ELF and access memory outside the unikernel memory.
> It's not just about determining whether or not the ELF file is malicious, it's about determining whether or not the replacement code is trusted or not. In my mind at the level of ukvm/the monitor it should only be providing mechanism and not policy. Policy decisions should be left to a higher-level component.

I agree that this policy decision should not be made by ukvm. The first option is to leave the decision to the unikernel itself, but i'm not liking that option that much. By enforcing W^X we make it hard to run new code, but we are not doing anything against ROP attacks. The issue is that an attacker could ROP to `solo5_exec` and actually run new code. The second option is to leave the decision to the deployer. @hannesm's vmmd (https://hannes.nqsb.io/Posts/VMM) is looking ideal for doing this.

> I'd rather have a clear user and consensus for this feature before merging it. Rarely-used-but-available features are a tried and true source of security bugs.

The key difference though is that most of this feature is in a module that can be left out by default (not even linked in by default).

> Ah, the "restore a checkpoint to get around slow runtime start-up" trick :-) This is probably best for an out-of-band discussion, but my general intuition is that once you get into the details of the actual implementation it turns out that the extra complexity is not worth it.

Yeap, don't really know how to solve that one. Let's discuss this out-of-band when we know more about what the bottlenecks are.

> I understand the limited scope of the changes in the implementation at the current point in time. What I'm concerned about are the wider consequences of these changes, specifically for a) further de-privileging the current ukvm implementation using seccomp and b) a future monitor targeting Linux userspace + seccomp. As far as I can see, for both of these cases (not just the userspace implementation!) the presence of a solo5_exec() hypercall makes it impossible to run the monitor with minimal privileges (e.g. disabling mmap() and friends).

We have a prototype implementation of userspace+seccomp. What about moving this PR to after we have the seccomp changes ready? I can keep this solo5_exec updated, so we can come back to it and see how things look like on top of userspace+seccomp. If things look ugly I will be the first to drop this; the seccomp change and keeping things simple should have higher priority.

> Sorry if this is out of line, but thought I should mention it. The OpenBSD port of ukvm would prevent this hypercall from happening as it is, chroot’d, prevledge dropped and pledged(stdio) before booting the kernel. Yes these could be removed, but wouldn’t think that is a good idea.

Definitely not out of line @adamsteen. I was hoping OpenBSD would be similar to FreeBSD. We could disable the feature for OpenBSD, like it's being disabled for ARM and muen at the moment. Disabling for ukvm means that the code is not even there (so it's more than just a flag).

mato:
@ricarkol Your plan looks reasonable. We can discuss the other aspects out-of-band at some stage.

mato:
@ricarkol AFAIR our consensus on this back in December was that it's not a good idea, can we close this?

ricarkol:
Yes, let's close it.

Issue #223, opened 2017-12-07
move malloc from Solo5 to libOS layer


djwillia:
Solo5 currently contains a malloc implementation.  In the MirageOS case, it is used by the libOS via ocaml-freestanding.  In the IncludeOS case, they have their own malloc, which gets memory from the Solo5 malloc.

I think malloc should be the responsibility of the libOS layer.  This has the advantage of removing duplicate functionality (in the IncludeOS case).  Also, it removes the malloc family of calls from the solo5 interface (solo5.h) and bringing it closer to alignment with the underlying isolation interface (e.g., the set of ukvm hypercalls).

I've prepared a PR for Solo5 and for ocaml-freestanding, which applies to the MirageOS case.  @ricarkol I would like your opinion on whether this will be suitable for IncludeOS as well.

ricarkol:
Thanks for the change @djwillia . It makes things easier on the IncludeOS side. We are currently using the IncludeOS malloc() which uses all of the solo5 heap.

mato:
@ricarkol Can you point me to the relevant code in the IncludeOS repository that initialises the heap when running on Solo5?

ricarkol:
@mato here it is:

https://github.com/hioa-cs/IncludeOS/blob/master/src/platform/x86_solo5/kernel_start.cpp#L72
```
mem_size = (uintptr_t)get_cpu_ebp();
...
kernel_start()
    OS::start(cmdline, mem_size)
        OS::memory_end_ = mem_size;
        OS::heap_max_ = OS::memory_end_;
        // OS::heap_begin is really reinterpret_cast<uintptr_t>(&_end);
```

ricarkol:
As you can see, using `ebp` as the heap end won't work on the seccomp case.

mato:
I've finally found some time to think about the implications of this change in depth, both from a design and implementation (WIP in #224, mirage/ocaml-freestanding#29) point of view.

To set the stage: I agree that a _high-level_ memory allocator (as defined by the _malloc()_ family of functions) need not be part of the Solo5 API, and we gain the following benefits by removing it:

- the application is free to manage memory as it sees fit in a way that is tailored to its requirements. This will be an advantage especially in the case of managed language runtimes,
- the Solo5 API and implementation both shrink in size, which is good.

However: I think that Solo5 should retain some control over the _low-level_ memory layout and allocation semantics (i.e. how _malloc()_ gets its memory at runtime). Rationale:

[...]

PR #224, opened 2017-12-07
remove malloc from solo5

djwillia:
This is part of addressing issue #223.
Most dynamic memory usage was in the virito backend, with some use in IRQ handlers.  

In the IRQ case, I have replace the dynamic memory usage in a queuing implementation to use static buffers.

In the virtio case, I have introduced a new function (`alloc_chunk_4K`) which dynamically adds a chunk of 4K aligned memory to the "bottom of the heap".  This function is only intended to be called before passing control to the libOS, which will see its available memory starting at a dynamically determined point based on how much `alloc_chunk_4K`ed memory was done by Solo5 at initialization time.  (I had originally tried to use large static buffers for virtio, but these dramatically reduced performance on QEMU for reasons I don't understand.)

 The libOS can discover the start location and size of the heap via a new Solo5 call: `solo5_get_info`.


kernel/kernel.h
```diff
@@ -78,6 +78,7 @@ void intr_irq_handler(uint64_t irq);
 
 /* mem.c: low-level page alloc routines */
 void mem_init(void);
+void *alloc_chunk_4K(size_t size);
```
ricarkol: Can you add a comment there saying that this function should only be called before `solo5_app_main`? as doing so will most likely mess with the unikernel heap. In fact, it would be nice to add some assert or some mechanism that ensures that.

kernel/mem.c
```diff
@@ -55,23 +55,28 @@ void mem_init(void)
...
 
-    /*
-     * dlmalloc guarantees increment values less than half of size_t, so this
-     * is safe from overflow.
-     */
-    brk += increment;
+    /* for 4K alignment */
+    prev = brk = brk + (0x1000 - (brk & 0xfff));
+
+    brk += size;
```
ricarkol: This can overflow. There should not be a `size` greater than 1k or something like that right? you could assert that.

kernel/solo5.h
```diff
@@ -38,6 +38,16 @@
  */
 int solo5_app_main(char *cmdline);
 
+/* 
+ * Platform information.  Currently only memory info.
+ */
+struct solo5_info {
+    char *cmdline;
```
ricarkol: Is the plan to stop passing cmdline thorugh `solo5_app_main`?

mato:
I'd prefer a slightly more elaborate set of interfaces:

Instead of:

```
void *alloc_chunk_4K(size_t size);
void solo5_get_info(struct solo5_info *info);
```

I propose:

```
void *mem_ialloc_pages(size_t num);
int solo5_mem_claim_heap(void *heap_start, void *heap_end);
```

Rationale:

1. The private `alloc_chunk_4K()` is now a (limited) page allocator in disguise, so let's just call it that. This also simplifies the implementation, and makes it immediately clear to the user that pages are the unit of allocation (which implies page alignment). The `i` in the `ialloc` refers to the limitation that the function is intended to be called only at _initialisation_ time, i.e. before `solo5_mem_claim_heap()` is called by the libOS. Any later calls to this function will panic (assert), and that should be documented.

2. For the public interface, to me `solo5_get_info()` means nothing by itself and does not reflect the side-effects it should enforce (be called only once and disable the private allocator from that point). Hence the use of `claim` in the name, which should also encourage anyone porting a libOS to look the function up in the documentation to understand the semantics.

What do you think?

kernel/mem.c
ricarkol: The name suggests that the memory returned will be rounded up to 4k. No caller is expecting that right?

mato:
Typo in the previous comment -- the public interface should be

```
int solo5_mem_claim_heap(void **start, void **end);
```

The function will return success (0) on the initial call only and failure (-1) on all further calls.

djwillia:
@mato I agree with the naming of `ialloc` and the invariant that it should only be called at solo5 initialization time.  

I don't think that the initialization time locking of the heap should be exposed to the user.  What about a private call right before `solo5_app_main` is called that ends initialization?  That way the user visible API doesn't have a weird side effect, and user doesn't need to claim anything.

Then we just need a way to get "information" about where the heap starts and ends.  I had called it info to ideally align it with some of the device info calls.  Right now it is only memory, so we could call it mem_info, but I don't see why it shouldn't be more generic info about the platform.  For example, it might have the number of each type of device.  

@ricarkol sorry I had meant to remove `cmdline` from the `struct solo5_info`; I think it should stay in `solo5_app_main` for now for cmdline parsing reasons.

djwillia:
After commit 0eafe28...

@ricarkol I think I have addressed your concerns now.

@mato I've addressed part of your concern with `mem_ialloc_pages`, but don't have a user-facing "claim" call.  I do have an assertion and mechanism to lock the memory layout via `mem_lock_heap`.  What do you think?

ricarkol:
@djwillia ACK on the new solo5_info struct (without the cmdline field). LGTM

Kensan:
Playing the devil's advocate: since **mem_ialloc_pages** is exclusively required by virtio, is there a way to move it from the general kernel to the virtio-specific implementation? Otherwise all implementations "pay the price" even though the functionality of (page-granular) memory allocation is not needed by ukvm and Muen...

djwillia:
@Kensan I would prefer what you suggest and I had actually tried to remove all dynamic memory allocation from Solo5.  In the virtio case there are some potentially large buffers (depending on the backend virtio implementation) and when I had tried with static (large) buffers QEMU started behaving strangely (taking ~5 seconds to boot).  I didn't get to the bottom of that, so this simple dynamic allocation was the simplest I could come up with that did not affect the QEMU boot time.  If you know of a way to remove it and not cause QEMU to freak out I'm all for it.

mato:
@djwillia _mem_ialloc_pages()_ needs to do the following, and nothing else:
```
void *mem_ialloc_pages(size_t num)
{
    assert(!mem_locked);
    
    uint64_t prev = heap_start;
    heap_start += num << PAGE_SHIFT;
    assert(heap_start < (unit64_t)&prev);

    return (void *)prev;
}
```

It's an internal function, so we don't care about overflow, only that we're not allocating more memory than we have in low-memory situations. All the stack guard and related code can go, it doesn't make sense here -- I will need to move it into ocaml-freestanding for Mirage.

Regarding your proposed `solo5_get_info()` interface, introducing kitchen-sink style calls like that is almost always a bad idea. I need to think about this some more, but I'm leaning towards supplying the start and end of the "heap" as arguments to `solo5_app_main()` instead.

@Kensan Regarding the cost of the implementation, it's basically those five lines of code above, so not  worth splitting out (and having separate `mem.c`'s again).

kernel/mem.c
```diff
@@ -32,6 +40,10 @@ void mem_init(void)
     mem_size = platform_mem_size();
     heap_start = ((uint64_t)&_end + PAGE_SIZE - 1) & PAGE_MASK;
     heap_top = heap_start;
+
+    /* assert heap is page-aligned (4K boundaries) */
+    assert(!(heap_top & 0xfff));
+
```
mato: Not needed. heap_start _is_ page aligned.

kernel/mem.c
```diff
@@ -24,6 +24,14 @@ static uint64_t heap_start;
 static uint64_t heap_top;
 static uint64_t stack_guard_size;
 
+/* 
+ * We lock the memory layout (disabling mem_ialloc_pages) after Solo5
+ * initialization but before passing control to the application via
+ * solo5_app_main().
+ */
+static int mem_locked = 0;
+void mem_lock_heap(void) { mem_locked = 1; }
```
mato: Style: Not on one line.

kernel/virtio/virtio_blk.c
```diff
@@ -164,8 +165,10 @@ void virtio_config_block(struct pci_config_info *pci)
 
     virtq_init_rings(pci->base, &blkq, 0);
 
-    blkq.bufs = calloc(blkq.num, sizeof (struct io_buffer));
+    pgs = (blkq.num * sizeof (struct io_buffer) + PAGE_SIZE - 1) / PAGE_SIZE;
```
mato: Use `((size) - 1 >> PAGE_SHIFT) + 1` to go from `(size)` bytes to pages.

kernel/virtio/virtio_blk.c
```diff
@@ -164,8 +165,10 @@ void virtio_config_block(struct pci_config_info *pci)
...
+    blkq.bufs = mem_ialloc_pages(pgs);
     assert(blkq.bufs);
+    memset(blkq.bufs, 0, pgs * PAGE_SIZE);
```
mato: Use `pages << PAGE_SHIFT`.

kernel/virtio/virtio_net.c
```diff
@@ -182,10 +183,15 @@ void virtio_config_network(struct pci_config_info *pci)
     virtq_init_rings(pci->base, &recvq, VIRTQ_RECV);
     virtq_init_rings(pci->base, &xmitq, VIRTQ_XMIT);
 
-    recvq.bufs = calloc(recvq.num, sizeof (struct io_buffer));
+    pgs = (recvq.num * sizeof (struct io_buffer) + PAGE_SIZE - 1) / PAGE_SIZE;
```
mato: As in `virtio_config_block`, use bit shifting in the same calculations in this function.

kernel/virtio/virtio_ring.c
```diff
@@ -87,14 +87,16 @@ int virtq_add_descriptor_chain(struct virtq *vq,
...
     uint8_t *data;
-
+    size_t pgs;
+    
     outw(pci_base + VIRTIO_PCI_QUEUE_SEL, selector);
     vq->last_used = vq->next_avail = 0;
     vq->num = vq->num_avail = inw(pci_base + VIRTIO_PCI_QUEUE_SIZE);
 
-    data = memalign(4096, VIRTQ_SIZE(vq->num));
+    pgs = (VIRTQ_SIZE(vq->num) + PAGE_SIZE - 1) / PAGE_SIZE;
```
mato: Same here...

Kensan:
@djwillia I see, do you still have the static buffer code variant somewhere so I might take a look? Maybe alignment of the **struct io_buffer**s had an adverse effect on performance? That said, I certainly do not want to hold up this PR with my "nitpick".

djwillia:
thanks @mato.  I updated to use shifts and simplified `mem_ialloc_pages` as you suggested in cc3117c

I think that passing the `heap_start` and `heap_end` through `solo5_app_main` can cause additional complexity for the ocaml-freestanding case because it means that mirage-solo5 (which implements `solo5_app_main`) must change and we must figure out how to get the information to ocaml-freestanding's sbrk/malloc.  Is your issue with (a) the generality of `solo5_get_info` (which could perhaps be solved by naming it `solo5_get_heap_info` or something), or (b) the existence of the call in the Solo5 API (e.g., getting information about the system that is not passed as a boot parameter)? 

@Kensan I'm afraid I don't have that version lying around.  I didn't even need to use it for `struct io_buffers` to see the issue.  Just the existence of the static buffer was enough to cause my QEMU boot slowness.  If you want to see if QEMU freaks out for you put a big static array of a few MB in size in any of the C files and see what happens to boot time.

mato:
> Is your issue with (a) the generality of solo5_get_info (which could perhaps be solved by naming it solo5_get_heap_info or something), or (b) the existence of the call in the Solo5 API (e.g., getting information about the system that is not passed as a boot parameter)?

Both.

(a) is a general issue with naming things, not only related to the problem we are discussing. Calling an API `solo5_get_info()` means nothing to me (what "information"?).

You are correct that (b) is an issue with the existence of the call itself. Thinking about it, if we remove `malloc()` from the Solo5 API then _it does not make sense_ for the application/libOS to run without having that information [about where it's heap should go]. Therefore, it's logical to _always_ pass in as arguments to `solo5_app_main()` and a `solo5_get_info()` call is superfluous.

Regarding modifying the ocaml-freestanding/mirage-solo5 dance, a simple and correct solution would be -- move `solo5_app_main(...)` to ocaml-freestanding, have it call `mirage_app_main(cmdline)` which would be in mirage-solo5.

mato:
> Regarding modifying the ocaml-freestanding/mirage-solo5 dance, a simple and correct solution would be -- move solo5_app_main(...) to ocaml-freestanding, have it call mirage_app_main(cmdline) which would be in mirage-solo5.

... or ... even better, don't have a main-type function in mirage-solo5 at all and have ocaml-freestanding call `caml_startup()` directly. Then we just need an interface to get the command line from ocaml-freestanding into mirage-solo5. So this might be an argument for having a `solo5_boot_info()` call or similar. I'll think about it some more.

djwillia:
Ah, I went down the route of moving things from mirage-solo5 into ocaml-freestanding and ran into the issue that you mention (related to the command line and the mirage-solo5/ocaml-freestanding relationship).  In the end I thought it was simpler to do it this way, with the added benefit of not needing changes to mirage-solo5 for this PR.  I figured we could figure out the correct relationship between ocaml-freestanding and mirage-solo5 at a later date if necessary.

More generally, I don't think I agree that stuffing everything into arguments to `solo5_app_main` is necessarily the right solution.  I like the symmetry of having a set of `_info` calls about parts of the underlying platform, and I almost think of them like the modules in ukvm.  We have `net_info`, `blk_info`, and I what I had in mind was almost like a `core_info`.

I suggest we rename the function to something that conveys some meaning, but leave the structure as is for now so we don't need to modify the ocaml-freestanding and mirage-solo5 relationship in this PR.  Then we can open an issue to clarify that relationship if we're unhappy with it.  What do you think?

mato:
Ok, can we compromise for now on a `mem_info`? I think it should accept/return a `uint64_t *heap_start` and a `size_t *mem_size`, or put them in a `struct` if you want. The reason for `mem_size` rather than a `heap_end` is that the maximum for the heap is dependent on how we manage heap vs. stack, and in this scheme we are relinquishing control of that to the libos layer. Does that work for you?

djwillia:
Thanks @mato, that sounds good.  I've updated this one and the accompanying https://github.com/mirage/ocaml-freestanding/pull/29

mato:
@djwillia Thanks. This looks fine. Will need to coordinate any merging with #201, so it'll be a bit more time before that happens.

mato:
@djwillia Not sure what you did here,  but you appear to have merged master into this branch rather than rebased it on top of master. Can you try and fix (use a temporary local branch to verify the result)?

danlythemanly:
@mato thanks; I'm not sure what I did either :)  I think this should be better?

mato:
This looks better, thanks!

mato:
Superseded by #244.

Issue #206, opened 2017-07-15
Porting ukvm to OpenBSD vmm APIs


adamsteen:
Hi

I am raising this issue after a post on Unikernel Devel, @hannesm said it would be preferable to discuss the situation here.

I am in the process of porting Solo5 to OpenBSD. I have successfully built the virtio kernel and tests using clang.

But when attempting to boot an image (created via mkimage on an alpine linux install) it fails with.

```
Booting from Hard Disk...

Booting from 0000:7c00

SYSLINUX 6.03 EDD 2014-10-06 Copyright (C) 1994-2014 H. Peter Anvin et al
enter handle_12:
   a=00000000  b=00000000  c=00000000  d=00000000 ds=0000 es=0000 ss=0000
  si=00000000 di=00000000 bp=00000000 sp=00007b2e cs=0000 ip=8c2d  f=0000
enter handle_12:
   a=00000200  b=00000000  c=00000000  d=00000000 ds=0000 es=0000 ss=0000
  si=00000000 di=00000000 bp=00000000 sp=00007b2e cs=0000 ip=8c2d  f=0000

SYSLINUX 6.03 2014-10-06 Copyright (C) 1994-2014 H. Peter Anvin et al
Loading unikernel.bin... ok

enter handle_12:
   a=00000000  b=00000000  c=00000000  d=00000000 ds=0000 es=0000 ss=0000
  si=00000000 di=00000000 bp=00000000 sp=00007b0e cs=0000 ip=8c2d  f=0000
Invalid Multiboot image: neither ELF header nor a.out kludge found

boot:
```

I expect this is due to an old binutils.
The output of "ld --version"
```
GNU ld version 2.17
Copyright 2005 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License.  This program has absolutely no warranty.
```

The output of readelf -S for both Linux and OpenBSD (these are different) is as follows
(built and readelf on the specified platform)

readelf -S test_hello.virtio

__OpenBSD__ 
```
There are 19 section headers, starting at offset 0x12f1d0:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .text             PROGBITS         0000000000100000  00100000
       0000000000009169  0000000000000000  AX       0     0     16
[...]

PR #226, opened 2017-12-08
Porting ukvm to OpenBSD vmm APIs

adamsteen:
All the changes to get Solo5 working on OpenBSD current.

I have not been able to get "ocaml-freestanding" working yet.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -43,16 +43,22 @@ cc_is_gcc()
     cc_maybe_gcc && ! cc_is_clang
 }
 
+ld_is_lld()
+{
+    ${LD} --version 2>&1 | grep -q '^LLD'
+}
+
 # Allow external override of CC.
 # TODO: This needs further work to provide full support for cross-compiling and
 # correctly pass through to ukvm-configure where required.
 CC=${CC:-cc}
+LD=${LD:-ld}
 
 TARGET=$(${CC} -dumpmachine)
 [ $? -ne 0 ] &&
     die "Error running '${CC} -dumpmachine', is your compiler working?"
 case ${TARGET} in
-    x86_64-*)
+    x86_64-*|amd64-*)
 	TARGET_ARCH=x86_64
         ;;
     aarch64-*)
@@ -124,6 +130,37 @@ case $(uname -s) in
         BUILD_VIRTIO="yes"
         BUILD_MUEN="yes"
         ;;
+    OpenBSD)
+        # On OpenBSD/clang we use -nostdlibinc which gives us access to the
+        # clang-provided headers for compiler instrinsics. We copy the rest
+        # (std*.h, cdefs.h and their dependencies) from the host.
+        cc_is_clang || die "Only 'clang' is supported on OpenBSD"
+        [ "${TARGET_ARCH}" = "x86_64" ] ||
+            die "Only 'x86_64' is supported on OpenBSD"
+        if ! ld_is_lld; then
+            LD='/usr/bin/ld.lld'
+            echo "using GNU 'ld' is not supported on OpenBSD, falling back to 'ld.lld'"
+            [ -e ${LD} ] || die "/usr/bin/ld.lld does not exist"
+        fi
+        INCDIR=/usr/include
+        SRCS_MACH="machine/_float.h machine/endian.h machine/cdefs.h machine/_types.h"
+        SRCS_SYS="sys/_null.h sys/cdefs.h sys/_endian.h sys/endian.h sys/_types.h"
+        SRCS_AMD64="amd64/_float.h amd64/stdarg.h amd64/endian.h"
+        SRCS="float.h stddef.h stdint.h stdbool.h stdarg.h"
+
+        mkdir -p ${HOST_INCDIR}
+        mkdir -p ${HOST_INCDIR}/machine ${HOST_INCDIR}/sys ${HOST_INCDIR}/amd64
+        for f in ${SRCS_MACH}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/machine; done
+        for f in ${SRCS_SYS}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/sys; done
+        for f in ${SRCS_AMD64}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/amd64; done
+        for f in ${SRCS}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}; done
+
+        HOST_CFLAGS="-fno-pie -fno-stack-protector -nostdlibinc"
+        HOST_LDFLAGS="-nopie"
+        BUILD_UKVM="yes"
+        BUILD_VIRTIO="no" # has not been tested
+        BUILD_MUEN="no" # has not been tested
+        ;;
     *)
         die "Unsupported build OS: $(uname -s)"
         ;;
@@ -135,6 +172,8 @@ BUILD_UKVM=${BUILD_UKVM}
 BUILD_VIRTIO=${BUILD_VIRTIO}
 BUILD_MUEN=${BUILD_MUEN}
 HOST_CFLAGS=${HOST_CFLAGS}
+HOST_LDFLAGS=${HOST_LDFLAGS}
 TARGET_ARCH=${TARGET_ARCH}
 CC=${CC}
+LD=${LD}
 EOM
--- a/tests/setup-tests.sh
+++ b/tests/setup-tests.sh
@@ -42,6 +42,13 @@ FreeBSD)
     sysctl -w net.link.tap.up_on_open=1
     ifconfig tap100 create 10.0.0.1/24 link0 up
     ;;
+OpenBSD)
+    (
+         cd /dev
+         ./MAKEDEV tap100
+    )
+    ifconfig tap100 inet 10.0.0.1 netmask 255.255.255.0
+    ;;
 *)
     exit 1
     ;;
--- a/ukvm/ukvm_hv_openbsd.c
+++ b/ukvm/ukvm_hv_openbsd.c
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file
+ *
+ * This file is part of ukvm, a unikernel monitor.
+ *
+ * Permission to use, copy, modify, and/or distribute this software
+ * for any purpose with or without fee is hereby granted, provided
+ * that the above copyright notice and this permission notice appear
+ * in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * ukvm_hv_openbsd.c: Architecture-independent part of OpenBSD vmm(4) backend
+ * implementation.
+ */
+
+#include <assert.h>
+#include <err.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#define _WITH_DPRINTF
+#include <stdio.h>
+
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#include <machine/vmmvar.h>
+#include <sys/param.h>
+
+#include "ukvm.h"
+#include "ukvm_hv_openbsd.h"
+
+/*
+ * TODO: To ensure that the VM is correctly destroyed on shutdown (normal or
+ * not) we currently install an atexit() handler. The top-level API will need
+ * to be changed to accomodate this, e.g. by introducing a ukvm_hv_shutdown(),
+ * however this is incompatible with the current "fail fast" approach to
+ * internal error handling.
+ */
+static struct ukvm_hv *cleanup_hv;
+
+static void cleanup_vmd_fd(void)
+{
+    if (cleanup_hv != NULL && cleanup_hv->b->vmd_fd != -1)
+        close(cleanup_hv->b->vmd_fd);
+}
+
+static void cleanup_vm(void)
+{
+    if (cleanup_hv != NULL && cleanup_hv->b->vcp_id != -1) {
+        struct vm_terminate_params vtp = { .vtp_vm_id = cleanup_hv->b->vcp_id };
+        if (ioctl(cleanup_hv->b->vmd_fd, VMM_IOC_TERM, &vtp) < 0)
+            err(1, "terminate vmm ioctl failed - still exiting");
+    }
+}
+
+struct ukvm_hv *ukvm_hv_init(size_t mem_size)
+{
+    struct ukvm_hv          *hv;
+    struct ukvm_hvb         *hvb;
+    struct vm_create_params *vcp;
+    struct vm_mem_range *vmr;
+    void *p;
+    
+    if(geteuid() != 0) {
+        errno = EPERM;
+        err(1, "need root privileges");
+    }
+
+    hv = calloc(1, sizeof (struct ukvm_hv));
+    if (hv == NULL)
+        err(1, "calloc hv");
+
+    hvb = calloc(1, sizeof (struct ukvm_hvb));
+    if (hvb == NULL)
+        err(1, "calloc");
+
+    hv->b = hvb;
+    hvb->vmd_fd = -1;
+
+    hvb->vmd_fd = open(VMM_NODE, O_RDWR);
+    if (hvb->vmd_fd == -1)
+        err(1, "VMM_NODE");
+
+    cleanup_hv = hv;
+    atexit(cleanup_vmd_fd);
+
+    vcp = calloc(1, sizeof (struct vm_create_params));
+    if (vcp == NULL)
+        err(1, "calloc");
+
+    vcp->vcp_ncpus = 1; // vmm only supports 1 cpu for now.
+    int size = snprintf(vcp->vcp_name, VMM_MAX_NAME_LEN, "ukvm%d", getpid());
+    if (size == -1)
+        err(1, "snprintf");
+
+    vcp->vcp_nmemranges = 1;
+    vcp->vcp_memranges[0].vmr_gpa = 0x0;
+    vcp->vcp_memranges[0].vmr_size = mem_size;
+
+    vmr = &vcp->vcp_memranges[0];
+    p = mmap(NULL, vmr->vmr_size, PROT_READ | PROT_WRITE,
+        MAP_PRIVATE | MAP_ANON, -1, 0);
+    if (p == MAP_FAILED)
+        err(1, "mmap");
+
+    vmr->vmr_va = (vaddr_t)p;
+    hv->mem = p;
+    hv->mem_size = mem_size;
+
+    if (ioctl(hvb->vmd_fd, VMM_IOC_CREATE, vcp) < 0)
+        err(1, "create vmm ioctl failed - exiting");
+
+    hvb->vcp_id = vcp->vcp_id;
+    hvb->vcpu_id = 0; // the first and only cpu is at 0
+
+    atexit(cleanup_vm);
+
+    return hv;
+}
+
--- a/ukvm/ukvm_hv_openbsd_x86_64.c
+++ b/ukvm/ukvm_hv_openbsd_x86_64.c
@@ -1,0 +1,240 @@
+/*
+ * Copyright (c) 2015-2017 Contributors as noted in the AUTHORS file
+ *
+ * This file is part of ukvm, a unikernel monitor.
+ *
+ * Permission to use, copy, modify, and/or distribute this software
+ * for any purpose with or without fee is hereby granted, provided
+ * that the above copyright notice and this permission notice appear
+ * in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * ukvm_hv_openbsd_x86_64.c: x86_64 architecture-dependent part of OpenBSD
+ * vmm(4) backend implementation.
+ */
+
+#define _GNU_SOURCE
+#include <assert.h>
+#include <err.h>
+#include <errno.h>
+#include <inttypes.h>
+#include <pwd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <sys/ioctl.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#include <machine/cpu.h>
+#include <machine/vmmvar.h>
+#include <machine/specialreg.h>
+
+#include "ukvm.h"
+#include "ukvm_hv_openbsd.h"
+#include "ukvm_cpu_x86_64.h"
+
+static struct vcpu_segment_info sreg_to_vsi(const struct x86_sreg *);
+
+static uint64_t get_tsc_freq(void)
+{
+    int invariant_tsc;
+    static const int invar_mib[2] = { CTL_MACHDEP, CPU_INVARIANTTSC };
+    static const int freq_mib[2] = { CTL_MACHDEP, CPU_TSCFREQ };
+    uint64_t tsc_freq;
+    size_t len;
+
+    len = sizeof(invariant_tsc);
+    if (sysctl(invar_mib, 2, &invariant_tsc, &len, NULL, 0) == -1)
+        err(1, "sysctl invariant_tsc");
+
+    if(invariant_tsc) {
+        len = sizeof(tsc_freq);
+        if (sysctl(freq_mib, 2, &tsc_freq, &len, NULL, 0) == -1)
+            err(1, "sysctl tsc_freq");
+    } else {
+        errx(1, "Host TSC is not invariant, cannot continue");
+    }
+
+    return tsc_freq;
+}
+
+static struct vcpu_segment_info sreg_to_vsi(const struct x86_sreg *sreg)
+{
+    struct vcpu_segment_info vsi = {
+        .vsi_sel = sreg->selector * 8,
+        .vsi_limit = sreg->limit,
+        .vsi_ar = (sreg->type
+            | (sreg->s << 4)
+            | (sreg->dpl << 5)
+            | (sreg->p << 7)
+            | (sreg->l << 13)
+            | (sreg->db << 14)
+            | (sreg->g << 15)
+            | (sreg->unusable << X86_SREG_UNUSABLE_BIT)),
+        .vsi_base = sreg->base
+    };
+    return vsi;
+}
+
+void ukvm_hv_mem_size(size_t *mem_size) {
+    ukvm_x86_mem_size(mem_size);
+}
+
+void ukvm_hv_vcpu_init(struct ukvm_hv *hv, ukvm_gpa_t gpa_ep,
+        ukvm_gpa_t gpa_kend, char **cmdline)
+{
+    struct ukvm_hvb *hvb = hv->b;
+
+    struct vm_resetcpu_params vrp = {
+        .vrp_vm_id = hvb->vcp_id,
+        .vrp_vcpu_id = hvb->vcpu_id,
+        .vrp_init_state = {
+            .vrs_gprs[VCPU_REGS_RFLAGS] = X86_RFLAGS_INIT,
+            .vrs_gprs[VCPU_REGS_RIP] = gpa_ep,
+            .vrs_gprs[VCPU_REGS_RSP] = hv->mem_size - 8,
+            .vrs_gprs[VCPU_REGS_RDI] = X86_BOOT_INFO_BASE,
+            .vrs_crs[VCPU_REGS_CR0] = X86_CR0_INIT,
+            .vrs_crs[VCPU_REGS_CR3] = X86_CR3_INIT,
+            .vrs_crs[VCPU_REGS_CR4] = X86_CR4_INIT,
+            .vrs_sregs[VCPU_REGS_CS] = sreg_to_vsi(&ukvm_x86_sreg_code),
+            .vrs_sregs[VCPU_REGS_DS] = sreg_to_vsi(&ukvm_x86_sreg_data),
+            .vrs_sregs[VCPU_REGS_ES] = sreg_to_vsi(&ukvm_x86_sreg_data),
+            .vrs_sregs[VCPU_REGS_FS] = sreg_to_vsi(&ukvm_x86_sreg_data),
+            .vrs_sregs[VCPU_REGS_GS] = sreg_to_vsi(&ukvm_x86_sreg_data),
+            .vrs_sregs[VCPU_REGS_SS] = sreg_to_vsi(&ukvm_x86_sreg_data),
+            .vrs_gdtr = { 0x0, X86_GDTR_LIMIT, 0x0, X86_GDT_BASE},
+            .vrs_idtr = { 0x0, 0xFFFF, 0x0, 0x0},
+            .vrs_sregs[VCPU_REGS_LDTR] = sreg_to_vsi(&ukvm_x86_sreg_unusable),
+            .vrs_sregs[VCPU_REGS_TR] = sreg_to_vsi(&ukvm_x86_sreg_tr),
+            .vrs_msrs[VCPU_REGS_EFER] = X86_EFER_INIT,
+            .vrs_msrs[VCPU_REGS_STAR] = 0ULL,
+            .vrs_msrs[VCPU_REGS_LSTAR] = 0ULL,
+            .vrs_msrs[VCPU_REGS_CSTAR] = 0ULL,
+            .vrs_msrs[VCPU_REGS_SFMASK] = 0ULL,
+            .vrs_msrs[VCPU_REGS_KGSBASE] = 0ULL,
+            .vrs_crs[VCPU_REGS_XCR0] = XCR0_X87
+        }
+    };
+
+    ukvm_x86_setup_gdt(hv->mem);
+    ukvm_x86_setup_pagetables(hv->mem, hv->mem_size);
+
+    struct ukvm_boot_info *bi =
+        (struct ukvm_boot_info *)(hv->mem + X86_BOOT_INFO_BASE);
+    bi->mem_size = hv->mem_size;
+    bi->kernel_end = gpa_kend;
+    bi->cmdline = X86_CMDLINE_BASE; 
+    bi->cpu.tsc_freq = get_tsc_freq();
+
+    if (ioctl(hvb->vmd_fd, VMM_IOC_RESETCPU, &vrp) < 0)
+        err(1, "Cannot reset VCPU - exiting.");
+
+    *cmdline = (char *)(hv->mem + X86_CMDLINE_BASE);
+}
+
+int ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
+{
+    struct ukvm_hvb         *hvb = hv->b;
+    struct vm_run_params    *vrp;
+    struct passwd *pw;
+    uid_t uid;
+    gid_t gid;
+
+    vrp = malloc(sizeof(struct vm_run_params));
+    if (vrp == NULL)
+        err(1, "malloc vrp");
+
+    vrp->vrp_exit = malloc(sizeof(union vm_exit));
+    if (vrp == NULL)
+        err(1, "malloc vrp_exit");
+
+    vrp->vrp_vm_id = hvb->vcp_id;
+    vrp->vrp_vcpu_id = hvb->vcpu_id;
+    vrp->vrp_continue = 0;
+
+    if ((pw = getpwnam(VMD_USER)) == NULL)
+        err(1, "can't get _vmd user");
+    uid = pw->pw_uid;
+    gid = pw->pw_gid;
+
+    if (chroot(pw->pw_dir) == -1)
+        err(1, "chroot: %s", pw->pw_dir);
+
+    if (chdir("/") == -1)
+        err(1, "chdir(\"/\")");
+
+    if (setgroups(1, &gid) ||
+        setresgid(gid, gid, gid) ||
+        setresuid(uid, uid, uid))
+        err(1, "unable to revoke privs");
+
+    /*
+     * pledge in the vm processes:
+     * stdio - for malloc and basic I/O including events.
+     * vmm - for the vmm ioctls and operations.
+     */
+    if (pledge("stdio vmm", NULL) == -1)
+        err(errno, "pledge");
+
+    for (;;) {
+        vrp->vrp_irq = 0xFFFF;
+
+        if (ioctl(hvb->vmd_fd, VMM_IOC_RUN, vrp) < 0) {
+            err(1, "ukvm_hv_vcpu_loop: vm / vcpu run ioctl failed");
+        }
+
+        union vm_exit *vei = vrp->vrp_exit;
+        if (vrp->vrp_exit_reason != VM_EXIT_NONE) {
+            switch (vrp->vrp_exit_reason) {
+                case VMX_EXIT_IO:
+                case SVM_VMEXIT_IOIO:
+                    if (vei->vei.vei_dir != VEI_DIR_OUT
+                            || vei->vei.vei_size != 4)
+                        errx(1, "Invalid guest port access: port=0x%x", vei->vei.vei_port);
+                    if (vei->vei.vei_port < UKVM_HYPERCALL_PIO_BASE ||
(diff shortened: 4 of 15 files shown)


ukvm/ukvm_hv_openbsd.c
```diff
@@ -0,0 +1,130 @@
...
+
+struct ukvm_hv *ukvm_hv_init(size_t mem_size)
+{
+    struct ukvm_hv          *hv;
+    struct ukvm_hvb         *hvb;
+    struct vm_create_params *vcp;
+    struct vm_mem_range *vmr;
+    void *p;
+    
+	if(geteuid() != 0) {
```
ricarkol: misaligned

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+    };
+    return vsi;
+}
+
+void ukvm_hv_mem_size(size_t *mem_size) {
+    ukvm_x86_mem_size(mem_size);
+}
+
+void ukvm_hv_vcpu_init(struct ukvm_hv *hv, ukvm_gpa_t gpa_ep,
+        ukvm_gpa_t gpa_kend, char **cmdline)
```
ricarkol: misaligned

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+
+    if ((pw = getpwnam(VMD_USER)) == NULL)
+        err(1, "can't get _vmd user");
+    uid = pw->pw_uid;
+    gid = pw->pw_gid;
+
+    if (chroot(pw->pw_dir) == -1)
+        err(1, "chroot: %s", pw->pw_dir);
+
+    if (chdir("/") == -1)
```
ricarkol: Don't think this is needed (at least it's not in Linux).

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+        setresgid(gid, gid, gid) ||
+        setresuid(uid, uid, uid))
+        err(1, "unable to revoke privs");
+
+    /*
+     * pledge in the vm processes:
+     * stdio - for malloc and basic I/O including events.
+     * vmm - for the vmm ioctls and operations.
+     */
+    if (pledge("stdio vmm", NULL) == -1)
```
ricarkol: I don't get why this doesn't include the block device and the network socket.

ukvm/ukvm_hv_openbsd.c
```diff
@@ -0,0 +1,130 @@
...
+    hvb->vmd_fd = -1;
+
+    hvb->vmd_fd = open(VMM_NODE, O_RDWR);
+    if (hvb->vmd_fd == -1)
+        err(1, "VMM_NODE");
+
+    cleanup_hv = hv;
+    atexit(cleanup_vmd_fd);
+
+    vcp = calloc(1, sizeof (struct vm_create_params));
```
ricarkol: check for failed calloc

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+                case VMX_EXIT_HLT:
+                case SVM_VMEXIT_HLT:
+                    // stop on exit
+                    break;
+                case VMX_EXIT_TRIPLE_FAULT:
+                case SVM_VMEXIT_SHUTDOWN:
+                    /* reset VM */
+                    err(1, "Triple Fault");
+                default:
+                    err(1, "unknown exit reason");
```
ricarkol: should be errx

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+                    fn(hv, gpa);
+                    break;
+                case VMX_EXIT_HLT:
+                case SVM_VMEXIT_HLT:
+                    // stop on exit
+                    break;
+                case VMX_EXIT_TRIPLE_FAULT:
+                case SVM_VMEXIT_SHUTDOWN:
+                    /* reset VM */
+                    err(1, "Triple Fault");
```
ricarkol: should be errx

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+        err(errno, "pledge");
+
+    for (;;) {
+        vrp->vrp_irq = 0xFFFF;
+
+        if (ioctl(hvb->vmd_fd, VMM_IOC_RUN, vrp) < 0) {
+            if(errno == EIO) // guest halted with interrupts disabled, ie cli, hlt
+                return;
+
+            err(errno, "ukvm_hv_vcpu_loop: vm / vcpu run ioctl failed");
```
ricarkol: same err(1) comment as above

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+{
+    struct ukvm_hvb         *hvb = hv->b;
+    struct vm_run_params    *vrp;
+    struct passwd *pw;
+    uid_t uid;
+    gid_t gid;
+
+    vrp = malloc(sizeof(struct vm_run_params));
+    if (vrp == NULL)
+        err(1, "calloc vrp");
```
ricarkol: the call was a malloc (not calloc), did you use malloc instead of calloc by mistake (just checking that you are not expected the zeroed out memory)

tests/setup-tests.sh
```diff
@@ -42,6 +42,12 @@ FreeBSD)
     sysctl -w net.link.tap.up_on_open=1
     ifconfig tap100 create 10.0.0.1/24 link0 up
     ;;
+OpenBSD)
+     cd /dev
+     ./MAKEDEV tap100
```
ricarkol: Better to wrap this in `()`'s, so we come back to the current dir (just in case we add something at the end of the script that has to be at curr dir).

ukvm/ukvm_hv_openbsd.c
```diff
@@ -0,0 +1,130 @@
...
+    if (cleanup_hv != NULL && cleanup_hv->b->vmd_fd != -1)
+        close(cleanup_hv->b->vmd_fd);
+}
+
+static void cleanup_vm(void)
+{
+    if (cleanup_hv != NULL && cleanup_hv->b->vcp_id != -1) {
+        struct vm_terminate_params vtp = { .vtp_vm_id = cleanup_hv->b->vcp_id };
+        if (ioctl(cleanup_hv->b->vmd_fd, VMM_IOC_TERM, &vtp) < 0)
+            err(errno, "terminate vmm ioctl failed - still exiting");
```
ricarkol: I think our tests are checking for return code 1 in case of any error. Also, err() will print the errno. You could change this to just `err(1, "ukvm...")`

ukvm/ukvm_hv_openbsd.c
```diff
@@ -0,0 +1,130 @@
...
+        MAP_PRIVATE | MAP_ANON, -1, 0);
+    if (p == MAP_FAILED)
+        err(1, "mmap");
+
+    vmr->vmr_va = (vaddr_t)p;
+    hv->mem = p;
+    hv->mem_size = mem_size;
+
+    if (ioctl(hvb->vmd_fd, VMM_IOC_CREATE, vcp) < 0)
+        err(errno, "create vmm ioctl failed - exiting");
```
ricarkol: same err(1) comment as above

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: Its not needed, but it is much more secure chroot'ing ukvm.

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: The block device and network socket are opened earlier, so we already have a handle on them and are allowed access.

ukvm/ukvm_hv_openbsd.c
adamsteen: fixed, but github is now confused with the squashed commit.

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: excuse my ignorance but whats misaligned

adamsteen:
Thanks for the review. updated and squashed the commits, a few comments above.

github might have got confused with the squashing the commits

ukvm/ukvm_hv_openbsd_x86_64.c
ricarkol: sounds good

ukvm/ukvm_hv_openbsd_x86_64.c
ricarkol: ah, i see, thanks

configure.sh
```diff
@@ -124,6 +133,33 @@ case $(uname -s) in
...
         BUILD_MUEN="yes"
         ;;
+    OpenBSD)
+        # On OpenBSD/clang we use -nostdlibinc which gives us access to the
+        # clang-provided headers for compiler instrinsics. We copy the rest
+        # (std*.h, cdefs.h and their dependencies) from the host.
+        cc_is_clang || die "Only 'clang' is supported on OpenBSD"
+        [ "${TARGET_ARCH}" = "x86_64" ] ||
+            die "Only 'x86_64' is supported on OpenBSD"
+        ld_is_lld || die "Only 'ld.lld' is supported on OpenBSD"
```
mato: Are there other versions of `ld` (that would be in a standard system $PATH) on OpenBSD?

ukvm/ukvm_elf.c
```diff
@@ -128,7 +128,9 @@ void ukvm_elf_load(const char *file, uint8_t *mem, size_t mem_size,
             || hdr.e_ident[EI_MAG3] != ELFMAG3
             || hdr.e_ident[EI_CLASS] != ELFCLASS64
             || hdr.e_type != ET_EXEC
-#if defined(__x86_64__)
+#if defined(__OpenBSD__) && defined(__x86_64__)
+            || hdr.e_machine != EM_AMD64
```
mato: Presumably this is the same value as `EM_X86_64` but OpenBSD calls it `EM_AMD64`?

ukvm/ukvm_hv_openbsd.c
```diff
@@ -0,0 +1,133 @@
...
+
+    cleanup_hv = hv;
+    atexit(cleanup_vmd_fd);
+
+    vcp = calloc(1, sizeof (struct vm_create_params));
+    if (vcp == NULL)
+        err(1, "calloc");
+
+    vcp->vcp_ncpus = 1; // vmm only supports 1 cpu for now.
+    strlcpy(vcp->vcp_name, "ukvm", VMM_MAX_NAME_LEN);
```
mato: Will a single name of `ukvm` work for multiple instances? In any case you might want to do something similar to what the FreeBSD code does and get the `ukvm` PID in here.

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+#include <machine/specialreg.h>
+
+#include "ukvm.h"
+#include "ukvm_hv_openbsd.h"
+#include "ukvm_cpu_x86_64.h"
+
+uint64_t get_tsc_freq();
+static struct vcpu_segment_info sreg_to_vsi(const struct x86_sreg *);
+
+uint64_t get_tsc_freq()
```
mato: Should be `static`, also please use `void` in the parameter list for consistency with other places. You probably don't need the separate prototype on L49.

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+
+    vrp->vrp_exit = malloc(sizeof(union vm_exit));
+    if (vrp == NULL)
+        err(1, "malloc vrp_exit");
+
+    vrp->vrp_vm_id = hvb->vcp_id;
+    vrp->vrp_vcpu_id = hvb->vcpu_id;
+    vrp->vrp_continue = 0;
+
+    if ((pw = getpwnam(VMD_USER)) == NULL)
```
mato: Is the user on _any_ OpenBSD system always going to be `_vmd`? I take it access to `vmm` APIs requires root?

ukvm/ukvm_hv_openbsd_x86_64.c
mato: With the Linux/FreeBSD ports we've been leaving any extra security (chroot, jails, containers, whatever) up to the user for now. If its the OpenBSD way to always do this in every tool then this is fine. (We can always change/refactor it later if we end up doing more of this ourselves also on other systems)

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+     * vmm - for the vmm ioctls and operations.
+     */
+    if (pledge("stdio vmm", NULL) == -1)
+        err(errno, "pledge");
+
+    for (;;) {
+        vrp->vrp_irq = 0xFFFF;
+
+        if (ioctl(hvb->vmd_fd, VMM_IOC_RUN, vrp) < 0) {
+            if(errno == EIO) // guest halted with interrupts disabled, ie cli, hlt
```
mato: Style: Please use `/* ... */` on the line above, as in other places.

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+            switch (vrp->vrp_exit_reason) {
+                case VMX_EXIT_INT_WINDOW:
+                case VMX_EXIT_CPUID:
+                case VMX_EXIT_EXTINT:
+                case SVM_VMEXIT_INTR:
+                case VMX_EXIT_EPT_VIOLATION:
+                case SVM_VMEXIT_NPF:
+                case SVM_VMEXIT_MSR:
+                case SVM_VMEXIT_CPUID:
+                    // nothing to be done here, as per vmd
```
mato: Is there _really_ nothing to be done here? Do we get these exits in normal operation? If not, they should be unhandled exits (same as unknown exit).

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+                    ukvm_hypercall_fn_t fn = ukvm_core_hypercalls[nr];
+                    if (fn == NULL)
+                        errx(1, "Invalid guest hypercall: num=%d", nr);
+
+                    ukvm_gpa_t gpa = vei->vei.vei_data;
+                    fn(hv, gpa);
+                    break;
+                case VMX_EXIT_HLT:
+                case SVM_VMEXIT_HLT:
+                    // stop on exit
```
mato: Is this the same as the `EIO` case further up? What's the difference?

ukvm/ukvm_hv_openbsd_x86_64.c
```diff
@@ -0,0 +1,254 @@
...
+                    break;
+                case VMX_EXIT_TRIPLE_FAULT:
+                case SVM_VMEXIT_SHUTDOWN:
+                    /* reset VM */
+                    errx(1, "Triple Fault");
+                default:
+                    errx(1, "unknown exit reason");
```
mato: Use something like the code from FreeBSD so that these can be debugged, otherwise the operator has no idea what happened:
```
default: {
            errx(1, "unhandled exit: exitcode=%d, rip=0x%" PRIx64,
                    vme->exitcode, vme->rip);
        }
```

configure.sh
```diff
@@ -124,6 +133,33 @@ case $(uname -s) in
...
+
+        mkdir -p ${HOST_INCDIR}
+        mkdir -p ${HOST_INCDIR}/machine ${HOST_INCDIR}/sys ${HOST_INCDIR}/amd64
+        for f in ${SRCS_MACH}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/machine; done
+        for f in ${SRCS_SYS}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/sys; done
+        for f in ${SRCS_AMD64}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/amd64; done
+        for f in ${SRCS}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}; done
+
+        HOST_CFLAGS="-fno-pie -fno-stack-protector -nostdlibinc"
+        HOST_LDFLAGS="-nopie"
```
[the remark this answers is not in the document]
adamsteen: yes, security mitigations are enabled by default and hard to turn off.

configure.sh
adamsteen: OpenBSD's default ld is GNU ld version 2.17. which is just too old for the Solo5's linker scripts. OpenBSD's amd64 port is compiled with clang but linked with ld, but always has ld.lld available.

ukvm/ukvm_elf.c
adamsteen: just a quick look on the net, both are 62.

[OpenBSD definition](https://github.com/openbsd/src/blob/master/sys/sys/exec_elf.h#L196)
[Linux definition](https://github.com/torvalds/linux/blob/ead751507de86d90fa250431e9990a8b881f713c/include/uapi/linux/elf-em.h#L31)

ukvm/ukvm_hv_openbsd.c
adamsteen: yes, but i just had a look at the FreeBSD code i think i will implement something like that.

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: yes, _vmd is the user setup for [vmd](https://man.openbsd.org/vmd) and access to vmm requires root. I tried to follow vmd (OpenBSD's userland vm daemon)

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: I was following what vmd does [vcpu_exit#L1471-L1479](https://github.com/openbsd/src/blob/7e7a8c998653110b57ed8bc9fdb7cb231e91d0a8/usr.sbin/vmd/vm.c#L1471-L1479)

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: again following vmd [vcpu_exit#L1485-L1499](https://github.com/openbsd/src/blob/7e7a8c998653110b57ed8bc9fdb7cb231e91d0a8/usr.sbin/vmd/vm.c#L1485-L1499)

i might need to look further into these two, and ask some questions.

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: It is the OpenBSD way to do this with most, if not all tools.

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: base on the comment (L1472-L1478)
```
We may be exiting to vmd to handle a pending interrupt but
at the same time the last exit type may have been one of
these. In this case, there's nothing extra to be done
here (and falling through to the default case below results
in more vmd log spam).
```
these are normal operation. Would you like me to chase it down further with mlarkin@openbsd?

ukvm/ukvm_hv_openbsd_x86_64.c
adamsteen: base on [vmm.c : vmx_handle_halt](https://github.com/openbsd/src/blob/master/sys/arch/amd64/amd64/vmm.c#L4237-L4283) first the return value comments and second line [L4275-L4282](https://github.com/openbsd/src/blob/master/sys/arch/amd64/amd64/vmm.c#L4275-L4282) The EIO case is for when the guest halts with interrupts disabed, (the case for solo5) and the this case is when vmm has receive a VMX_EXIT_HLT exit reason but the guest has not halted with interrupts disabled, but i am not sure if this will ever occur.

based on this, we could remove these lines and if an occurrence pops up we can always add them back.

ukvm/ukvm_hv_openbsd_x86_64.c
mato: Given that we don't use interrupts for anything, based on that comment I'd be inclined to remove these (turn them into an unhandled exit). It's easy enough to try, e.g. see if test_ping_serve runs fine with the change.

ukvm/ukvm_hv_openbsd_x86_64.c
mato: Ok, fine. If someone wants different behaviour we can change it later.

configure.sh
mato: What do you mean by "always has ld.lld available"? Does this require manually using `LD=ld.lld` in various places to make the build work?

ukvm/ukvm_hv_openbsd_x86_64.c
mato: Yes, let's do that.

configure.sh
adamsteen: yes, before working on solo5 and MirageOS i run `export LD=ld.lld` in my shell. and everything seems to build correctly.

ukvm/ukvm_elf.c
adamsteen: When i first wrote the code you presumption was correct, i went back and checked when i had more time and it does defined both, so i switched back to `EM_X86_64`

mato: Thanks!

configure.sh
mato: Hmm. We should have a way to make this work w/o relying on that step. I'll go look back at the dance we did ages ago when we were still depending on `/usr/local/bin/ld` for the FreeBSD port and get back to you.

adamsteen: where you thinking something along the lines of [encode ld.lld as a default LD, similar to FreeBSD a while ago](https://github.com/adamsteen/solo5/commit/c69f83462c3023022ff525ba796cb5e3f46c5831) like what was removed in [remove LD hacks](https://github.com/Solo5/solo5/commit/a39d7f69e9044a6bd90571fe1d6391a347591e46#diff-735003ab6f5797c8b8bdecbbbb9322a8)?

mato: Yeah, that's it. Unfortunately that requires putting it back in downstream (Mirage) as well. Could you find out if there are any plans (e.g. by the next OpenBSD release) to make `ld.lld` the default linker? If that were the case then I'd rather just wait until that happens.

adamsteen: I doubt ld.lld will be made the default linker any time soon, due to [Bug 30815 - linking OpenBSD/amd64 kernel](https://bugs.llvm.org/show_bug.cgi?id=30815) (The linked kernel doesn't boot.) but i have sent an [email](https://marc.info/?l=openbsd-misc&m=151330492417254&w=2) any ways to see what the projects plans are.

hannesm:
about `LD`: I don't remember whether we discussed the details why the OpenBSD default ld won't work with solo5.  I can see two solutions here:
- re-add `LD` to `pkg-config` and callout to this (as we used to do for some time to use a newer ld on FreeBSD)
- get OpenBSDs ld to work with solo5

I clearly prefer the latter option, but if it is not possible, I can dig and try to find the respective patches for mirage etc.  please tell if the second option is not viable.

kernel/malloc.c
```diff
@@ -70,7 +70,7 @@ void *solo5_calloc(size_t, size_t) __attribute__ ((alias ("calloc")));
 void *solo5_realloc(void *, size_t) __attribute__ ((alias ("realloc")));
 
 /* disable null-pointer-arithmetic warning on clang */
-#if defined(__clang__)
+#if defined(__clang__) && !defined(__OpenBSD__) 
 #pragma clang diagnostic ignored "-Wunknown-pragmas"
```
hannesm: for older (3.8, 4.0) clang on FreeBSD, ignoring the `unknown-pragmas` -- as done here, was sufficient for me.  doesn't this work on OpenBSD?

adamsteen:
Just bringing this upto date.

OpenBSD errors with "malloc.c:75:34: error: unknown warning group '-Wnull-pointer-arithmetic', ignored [-Werror,-Wunknown-warning-option]"

I don't think clang 5 supports "-Wnull-pointer-arithmetic" [Clang 5 Documentation - Diagnostic flags in Clang](http://releases.llvm.org/5.0.0/tools/clang/docs/DiagnosticsReference.html)

thus `#if defined(__clang__) &&  __clang_major > 5`

I do like the exit status change it actually cleans up the code a little bit.

hannesm:
@adamsteen I opened #241 which hopefully works with your clang5 smoothly.  the `null-pointer-arithmetic` warning got introduced in clang6, thus on any platform using an earlier clang it will error (disabling `unknown-pragma` via a pragma worked on FreeBSD with earlier clang, but since this does not work for you, I propose #241).

kernel/malloc.c
adamsteen: @hannesm before you commented i added the change
 `#if defined(__clang__)  &&  __clang_major > 5`
which seems better as it what the old code had for earlier version of clang
`#if defined(__clang__) && !defined(__OpenBSD__) `

hannesm: yes, but it needs to be `__clang_major__` (i.e. you missed the `__` at the end) AFAICT -- would be great if you can try #241 instead of the `!defined(__OpenBSD__)` :)

adamsteen: yes #241 works perfectly. and merged into my branch.

hannesm:
dear @adamsteen, I read your toot (at https://bsd.network/@adams/99750047461182320), error (of ocaml-freestanding, sorry for using the wrong repository, but there's no issue on ocaml-freestanding) at https://gist.github.com/adamsteen/151017e95768b404f5ec0ad73453d025

excerpt from your posed log (ignoring all the warnings during compilation of openlibm):
```
In file included from major_gc.c:21:
In file included from ../byterun/caml/compact.h:21:
../byterun/caml/config.h:79:25: error: typedef redefinition with different types ('long' vs '__int64_t' (aka 'long long'))
typedef ARCH_INT64_TYPE int64_t;
                        ^
/home/asteen/.opam/4.06.1/lib/pkgconfig/../../include/solo5-kernel-ukvm/include/host/stdint.h:62:20: note: previous definition is here
typedef __int64_t               int64_t;
                                ^
In file included from major_gc.c:21:
In file included from ../byterun/caml/compact.h:21:
../byterun/caml/config.h:80:26: error: typedef redefinition with different types ('unsigned long' vs '__uint64_t' (aka 'unsigned long long'))
typedef ARCH_UINT64_TYPE uint64_t;
                         ^
/home/asteen/.opam/4.06.1/lib/pkgconfig/../../include/solo5-kernel-ukvm/include/host/stdint.h:67:21: note: previous definition is here
typedef __uint64_t              uint64_t;
                                ^
2 errors generated.
gmake[1]: *** [Makefile:150: major_gc.o] Error 1
```

so, the solo5 build system copies your `sys/stdint.h` around, and ocaml-freestanding (in confg.in/m.x86_64.h) defines the following:
```
#define ARCH_INT64_TYPE long
#define ARCH_UINT64_TYPE unsigned long
```

I suspect these don't work on OpenBSD, can you try to redefine them in m.x86_64.h (to `long long` resp. `unsigned long long`)?

adamsteen:
@hannesm The tweak to ARCH_INT64_TYPE  and ARCH_UINT64_TYPE has done the trick

```
#define ARCH_INT64_TYPE long long
#define ARCH_UINT64_TYPE unsigned long long
```

I now have trouble with depext, but should be able to sort that out and hopefully i will have something working soon

hannesm:
@adamsteen great! what is the status of `ld`?  do we need binutils from ports on OpenBSD, or does the default `ld` (or clang's `ld.lld`) work fine?  (there's some discussion about this above, and you seem to be using `ld.lld` successfully!?)

adamsteen:
@hannesm I have been using ld.lld successfully so far and all the solo5 testing pass, Ihave a few minutes tonight too look at things, i hope to have something (even hackish) working end to end.

adamsteen:
@hannesm I think i attributed the problem to the wrong thing, The problem is an inconsistency between compiling and linking with PIC and PIE flags.

It is my understanding that the following means some of the libs have been compiled without `no-pie` as the default on OpenBSD is `-fpic`.
```
ocamlfind ocamlmklib -o lib/mirage-solo5_bindings lib/bindings/alloc_pages_stubs.o lib/bindings/atomic_stubs.o lib/bindings/barrier_
    stubs.o lib/bindings/checksum_stubs.o lib/bindings/clock_stubs.o lib/bindings/cstruct_stubs.o lib/bindings/main.o lib/bindings/mm_st
    ubs.o lib/bindings/solo5_block_stubs.o lib/bindings/solo5_console_stubs.o lib/bindings/solo5_net_stubs.o
 16 + ocamlfind ocamlmklib -o lib/mirage-solo5_bindings lib/bindings/alloc_pages_stubs.o lib/bindings/atomic_stubs.o lib/bindings/barrie
    r_stubs.o lib/bindings/checksum_stubs.o lib/bindings/clock_stubs.o lib/bindings/cstruct_stubs.o lib/bindings/main.o lib/bindings/mm_
    stubs.o lib/bindings/solo5_block_stubs.o lib/bindings/solo5_console_stubs.o lib/bindings/solo5_net_stubs.o
```
`/usr/bin/ld: error: can't create dynamic relocation R_X86_64_PC32 against symbol: caml_local_roots in readonly segment
`

I did check `solo5-kernel-ukvm.pc` and everything looks correct but the `$LD` was not picked up correctly.
```
prefix=${pcfiledir}/../..
exec_prefix=${prefix}
includedir=${prefix}/include/solo5-kernel-ukvm/include
libdir=${exec_prefix}/lib/solo5-kernel-ukvm
ld=!LD!
ldflags=-nostdlib -z max-page-size=0x1000 -static -nopie -T ${libdir}/solo5.lds ${libdir}/solo5.o

Name: solo5-kernel-ukvm
Version: 0.1
Description: Solo5 unikernel base (ukvm target)
Cflags: -fno-pie -fno-stack-protector -nostdlibinc -ffreestanding -mno-red-zone -isystem ${includedir}/host -I${includedir}
```

adamsteen:
@mato

Hi Mato
it looks like i have hit the problem with no-pie in mirage-solo5 as specified in configure.sh

```
# Recent distributions now default to PIE enabled. Disable it explicitly
# if that's the case here.
# XXX: This breaks MirageOS in (at least) the build of mirage-solo5 due
# to -fno-pie breaking the build of lib/dllmirage-solo5_bindings.so.
# Keep this disabled until that is resolved.
```

[Port to topkg](https://github.com/mirage/mirage-solo5/pull/15) point 3
[kernel: -fno-pie breaks mirage-solo5](https://github.com/Solo5/solo5/commit/e8b44b5d6bcf4eaff2f386e2adb04ff0ffbc739a#diff-735003ab6f5797c8b8bdecbbbb9322a8)

any idea how i can work around this easily?

Cheers
Adam[

adamsteen:
Hi @hannesm 

I think you last fix got a me long way!

```
asteen:/home/asteen/devl/mirage-skeleton/tutorial/hello:$ uname -a
OpenBSD x220.adamsteen.com.au 6.3 GENERIC.MP#129 amd64
asteen:/home/asteen/devl/mirage-skeleton/tutorial/hello:$ doas ukvm-bin hello.ukvm
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5:       text @ (0x100000 - 0x21dfff)
Solo5:     rodata @ (0x21e000 - 0x254fff)
Solo5:       data @ (0x255000 - 0x4ffff)
Solo5:       heap >= 0x50000 < stack < 0x20000000
Solo5: trap: type=#UD ec=0x0 rip=0x1008 rsp=0x1ffffdf8 rflags=0x10002
Solo5: ABORT: cpu_x86_64.c:171: Fatal trap
```

I don't have time to look into this futher! but we are getting closer and closer! Thank you for you help!

adamsteen:
Looks like the Trap is occuring somewhere inside ocaml, I add some log statements, and found it fails in the call `caml_startup`. I am not sure where to look next?

adamsteen:
Hi @mato @hannesm @ricarkol @ricarkol @avsm 

I finally have MirageOS on OpenBSD working.

```
asteen:/home/asteen/devl/mirage-skeleton/tutorial/hello:2238$ uname -a
OpenBSD x220.adamsteen.com.au 6.3 GENERIC.MP#0 amd64
asteen:/home/asteen/devl/mirage-skeleton/tutorial/hello:2239$ doas ukvm-bin hello.ukvm
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5:       text @ (0x100000 - 0x21efff)
Solo5:     rodata @ (0x21f000 - 0x24efff)
Solo5:       data @ (0x24f000 - 0x30cfff)
Solo5:       heap >= 0x30d000 < stack < 0x20000000
2018-04-08 06:28:01 -00:00: INF [application] hello
2018-04-08 06:28:02 -00:00: INF [application] hello
2018-04-08 06:28:03 -00:00: INF [application] hello
2018-04-08 06:28:04 -00:00: INF [application] hello
Solo5: solo5_exit(0) called
asteen:/home/asteen/devl/mirage-skeleton/tutorial/hello:2240$
```

using the following forks/PRs
[opam-depext  'update OpenBSD Support' PR95](https://github.com/ocaml/opam-depext/pull/95)
this PR 'Porting ukvm to OpenBSD vmm APIs' #226 
[ocam-freestanding 'Porting ukvm to OpenBSD vmm APIs' PR35](https://github.com/mirage/ocaml-freestanding/pull/35)
[mirage-solo5 'Porting ukvm to OpenBSD vmm APIs' PR30](https://github.com/mirage/mirage-solo5/pull/30)
[mirage 'respect ld from pkg-config' PR898](https://github.com/mirage/mirage/pull/898)

the one issue i am not sure the best way to solve, i have to symbolically link pkfconf to pkg-config early in my path

`ln -s /usr/local/bin/pkgconf $HOME/.local/bin/pkg-config`

i have tested all the tututorials, device-usage and applications
- device-usage/conduit_server fails in gmp-freestanding
- device-usage/http-fech fails in gmp-freestanding
- device-usage/prng fails in gmp-freestanding
- applicatons/static_website_tls  fails in gmp-freestanding

it appears the above failure was just an instability in current.

Cheers
Adam

mato:
@adamsteen:

> I finally have MirageOS on OpenBSD working.

Thank you very much for your work and perseverance! I'm getting through reviewing some of the dependent PRs, and will get to this over the next few days.

configure.sh
```diff
@@ -137,7 +137,10 @@ case $(uname -s) in
         cc_is_clang || die "Only 'clang' is supported on OpenBSD"
         [ "${TARGET_ARCH}" = "x86_64" ] ||
             die "Only 'x86_64' is supported on OpenBSD"
-        ld_is_lld || die "Only 'ld.lld' is supported on OpenBSD"
+        if [ !ld_is_lld ]; then
+            LD='/usr/bin/ld.lld'
+            echo "WARNING: using 'ld.lld' as 'ld' is not supported on OpenBSD"
```
mato: This should include a check that `/usr/bin/ld.lld` exists and is executable, and `die` if not. This is so that users on OpenBSD systems without `ld.lld` are not confused when things start failing mysteriously further down the chain.

adamsteen: I am very sure any OpenBSD 6.2 or later with have ld.lld, as it is part of base, but there is no harm in having the check.

I have added this to configure.sh.

configure.sh
```diff
@@ -124,6 +130,37 @@ case $(uname -s) in
...
         BUILD_MUEN="yes"
         ;;
+    OpenBSD)
+        # On OpenBSD/clang we use -nostdlibinc which gives us access to the
+        # clang-provided headers for compiler instrinsics. We copy the rest
+        # (std*.h, cdefs.h and their dependencies) from the host.
+        cc_is_clang || die "Only 'clang' is supported on OpenBSD"
+        [ "${TARGET_ARCH}" = "x86_64" ] ||
+            die "Only 'x86_64' is supported on OpenBSD"
+        if [ !ld_is_lld ]; then
```
mato: This is wrong, it should probably read `if ! ld_is_lld; then` -- you can't use `!` in this way with `[`.

Also, I don't understand the following warning message, did you mean "using GNU 'ld' is not supported on OpenBSD, falling back to 'ld.lld'" or something like that?

mato:
@adamsteen I think this is almost ready to merge (pending last comment about the bogus ld.lld test). I've merged most of the downstream PRs with the exception of mirage/ocaml-freestanding#35 which needs an update.

Before I do that, could you please rebase onto latest master, squashing your commits in the process? In case you're not familiar with the process, it goes like this:

    $ git branch temp # To give you a staging branch in case things go wrong
    $ git checkout temp
    $ git rebase -i master
    # (Follow the instructions)
    # Check "git log", and verify everything still builds...
    # You can then shortcut and force-push your staging branch (if you used one to this PR, or just repeat the steps on the original branch. Shortcut is:
    $ git push --force origin openbsd

I'll then click the big green button ASAP assuming all goes well!

adamsteen:
@mato all fixed, tested and squashed!

Any ideas i how it should look at handling?

> the one issue i am not sure the best way to solve, i have to symbolically link pkfconf to pkg-config early in my path
> 
> `ln -s /usr/local/bin/pkgconf $HOME/.local/bin/pkg-config`

mato:
@adamsteen Sorry, I missed that. On my OpenBSD 6.3 VM I have a `pkg-config` in `/usr/bin` by default, can you check if this is a new thing that has changed in recent versions? If so, then we might be able to use that one rather than depend on the `pkgconf` port. I'll investigate some more in a bit.

adamsteen:
@mato 
For me the `pkg-config` in`/usr/bin` doesn't or didn't work correctly. I think the path returned to find things was just empty. see [Issue 206 # comment](https://github.com/Solo5/solo5/issues/206#issuecomment-339835821)

Have you tested things (build or run) yet?

mato:
@adamsteen I've not tested more than Solo5 itself so far; going to do that now and will report back.

Regarding the `WARNING`, I think we can kill it.

mato:
@adamsteen FTR, discussion from IRC:

Results of today's testing on OpenBSD 6.3 are that the ukvm end to end tests all pass, however attempting to run the `tutorial/hello` Mirage unikernel from `mirage-skeleton` gives me this:

````
12-00-00-ef-01-02$ doas ./ukvm-bin ./hello.ukvm 
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5:       text @ (0x100000 - 0x21dfff)
Solo5:     rodata @ (0x21e000 - 0x254fff)
Solo5:       data @ (0x255000 - 0x4ffff)
Solo5:       heap >= 0x50000 < stack < 0x20000000
Solo5: trap: type=#UD ec=0x0 rip=0x1008 rsp=0x1ffffdf8 rflags=0x10002
Solo5: ABORT: cpu_x86_64.c:171: Fatal trap

````

Note the obviously bogus values for "end of data segment" (0x4ffff) and thus also start of heap. I suspect the #UD is coming from a simple memory overwrite, but have not had time to investigate yet where/why the bogus values are coming from, though I suspect something in the Mirage-specific bits of the build process rather than at the Solo5/ukvm layer.

mato:
> For me the pkg-config in/usr/bin doesn't or didn't work correctly. I think the path returned to find things was just empty. see Issue 206 # comment

I've tracked this down to the OpenBSD reimplementation of pkgconfig not supporting `${pcfiledir}`. There are several ways to fix this, none of them particularly clean at this point in time, so I think we should just document the need to manually run `ln -s /usr/local/bin/pkgconf $(opam config var bin)/pkg-config` on OpenBSD.

Alternatively, maybe the `solo5-kernel-ukvm` `opam-ukvm-install` target could do this automatically for OpenBSD as a stopgap. /cc @hannesm What do you think? Is this too bad of a hack?

hannesm:
@mato if that works out reliable, I guess we can do that.  what does OpenBSD's `pkgconf` do when it encounters `${pcfiledir}`?

mato:
> what does OpenBSD's pkgconf do when it encounters ${pcfiledir}?

@hannsem Nothing, i.e. replace it with an empty value, so the various paths are all messed up. We could also modify _all_ Mirage packages using `pkg-config` to set `prefix` using an absolute path based on `$(opam config var  prefix)` but that would be much more work and more invasive.

A clean fix would be to drop `pkg-config` entirely and use per-package variables to pass the values around, but that would make us depend on OPAM 2 which I don't think is doable today.

hannesm:
ok, this means when the symlink is installed on OpenBSD, `pkg-config` just works since it's in a similar location (with right nesting level) of `opam prefix`!?

mato:
> ok, this means when the symlink is installed on OpenBSD, pkg-config just works since it's in a similar location (with right nesting level) of opam prefix!?

No, what that symlink does is interpose `/usr/local/bin/pkgconf` (the pkg-config from the `pkgconf` port, which is the GNU version that supports `${pcfiledir}`) as `pkg-config` on the user's `$PATH` (since `$(opam config var bin)` is included in `$PATH` by opam), thus making our `.pc` files work correctly.

However, I'm not 100% convinced that this is a good idea, at least not without printing a big fat warning, since it "silently" changes which `pkg-config` gets used by the user's shell.

adamsteen:
@mato 

I have setup a OpenBSD 6.3 and reproduced the error `Solo5: ABORT: cpu_x86_64.c:171: Fatal trap` with both OPAM 1.2.2 and OPAM 2.0,0rc1.

I also tested
- hello.ukvm produced on Current on Current - Succeeded
- hello.ukvm produced on Current on 6.3 - Succeeded
- hello.ukvm produced on 6.3 on 6.3 - Failed
- hello.ukvm produced on 6.3 on Current - Failed

The biggest difference between 6.3 and Current is 6.3 has clang/llvm 5 and Current has clang/llvm 6.

adamsteen:
@hannesm 

- I think that might try and tackle extending the OpenBSDs perl script to support pcfiledir. Then i can ditch the need for the extra port.
- The works for me. I have had little time the last few weeks and i am off the grid for a next week or so. If the problem with OpenBSD and i can't find a workaround, the fix won't be backported, it just not OpenBSD's policy.
- Excellent, i will follow the fix, when i get some time I will give it a try too.

I am off the grid for the next week or so, i will check back in when i get back

mato:
@hannesm @adamsteen I agree with all your conclusions. I'll try and test this myself on -current this week just to be sure things work as expected, and then probably merge this (+ a fix to `conf-pkg-config`) later in the week so that I can continue some core work without having to worry about integrating the OpenBSD support.

adamsteen:
@mato 

Is there anything else you need from me? I am following Solo5 and a few other Mirage repos. I am focusing on OpenBSD's pkg-config's support of pcfiledir, and small MirageOS projects, so i won't be very active on GitHub for a bit.

Please don't hesitate to let me know if you need me to look at something OpenBSD related.

Cheers
Adam

mato:
@adamsteen I think this is good to go, am mainly waiting for a resolution on ocaml/opam#12018 to get a working `conf-pkg-config`.

I do have two small requests:

1. Could you please amend the Git commit message with a short summary of the changes done / new files added / anything else you think is important and re-push the branch? The information in the PR can get lost/forgotten about, but the commit message will stay in the history forever.

2. Also, could you add yourself to the `AUTHORS` file? Sorted lexicographically.

Thanks!

mato:
@adamsteen I had to expedite the merge of #246 since the changes for #258 were blowing my mind. This is causing the conflicts you're seeing, fixing them should be trivial, just add the relevant OpenBSD clause to `test.bats`.

The FreeBSD CI failure was transient and is fixed now. Unfortunately there's no way to restart jobs, please make a trivial change and re-push.

Once all the checks are green this is good to merge.

adamsteen:
@mato
I fixed the conflicts earlier(I thought) will double check in the morning and re-push a change!

mato:
@adamsteen Ah, you have indeed. I'll just merge this then, no point in delaying.

Issue #213, opened 2017-07-31
Update the GDB usage in the README


ricarkol:

mato:
I think this is "solved" by #227, and we will need to rewrite all our docs post-refactoring anyway, so closing for now.

PR #227, opened 2017-12-12
Add quickstart

ricarkol:
Adding a quickstart, and updating the gdb section of the README.

Also, adding the gdb module to tests/test_hello so gdb debugging is easier to play with. It will also be used for the gdb tests to be added after.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,16 @@
 # About Solo5
 
 Solo5 is most useful as a "base layer" to run
-[MirageOS](https://mirage.io/) unikernels, either on various existing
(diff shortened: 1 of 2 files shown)


djwillia:
I know you want to close https://github.com/Solo5/solo5/issues/213 and I think we may as well merge this for now.

But the README is really out of date in other ways too (e.g., naming, some of the other backends, etc.)  Perhaps we could have a new issue that collects some of those README todos all in the same place?

mato:
This is fine so I'll just merge it in the mean time. README basically needs to be completely restructured and rewritten, I'm happy to do that and have a clear idea of how it should work, but I need to do #201 first and need a couple of days of uninterrupted concentration to do it :-/

PR #228, opened 2018-01-03
Generating core file for solo5/ukvm

niks3089:
Following is the PR on changes done to generate a core dump for the unikernel running on solo5 with ukvm 

**Changes done:**

When the Unikernel application encounters an error, a core dump named "core.unikernel" is created when the application is run with ukvm.  ("unikernel" is appended to the core file name because if the ukvm crashes when creating the core dump, we might lose the ukvm core dump info)
When an error is encountered we call platform_dump_core() is called which is triggers the hypercall UKVM_HYPERCALL_DUMP_CORE for ukvm platform and then calls platform_exit() and for other platforms, platform_dump_core() simply calls platform_exit()
platform_dump_core() accepts a pointer to the data and the length of the data which is passed in the hypercall to ukvm. When the CPU trap happens, the register information is passed and if the application voluntarily asserts, the no data is required.

Currently only supported on x86

**New solo5 APIs:**

solo5_dump_core(): API which allows an application to generate a core dump when it needs to. 
solo5_assert(): Helper macro which wraps around solo5_dump_core()  that lets the applications use as an assert and if supported can generate a core dump.

**Testing done:**
Added test_assert in the tests directory to test the solo5_assert API

**Sample output**

```
 ./ukvm-bin test_exception.ukvm
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5:       text @ (0x100000 - 0x106fff)
Solo5:     rodata @ (0x107000 - 0x108fff)
Solo5:       data @ (0x109000 - 0x10dfff)
Solo5:       heap >= 0x10e000 < stack < 0x20000000

**** Solo5 standalone test_exception ****

Solo5: trap: type=#PF ec=0x2 rip=0x106cb5 rsp=0x1fffffd0 rflags=0x10002
Solo5: trap: cr2=0x400000000
Solo5: ABORT & DUMPING CORE: cpu_x86_64.c:171: Fatal trap


osboxes:~/projects/solo5/tests/test_exception]$ ls -ltr
total 1049816
-rw------- 1 osboxes osboxes 536871444 Dec 22 03:42 core.unikernel

!gdb
gdb ./test_exception.ukvm core.unikernel
Reading symbols from ./test_exception.ukvm...done.
[New LWP 123546]
#0  0x0000000000106cb5 in solo5_app_main (cmdline=cmdline@entry=0x38 "") at test_exception.c:34
34          *(uint8_t *)addr_invalid = 1;
(gdb) bt
#0  0x0000000000106cb5 in solo5_app_main (cmdline=cmdline@entry=0x38 "") at test_exception.c:34
#1  0x0000000000100089 in _start (arg=0x10be98 <cpu_trap_stack+3640>) at ukvm/kernel.c:42

```
**Additional notes**
I was trying to pass register info between solo5 kernel and ukvm and I was doing a malloc in kernel/ukvm/platform.c after CPU trap, however, malloc is failing and @ricarkol debugged this and the issue is that the exception handler uses a different stack which happens to be below the heap, and therefore the sbrk check fails and he suggested we don't have to remove the check instead we use the stack where we specify the maximum bytes for passing the resister info


hannesm:
why do the surf-builds not pick this PR up? at least the FreeBSD ones are running, but I don't see any output on the console...

niks3089:
Yes, even I wondering the same. There were few build errors when CI was run so I fixed it after which I don't see anymore checks being run.

ricarkol:
Hey @niks3089 , I was reviewing this and was thinking that the unikernel side was doing too much. It should only be telling ukvm-bin that it failed and ukvm-bin should take care of everything else, like deciding when to dump a core.

The other change, that you can start working on, is to move everything ELF specific to ukvm_elf.c. For example, creating the `notes` looks very ELF specific.

niks3089:
Ok @ricarkol. I went over your pull request. I agree it's going to be easier for if the ukvm-bin does most of the work. I'll wait for your PR to get merged and will re-write to 

Also, what do you think about the idea of exposing an API in solo5.h which allows application to assert and dump the core when *it* wants to ? The application can wrap solo5_assert() around assert(). 

Regarding your point on moving elf specific to ukvm_elf.c, all the generic stuff are in ukvm_elf.c, the functions ukvm_hv_dump_notes() and ukvm_hv_get_notes_size() is platform specific. Still, let me try and move some generic bits of it to ukvm_elf.c

hannesm:
> Also, what do you think about the idea of exposing an API in solo5.h which allows application to assert and dump the core when it wants to ? The application can wrap solo5_assert() around assert().

imho `ukvm-bin` (e.g. by a command-line argument) should be responsible for dumping cores.  once the unikernel exits (with a non-0 status code, as proposed in #230), `ukvm-bin` should dump core.  what would be a use case for the unikernel to decide when to dump a core?

ricarkol:
Hi @niks3089 , as @hannesm is mentioning, I can't really think of any use case for having the user dump a core and keep going. Btw, failed asserts should be exiting with error with the latest #230 .

mato:
Hi @niks3089. Thank you for your contribution. You have picked a useful feature, but the design and safe implementation of this is quite challenging.

Starting with some design (meta-)points:

1. Conceptually, as others have already pointed out, it should be the monitor's (`ukvm-bin`) job to decide whether or not and when to dump core. Anything the unikernel has to say about this should be considered a *hint* only, and any information passed from the unikernel to the monitor (or fished
out of the unikernel's memory) should be considered untrusted and treated with extreme prejudice.
2. The Solo5 APIs and `ukvm` hypercall changes to support this functionality should be as minimal as possible, if any.
3. Likewise, the associated `ukvm` changes should be minimal, and (at least initially) implemented as an _optional_ module.

Thinking about the design and possible implementation in the context of the above points, I would suggest re-working this as follows:

1. Put your code into a module, e.g. `ukvm_dumpcore.c`, activated with `--dumpcore`. This includes the ELF core dumping code and associated structures, and the `ukvm_hv_get_regs()`, for now. Use the appropriate `#ifdef` logic to pick the correct implementation and start with just the arch/host combination that is easiest for you to test (presumably x86_64/Linux).
2. Add a `void solo5_abort(void)` call to the Solo5 public API, and a private `void platform_abort(struct trap_regs *regs)` function that contains the platform-specific implementation. `solo5_abort()` should just call through to `platform_abort(NULL)`.
  a. On non-ukvm platforms `platform_abort()` can just call `platform_exit()` for now.
  b. On ukvm, in your dumpcore module, register an `UKVM_ABORT` hypercall, using the following definition:
````
struct ukvm_abort {
    /* IN */
    UKVM_GUEST_PTR(void *) regs;
};
````
3. The `UKVM_ABORT` hypercall will do the actual work of dumping the core file. It should: 
  a. Retrieve the guest register state from the hypervisor.
  b. _If and only if_ `abort->regs` is non-NULL _and_ points to a valid `gpa`, it should fish the "real" register values out of guest memory, overriding any it retrieved from the hypervisor in the previous step.

I don't want to spoil all your fun and do your work for you, so I'm leaving the details of how exactly point (3)(b) above should work as an exercise. All I'll say is that you shouldn't need to modify any of the guest-side trap handlers ;-).

Once you have a minimal implementation (again, don't worry about more than one arch/OS combination for now) along the lines of the above then we can look at improving that and merging it.

Does this make sense? Please ask if anything is unclear.

One last thing: Unless I'm missing something, you should not need access to the original ELF file to write out a core dump, so please don't rely on that in the core dumping code. Also, for now, don't worry about the fact that the `UKVM_ABORT` hypercall will not be present if the module is not compiled in.

**Comment by mato** [comment]
Missed a couple of points in the cut-n-paste, sorry:

In step (2)(b), `platform_abort()` should invoke the `UKVM_ABORT` hypercall.

In step (3)(b) the only change you should need is to replace the call to `PANIC()` in `cpu_trap_handler()` with a `puts(...)` followed by a call through to `platform_abort(regs)`.

**Comment by mato** [comment]
Regarding the problems with CI mentioned here, yes, it is somehow flaky. I'm in the process of building a new CI that will also run the end to end tests, which is also why the Linux-based CIs are down at the moment. Once that is ready I will look into the reliability issues.

**Comment by niks3089** [comment]
@hannesm, @ricarkol, @mato: Thanks for the comments! Will make changes accordingly and update.

**Comment by niks3089** [comment]
@mato, regarding your points:

1. I've added a new module coredump which dumps the core when ran with --dumpcore option. I am thinking may be it's a good idea to add an optional argument to specify the core filename. What do you think?

2.  I had added the function solo5_dump_core() and platform_dump_core() which essentially did the same thing you mentioned in my initial PR. I just replaced all the instances of *_dump_core() with *_abort(). 

3a/b. These changes are also addressed in the initial PR.
3b: Passing of registers from unikernel to ukvm-bin is only required when a CPU trap happens. If the application directly calls solo5_abort(), platform_abort() can be called with NULL argument as the register information is not required and can be retrieved from the hypervisor(kvm)

Pending things:
1. I will not need to modify the trap handler after @ricarkol changes done for #230. I'll also cleanup the abort/panic and kernel/platform piece of code as well.

2. I'll check if the core file can be constructed without the initial elf file.

**Comment by mato** [comment]
@niks3089:

> I am thinking may be it's a good idea to add an optional argument to specify the core filename. What do you think?

I would use a default of `core.ukvm.<getpid()>`. To avoid possible format-string bugs, if you add an option to change it, make it so that it only adds a _prefix_ to the core dump path, i.e. `--dumpcore-path=/var/crash/ukvm/`. But this is not hugely important right now so you can also leave it out.

> I will not need to modify the trap handler after @ricarkol changes done for #230. I'll also cleanup the abort/panic and kernel/platform piece of code as well.

Ok, I'll check back and do a detailed review once #230 is merged and this is rebased on top of that.

> I'll check if the core file can be constructed without the initial elf file.

Yes please, this should be possible, you'll just need to do some more work to write out the full ELF headers yourself.

**Comment by niks3089** [comment]
@mato 

I'm done with the pending items.  

- I'm creating a core file "core.ukvm.<getpid()> without loading the elf file
-  I've added a new module "dumpcore" which dumps the core with --dumpcore option. 
-  I've added a test(test_assert) to verify the dumping of the core file
-  I'm not looking at the exit status since when a CPU trap happens, a hypercall needs to be invoked to pass the register info from guest to ukvm.
-  I'm always including dumpcore module to avoid hypercall registration failure errors when exception happens when this module is not included.

Please review the changes and let me know what you think.

PS: I've tested only on x86/kvm.

**Comment by ricarkol** [comment]
Thanks @niks3089 

There is just one thing that's kind of bothering me: the fact that `UKVM_HYPERCALL_ABORT` and `platform_exit(SOLO5_EXIT_ABORT);` have to be called together, always. It doesn't make sense to call ABORT without EXIT_ABORT, nor EXIT_ABORT without ABORT. These are the only two ways I can think about merging both of them into `platform_exit(SOLO5_EXIT_ABORT);` (and dump a core if rc=ABORT):

1. add a `regs` argument to `platform_exit` (`platform_exit(SOLO5_EXIT_ABORT, regs);`. Inside `ukvm-bin`, dump the core using `regs` if not NULL, or get them from `KVM` with an ioctl if they are NULL.

2. don't pass `regs` down to `ukvm-bin`, and instead (hackishly) get `regs` by walking up the stack. The value of the `regs` pointer is two frames up, here:
```
#0  platform_exit (status=status@entry=255) at ukvm/platform_lifecycle.c:29
#1  0x00000000001006b3 in _abort (file=file@entry=0x10443c "cpu_x86_64.c", line=line@entry=0x104438 "171", s=s@entry=0x10442d "Fatal trap") at abort.c:60
#2  0x00000000001012e3 in cpu_trap_handler (num=14, regs=0x109048 <cpu_trap_stack+4040>) at cpu_x86_64.c:171 <== HERE
```

Don't want to complicate things any further. I would go for option 1., but don't want to make this an eternal PR. If both of you guys, @niks3089 and @mato, agree about the current solution of having the two hypercalls, I would be perfectly happy with it.

**Comment by niks3089** [comment]
@ricarkol : I am okay with doing 1. We can merge UKVM_ABORT and UKVM_HALT hypercalls and pass the register info and then add an argument to `ukvm_hv_vcpu_loop() ` to pass the register information for the dumpcore module which is invoked based on the return value to dump the core.

Will wait for @mato's opinion

File: kernel/ukvm/platform.c [hunk 1]
```diff
@@ -19,6 +19,7 @@
  */
 
 #include "kernel.h"
+#include "ukvm_guest.h"
```
mato: This is automatically included by `kernel.h` on the ukvm target.

tests/test_assert/test_assert.c
```diff
@@ -0,0 +1,34 @@
+/*
```
mato: Rename to test_abort or similar and just call solo5_abort() directly. (See comments above).

ukvm/ukvm.h
```diff
@@ -209,4 +210,8 @@ int ukvm_gdb_add_breakpoint(struct ukvm_hv *hv, gdb_breakpoint_type type,
 int ukvm_gdb_remove_breakpoint(struct ukvm_hv *hv, gdb_breakpoint_type type,
                                ukvm_gpa_t addr, size_t len);
 
+ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset);
```
mato: I don't see this being used anywhere, left-over change from previous iteration?

ukvm/ukvm_elf.c
```diff
@@ -41,7 +41,7 @@
 
 #include "ukvm.h"
 
-static ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
+ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
```
mato: I don't see this being used anywhere, left-over change from previous iteration?

kernel/abort.c
```diff
@@ -30,6 +30,18 @@ static void puts(const char *s)
     (void)platform_puts(s, strlen(s));
 }
 
+static inline void print_abort_logs(const char *file, const char *line,
```
mato: Not required (see summary).

kernel/abort.c
```diff
@@ -44,12 +56,13 @@ void _assert_fail(const char *file, const char *line, const char *e)
...
-    puts(":");
-    puts(line);
-    puts(": ");
-    puts(s);
-    puts("\n");
+    print_abort_logs(file, line, s);
     platform_exit(SOLO5_EXIT_ABORT);
 }
+
+void _abort_and_dump(const char *file, const char *line, const char *s,
```
mato: Not required (see summary).

kernel/exit.c
```diff
@@ -25,3 +25,9 @@ void solo5_exit(int status)
     log(INFO, "Solo5: solo5_exit(%d) called\n", status);
     platform_exit(status);
 }
+
+void solo5_abort(void)
```
mato: This can just call directly through to `platform_exit(SOLO5_EXIT_ABORT, NULL)` (See summary).

kernel/cpu_x86_64.c
```diff
@@ -168,7 +168,7 @@ void cpu_trap_handler(uint64_t num, struct trap_regs *regs)
         traps[num], regs->ec, regs->rip, regs->rsp, regs->rflags);
     if (num == 14)
         log(INFO, "Solo5: trap: cr2=0x%lx\n", regs->cr2);
-    PANIC("Fatal trap");
+    PANIC_AND_DUMP("Fatal trap", regs, sizeof(struct trap_regs));
```
mato: Extend PANIC instead (see summary).

kernel/kernel.h
```diff
@@ -51,6 +51,13 @@ void _assert_fail(const char *, const char *, const char *)
     __attribute__((noreturn));
 void _abort(const char *, const char *, const char *)
     __attribute__((noreturn));
+void _abort_and_dump(const char *, const char *, const char *, void *, size_t)
```
mato: Not required (see summary).

kernel/kernel.h
```diff
@@ -51,6 +51,13 @@ void _assert_fail(const char *, const char *, const char *)
...
+    __attribute__((noreturn));
+
+#define PANIC_AND_DUMP(s, r, l)                            \
```
mato: Extend PANIC instead (see summary).

kernel/kernel.h
```diff
@@ -103,6 +110,7 @@ int isspace(int c);
 void platform_init(void *arg);
 const char *platform_cmdline(void);
 uint64_t platform_mem_size(void);
+void platform_abort(void *regs, size_t) __attribute__((noreturn));
```
mato: Not required (see summary).

kernel/solo5.h
```diff
@@ -131,6 +131,12 @@ int solo5_console_write(const char *buf, size_t n);
 #define SOLO5_EXIT_ABORT	255
 void solo5_exit(int status) __attribute__((noreturn));
 
+/*
+ * Creates a core dump and exits the application
+ * Note: dumpcore needs to be added as a module
+ */
+void solo5_abort(void) __attribute__((noreturn));
```
mato: Documentation should look a bit different, but I will fix that.

kernel/ukvm/platform.c
```diff
@@ -40,3 +41,16 @@ uint64_t platform_mem_size(void)
 {
     return mem_size;
 }
+
+void platform_abort(void *regs, size_t len)
```
mato: Not required (see summary).

kernel/virtio/platform.c
```diff
@@ -97,6 +97,15 @@ uint64_t platform_mem_size(void)
     return mem_size;
 }
 
+void platform_abort(void *regs, size_t len)
```
mato: Not required (see summary).

ricarkol:
@mato Yes, i agree with all of the above, and specially with extending UKVM_HYPERCALL_HALT with an arg for registers.

niks3089:
@mato , @ricarkol : I've updated the diff based on the new design. Please take a look.
Currently on the monitor end, I'm doing it via hypercall and can it later when the shutdown hooks are introduced.

tests/run-tests.sh
```diff
@@ -193,6 +201,8 @@ run_test ()
                 STATUS=0
             elif [ -n "${WANT_ABORT}" ] && [ "${STATUS}" -eq "255" ]; then
                 STATUS=0
+            elif [ -n "${WANT_COREDUMP}" ] && [ "${STATUS}" -eq "255" ]; then
+                grep -q "unikernel core file" ${LOGS} && STATUS=0
```
ricarkol: could also check that the core file is there

tests/test_abort/test_abort.c
```diff
@@ -0,0 +1,43 @@
...
+            puts(#e);                     \
+            puts("\n");                   \
+            solo5_abort();                \
+        }                                 \
+    } while (0)
+
+int solo5_app_main(const struct solo5_start_info *si __attribute__((unused)))
+{
+    puts("\n**** Solo5 standalone test_assert****\n\n");
+    assert(0);
```
ricarkol: just call `solo5_abort` from here.

ukvm/ukvm_cpu_x86_64.h
```diff
@@ -182,6 +182,33 @@ static const struct x86_sreg ukvm_x86_sreg_unusable = {
...
+    uint64_t ds, es, fs, gs;
+} x86_64_user_regs_struct;
+
+typedef struct {
+    char pad1[32];
+    uint32_t pid;
+    char pad2[76];
+    x86_64_user_regs_struct regs;
+    char pad3[8];
+} x86_elf_prstatus;
```
ricarkol: This looks too Linux or gdb specific to be general to all x86_64 cases (not sure which one defines the format). Where is this documented (just curious)?

ukvm/ukvm_module_dumpcore.c
```diff
@@ -0,0 +1,218 @@
...
+{
+    if (use_dumpcore) {
+        struct ukvm_halt *t =
+            UKVM_CHECKED_GPA_P(hv, gpa, sizeof (struct ukvm_halt));
+
+        if (ukvm_dumpcore_get_regs(hv) == 0) {
+            ukvm_dumpcore(hv, t);
+        }
+    } else {
+        warnx("Not dumping core since --dumpcore option is not set");
```
ricarkol: Better not to print anything. This module will be enabled on most cases, and this will get printed even if the unikernel exits cleanly (rc=0).

ukvm/ukvm_dumpcore_kvm_x86_64.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+int ukvm_dumpcore_get_regs(struct ukvm_hv *hv)
+{
+    if (ioctl(hv->b->vcpufd, KVM_GET_REGS, &hv->b->kregs) < 0) {
+        warnx("KVM: Failed to get register info");
+        return -1;
+    }
+
+
```
ricarkol: extra empty line

ukvm/ukvm_cpu_x86_64.h
```diff
@@ -182,6 +182,33 @@ static const struct x86_sreg ukvm_x86_sreg_unusable = {
...
+    uint64_t ec;
+    uint64_t rip;
+    uint64_t cs;
+    uint64_t rflags;
+    uint64_t rsp;
+    uint64_t ss;
+};
+
+typedef struct {
+    uint64_t r15, r14, r13, r12, rbp, rbx, r11, r10;
```
ricarkol: The order of the registers look suspicious, sure it's right?

ukvm/ukvm_dumpcore_kvm_x86_64.c
```diff
@@ -0,0 +1,141 @@
...
+    }
+    if (write(core_fd, &prstatus, sizeof(x86_elf_prstatus)) < 0) {
+        return -1;
+    }
+    return 0;
+}
+
+int ukvm_dumpcore_get_regs(struct ukvm_hv *hv)
+{
+    if (ioctl(hv->b->vcpufd, KVM_GET_REGS, &hv->b->kregs) < 0) {
```
ricarkol: would prefer to not keeping the latest state of the registers in hv->b->kregs. Keeping this state cached in there seems a bit confusing: let's say we wanted to use it for something else in ukvm, when should the values valid? Plus, in all other cases ukvm just gets the registers using the GET_REGS ioctl and does something with the values immediately (without caching them).

I suggest just getting the regs in a buffer and not stash them in `hv`.

ukvm/ukvm_hv_kvm.h
```diff
@@ -30,6 +30,8 @@ struct ukvm_hvb {
     int vmfd;
     int vcpufd;
     struct kvm_run *vcpurun;
+    struct kvm_regs kregs;
```
ricarkol: Would prefer not to add these two structures. These are just caches of the register state. It's much clearer to just get the register values directly from KVM using ioctls, and not stash/cache anything in `hv`. Otherwise we could get confused and just read or write to this `kregs` assuming that the values are up to date, or that writing to them will somehow make it to the actual VM registers.

ukvm/ukvm_dumpcore_kvm_x86_64.c
```diff
@@ -0,0 +1,141 @@
...
+    *num_notes = 1;
+    return (sizeof(Elf64_Nhdr) + 8 + sizeof(x86_elf_prstatus));
+}
+
+int ukvm_dumpcore_dump_notes(int core_fd,
+        struct ukvm_hv *hv, struct ukvm_halt *info)
+{
+    Elf64_Nhdr nhdr;
+    x86_elf_prstatus prstatus;
+    const char *name = "UNICORE";
```
ricarkol: There is a unikernel related project called "Unicore", better to call it something else

ukvm/ukvm_dumpcore_kvm_x86_64.c
```diff
@@ -0,0 +1,141 @@
...
+
+    memset((void *)&nhdr, 0, sizeof(Elf64_Nhdr));
+    memset((void *)&prstatus, 0, sizeof(x86_elf_prstatus));
+
+    ukvm_hv_fill_elf_prstatus(&prstatus, hv, info);
```
ricarkol: `info` here is the `cookies` coming from the unikernel right? If that's the case then this can be considerably simplified. I suggest two things to make it simpler (in the fist case):

1. replace `info` with `cookies`, or some other name as long as they are called the same
2. getting the registers from the unikernel is split across `ukvm_hv_fill_elf_prstatus` and `ukvm_dumpcore_get_regs`. I would move all the parsing of the guest's `info` values to `ukvm_dumpcore_get_regs` and then use `ukvm_hv_fill_elf_prstatus` to just fill the ELF portion with the register values. Also, `ukvm_dumpcore_get_regs` should be side-effect free: it's only job is getting the registers from KVM and/or the guest (don't stash anything in `hv`).

ukvm/ukvm_guest.h
```diff
@@ -264,9 +264,12 @@ struct ukvm_poll {
     int ret;
 };
 
+#define UKVM_HYPERCALL_MAX_DUMP_INFO_SIZE 100
```
ricarkol: Once you make `ukvm_halt.data` a UKVM_GUEST_PTR (see below), then there is no need for this max len. Every access to UKVM_GUEST_PTR gets properly checked: no weird memory access above the unikernel memory.

ukvm/ukvm_module_dumpcore.c
```diff
@@ -0,0 +1,218 @@
...
+    close(core_fd);
+}
+
+static void hypercall_dumpcore(struct ukvm_hv *hv, ukvm_gpa_t gpa)
+{
+    if (use_dumpcore) {
+        struct ukvm_halt *t =
+            UKVM_CHECKED_GPA_P(hv, gpa, sizeof (struct ukvm_halt));
+
+        if (ukvm_dumpcore_get_regs(hv) == 0) {
```
ricarkol: Only dump core on failure, or when `t->exit_status != 0`

ukvm/ukvm_guest.h
```diff
@@ -264,9 +264,12 @@ struct ukvm_poll {
...
 /* UKVM_HYPERCALL_HALT */
 struct ukvm_halt {
     /* IN */
+    char data[UKVM_HYPERCALL_MAX_DUMP_INFO_SIZE];
```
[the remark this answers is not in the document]
niks3089: @ricarkol, we can't make this a pointer. malloc fails. Remember you debugged this and the issue is that the exception handler uses a different stack which happens to be below the heap, and therefore the sbrk check fails and you suggested we don't have to remove the check instead we use the stack where we specify the maximum bytes for passing the resister info

tests/run-tests.sh
niks3089: We wouldn't know the name of the file since it's core.ukvm.<pid>. Also, if we run ./runtests.sh, multiple times, there will several core files generated eating up the space. We could always delete the core file after each ./runtests but I prefer just naming the core file `core.ukvm`. Thoughts?

kernel/ukvm/platform_lifecycle.c
```diff
@@ -25,11 +25,18 @@ void platform_init(void *arg)
     process_bootinfo(arg);
 }
 
-void platform_exit(int status)
+void platform_exit(int status, void *cookie, size_t len)
 {
     struct ukvm_halt h;
+    memset((void *)&h, 0, sizeof(struct ukvm_halt));
```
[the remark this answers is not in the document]
niks3089: That resulted in the jenkins error:

```
ukvm/platform_lifecycle.c: In function ‘platform_exit’:
ukvm/platform_lifecycle.c:30:12: error: missing braces around initializer [-Werror=missing-braces]
     struct ukvm_halt h = { 0 };
            ^
ukvm/platform_lifecycle.c:30:12: error: (near initialization for ‘h.data’) [-Werror=missing-braces]
```

ricarkol:
Thanks @niks3089 , just submitted a PR against your branch with two suggestions: check for the core file in the tests, and pass data as a pointer.

niks3089:
Thanks @ricarkol. I tested your changes and merged them.

ricarkol:
Thanks @niks3089, all good from my side.

@mato, the changes now address all your concerns from Mar 1 (and following the suggested design).

mato:
@niks3089 Thanks. Could you please rebase this on top of latest `master` and squash some (or all, up to you) of your commits, then I'll review it again in full.

I see that you are basing this off your `master` branch on your Github fork, that is generally the wrong thing to do, so you might want to correct that by pushing to a new branch entirely and opening a new PR in place of this one.

PR #230, opened 2018-01-19
Get the exit status of a unikernel

ricarkol:
The idea with this proposed change is to have unikernels return an exit status. This is useful for some things. First, it improves our negative tests. With an exit code, we don't have to check for strings in the logs to know that there was a failure. Second, it will make dumping cores much easier. PR #228 could be significantly simplified if the unikernel didn't have to decide when to dump a core: it should be all done by ukvm-bin whenever the unikernel exits with a failure. Third, we want to use unikernels in the serverless world, and actions exit status are fundamental for it. Having to check if an action failed or succeeded using strings in stdout is not nice. And fourth, `solo5_app_main` already returns a value (which is basically ignored).

These were all the changes:
- platform_exit and solo5_exit take a new arg called exit_status (0 means success).
- The UKVM_HYPERCALL_HALT hypercall is now used for all halts (not just in
  ARM). Replaced all ukvm uses of the halt instruction with the HALT hypercall. Even
  removed the definition of cpu_halt(). `platform_exit` uses UKVM_HYPERCALL_HALT
  in ukvm, ACPI on virtio, and the halt instruction in bhyve.
- If a ukvm unikernel exits with a halt instruction, ukvm-bin treats it like an error.
- Unikernels can only exit through platform_exit(status), which is called from
  solo5_exit(status), failed asserts (status=1), and faults (status=1).
- ukvm-bin exits with the exit_status of the unikernel.
- Adapted run-tests.sh, so it checks for status=1 for expected failures.
- The exit status is ignored in virtio. There is a way of passing the exit status to QEMU (that's how we return 83 in ), but no equivalent in bhyve. So, just to keep things consistent, let's keep things as they are for QEMU. Unless, somebody wants to return the exit status in QEMU.

@mato: the relevant part of this for the API changes is that this proposes using `solo5_exit(status)` instead of `solo5_exit(void)`.

@niks3089: was going over your PR, and though that having return codes would make your change much simpler. I've been thinking about unikernels returning exit status, so this was a good time for this proposed change.

This was tested on Linux x86 and freebsd x86 (ukvm and virtio backends). Don't have access to any ARM box or muen.


mato:
This is good, I've wanted to standardise on a HALT hypercall for all archs for some time, and I had a version of #201 with the exit status already partly done somewhere...

General comments here, code review coming up separately.

> platform_exit and solo5_exit take a new arg called exit_status (0 means success).

The argument (for these functions, not the ukvm hypercall) can just be called `status`. Also, please define `SOLO5_EXIT_FAILURE` and `SOLO5_EXIT_SUCCESS` in solo5.h and use those throughout.

Note that the change to `solo5_exit()` is an API break, so will need to be coordinated with the various other upcoming breaking changes (#201 #224).

>The UKVM_HYPERCALL_HALT hypercall is now used for all halts (not just in ARM). Replaced all ukvm uses of the halt instruction with the HALT hypercall.

Muen will need adapting for this once it's ready for merging /cc @Kensan are you OK with this approach?

> ACPI on virtio

Presumably you decided not to do that. Very wise, believe me that we do _not_ want to get into ACPI in any shape or form.

> The exit status is ignored in virtio. There is a way of passing the exit status to QEMU (that's how we return 83 in ), but no equivalent in bhyve. So, just to keep things consistent, let's keep things as they are for QEMU. Unless, somebody wants to return the exit status in QEMU.

That's fine, there is no such thing as an "exit status" in legacy-OS-hypervisor land.

kernel/abort.c
```diff
@@ -39,7 +39,7 @@ void _assert_fail(const char *file, const char *line, const char *e)
     puts(": Assertion `");
     puts(e);
     puts("' failed\n");
-    platform_exit();
+    platform_exit(1);
```
mato: Use `SOLO_EXIT_FAILURE`.

kernel/abort.c
```diff
@@ -51,5 +51,5 @@ void _abort(const char *file, const char *line, const char *s)
     puts(": ");
     puts(s);
     puts("\n");
-    platform_exit();
+    platform_exit(1);
```
mato: Use `SOLO5_EXIT_FAILURE`.

kernel/solo5.h
```diff
@@ -110,7 +110,7 @@ int solo5_console_write(const char *buf, size_t n);
 /*
  * Exits the application.
  */
-void solo5_exit(void) __attribute__((noreturn));
+void solo5_exit(int exit_status) __attribute__((noreturn));
```
mato: Use `status` instead of `exit_status`.

Do we want to follow the UNIX `exit()` and type `status` as `int`, or not? I'm happy keeping `int` since it's easier to translate back to an exit status of the `ukvm` process, but just checking for opinions.

kernel/ukvm/platform.c
```diff
@@ -41,10 +41,11 @@ uint64_t platform_mem_size(void)
...
 }
 
-void platform_exit(void)
+void platform_exit(int exit_status)
 {
-    /*
-     * Halt will cause an exit (as in "shutdown") on ukvm.
-     */
-    cpu_halt();
+    struct ukvm_halt status;
```
mato: `struct ukvm_halt h` is clearer

tests/run-tests.sh
```diff
@@ -177,28 +177,54 @@ run_test ()
...
+
+        wait
+        return 99
+    fi
+
+    case ${NAME} in
+    *.ukvm)
+        case ${STATUS} in
+        0|1) 
+            if   [ -z "${WANT_ABORT}" ] && [ "${STATUS}" -eq "0" ] \
```
mato: I don't understand this change. Are you looking at the logs _and_ the exit status? If so, why?

ukvm/ukvm_hv_freebsd_x86_64.c
```diff
@@ -187,9 +187,8 @@ void ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
 
         switch (vme->exitcode) {
         case VM_EXITCODE_SUSPENDED:
```
mato: Please remove this entire `case` block, so that execution falls through to `default:` below.

ukvm/ukvm_hv_freebsd_x86_64.c
```diff
@@ -201,11 +200,19 @@ void ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
                         vme->u.inout.port);
 
             int nr = vme->u.inout.port - UKVM_HYPERCALL_PIO_BASE;
+            ukvm_gpa_t gpa = vme->u.inout.eax;
+
+            /* Guest has halted the CPU. */
+            if (nr == UKVM_HYPERCALL_HALT) {
```
mato: Please do the declaration and load of `gpa` at the start of this `if` block, and keep the original declaration and load of `gpa` below, before the call to `fn(...)`.

ukvm/ukvm_hv_kvm_aarch64.c
```diff
@@ -370,16 +370,19 @@ void ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
                 errx(1, "Invalid guest mmio access: mmio=0x%llx", run->mmio.phys_addr);
 
             int nr = UKVM_HYPERCALL_NR(run->mmio.phys_addr);
+            ukvm_gpa_t gpa = mmio_read32(run->mmio.data);
 
-            /* Guest has halted the CPU, this is considered as a normal exit. */
-            if (nr == UKVM_HYPERCALL_HALT)
-                return;
+            /* Guest has halted the CPU. */
+            if (nr == UKVM_HYPERCALL_HALT) {
```
mato: Please do the declaration and load of `gpa` at the start of this `if` block, and keep the original declaration and load of `gpa` below, before the call to `fn(...)`.

ukvm/ukvm_hv_kvm_x86_64.c
```diff
@@ -182,8 +182,8 @@ void ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
 
         switch (run->exit_reason) {
         case KVM_EXIT_HLT:
```
mato: Please remove this entire `case` block, so that execution falls through to `default:` below.

ukvm/ukvm_hv_kvm_x86_64.c
```diff
@@ -194,12 +194,20 @@ void ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
                 errx(1, "Invalid guest port access: port=0x%x", run->io.port);
 
             int nr = run->io.port - UKVM_HYPERCALL_PIO_BASE;
+            ukvm_gpa_t gpa =
+                *(uint32_t *)((uint8_t *)run + run->io.data_offset);
+
+            /* Guest has halted the CPU. */
+            if (nr == UKVM_HYPERCALL_HALT) {
```
mato: Please do the declaration and load of `gpa` at the start of this `if` block, and keep the original declaration and load of `gpa` below, before the call to `fn(...)`.

ricarkol:
Thanks for the review @mato . Went through all comments. Some answers are above.

kernel/abort.c
ricarkol: done

kernel/abort.c
ricarkol: done

kernel/solo5.h
ricarkol: Changed.

The main reason i can think of is the one you mention: map unikernel errors with ukvm. Would the alternative be something more than an int? something that includes a reason for failure (i.e., a string)?

kernel/ukvm/platform.c
ricarkol: done

tests/run-tests.sh
ricarkol: Yes, it was checking for both the logs and the exit status. Just checking exit status now.

ukvm/ukvm_hv_freebsd_x86_64.c
ricarkol: removed

ukvm/ukvm_hv_freebsd_x86_64.c
ricarkol: changed

ukvm/ukvm_hv_kvm_aarch64.c
ricarkol: changed

ukvm/ukvm_hv_kvm_x86_64.c
ricarkol: changed

ukvm/ukvm_hv_kvm_x86_64.c
ricarkol: changed

Kensan:
> Muen will need adapting for this once it's ready for merging /cc @Kensan are you OK with this approach?

I am fine with the changes if on Muen we can provide a separate implementation for **platform_exit()** analogous to **platform_init()**. The Muen-specific implementation would basically be along the lines of what Virtio does minus the port poking. Maybe we can put **plaform_init** and **platform_exit** into the same C files? I was thinking of something along these lines: aa3b073 02ec8ce. Disclaimer/Warning: compile-tested only. Will do proper testing on Muen once agreement is reached.

The motivation from the Muen PoV is to be able to retain the property of not requiring a separate monitor component.

Kensan:
Another comment: seeing as I think I stumbled upon an off-by-one error in Virtio's **platform_exit()** maybe the log message/output should be handled by some common code which is shared by all platforms? Or should the logging be/remain platform-specific as maybe some platform do not want to have a message at all?

mato:
> I was thinking of something along these lines: aa3b073 02ec8ce.

Yes, that's fine, but see below.

> Another comment: seeing as I think I stumbled upon an off-by-one error in Virtio's platform_exit() maybe the log message/output should be handled by some common code which is shared by all platforms? Or should the logging be/remain platform-specific as maybe some platform do not want to have a message at all?

What's the off-by-one? (lazy)

IMO after/in this PR we can kill logging of exit status / return value from solo5_app_main for `ukvm`. For all the other platforms that have no formalised way of getting an "exit status" out we should keep the logging since there is no other way to get that information.

kernel/solo5.h
mato: > The main reason i can think of is the one you mention: map unikernel errors with ukvm. Would the alternative be something more than an int? something that includes a reason for failure (i.e., a string)?

Eh, forget it. Not worth it. And returning strings in a C-style ABI is horrible anyway.

ukvm/ukvm_guest.h
```diff
@@ -264,4 +264,10 @@ struct ukvm_poll {
     int ret;
 };
 
+/* UKVM_HYPERCALL_HALT */
+struct ukvm_halt {
+    /* OUT */
```
mato: Shouldn't this be `IN` ?

hannesm:
FWIW updates to ocaml-freestanding following this API change are trivial ;) (see https://github.com/hannesm/ocaml-freestanding/commit/45d3360d96cd40e37be2870b8df729bf7a6d0080)

tests/test_ping_serve/test_ping_serve.c
```diff
@@ -26,15 +26,6 @@ static void puts(const char *s)
     solo5_console_write(s, strlen(s));
 }
 
-#define assert(e) do {                              \
```
mato: Why is this removed? Is it unused here?

tests/run-tests.sh
mato: Ok, I'll just trust you that it works, the test code is horrible anyways.

mato:
> FWIW updates to ocaml-freestanding following this API change are trivial ;) (see hannesm/ocaml-freestanding@45d3360)

True, thanks for pointing that out.

Kensan:
> What's the off-by-one? (lazy)

I think the string length here should be 14:

https://github.com/Solo5/solo5/blob/62aa72aec39dbfc129496de083852b8639f9b4d3/kernel/virtio/platform.c#L113

> IMO after/in this PR we can kill logging of exit status / return value from solo5_app_main for ukvm. For all the other platforms that have no formalised way of getting an "exit status" out we should keep the logging since there is no other way to get that information.

Sounds good to me.

mato:
> I think the string length here should be 14:

Good catch. It's harmless (since there's a `\0` at the end) but feel free to fix.

tests/test_ping_serve/test_ping_serve.c
ricarkol: Yes, it was not used. It had a platform_exit() call which I could have updated, but as it was not used, I just removed it.

ukvm/ukvm_guest.h
ricarkol: Yes, it should be an IN, just changed it.

ricarkol:
Thanks @mato, @Kensan , and @hannesm for the review and changes.

hannesm:
I just skimmed through the final diff, it is very short and concise, thanks for your work! :)

Kensan:
> used your changes. Although, I couldn't use your commits as-is so they lost the author. They're not in a branch, so couldn't git fetch them (how did you create those?).

@ricarkol I fetched your branch and pushed it to here yesterday: https://github.com/codelabs-ch/solo5/tree/get-unikernel-exit-status
I amended the 2d43312 commit just now (dropped the **static** keyword - don't know why I added that there...) and pushed them to the above repo. Maybe you can pick the two commits from there if you don't mind?

Kensan:
@ricarkol Oh and thanks for taking care of the off-by-one error too! :)

ricarkol:
No worries @Kensan, just got your commits from the code-labs repo.

kernel/kernel.h
```diff
@@ -65,7 +65,6 @@ void _abort(const char *, const char *, const char *)
 
 /* cpu_<architecture>.c: low-level CPU functions */
 void cpu_init(void);
-void cpu_halt(void) __attribute__((noreturn));
```
mato: Nit: Can we please keep the arch-dependent `cpu_halt()` implementations and call those from the muen and virtio `platform_exit()`? There should not be any arch-dependent inline `asm` outside of the `cpu_*` modules.

kernel/ukvm/kernel.c
```diff
@@ -42,5 +42,5 @@ void _start(void *arg)
     ret = solo5_app_main(cmdline);
     log(DEBUG, "Solo5: solo5_app_main() returned with %d\n", ret);
 
-    platform_exit();
+    platform_exit(ret);
```
mato: I think this should be `solo5_exit(ret)` for consistency, also in the virtio version.

kernel/exit.c
```diff
@@ -20,8 +20,8 @@
 
 #include "kernel.h"
 
-void solo5_exit(void)
+void solo5_exit(int status)
 {
     log(INFO, "Solo5: solo5_exit() called\n");
```
mato: Log the exit status here?

mato:
@ricarkol Regarding refactoring/replacing the tests, can you open a separate issue to discuss this?

kernel/exit.c
ricarkol: logging exit status now

kernel/kernel.h
ricarkol: Added back. The only change now is that `cpu_halt` for ARM is not using the ukvm hypercall anymore. I "borrowed" the freebsd cpu_halt implementation from arm64. Keep in mind that `cpu_halt` is only used from virtio, and we have never tested virtio on arm64.

kernel/ukvm/kernel.c
ricarkol: changed on both ukvm and virtio.

ricarkol:
Thanks for the review @mato . Addressed your comments. Everything was straightforward, except cpu_halt on arm.

avsm:
My only minor comment is to document that applications should keep their return codes between 1<=ret<255 so that the abort code can be reserved for solo5-specific failures such as asserts.

ricarkol:
Thanks @avsm , added some comments into solo5.h:  https://github.com/Solo5/solo5/pull/230/commits/aab9326f4853678d0d7a82ed18499e01a0072134

mato:
Note that a status `>255` will in any case be passed out of ukvm-bin `exit()` with only the least significant 8 bits present (see e.g. the POSIX docs, this is well-known actual behaviour though not documented e.g. in the Linux man page). I'll fix the comments to clarify that once merged.

ricarkol:
Thanks @mato , rebased/squashed into two commits (mine and Adrian's).

PR #233, opened 2018-01-25
CI: Enable tests on surf nodes that support them

mato:
`x86_64-Debian9-gcc630` now uses the new CI infrastructure, based on
https://github.com/roburio/vm. More nodes will migrate to this in due
course.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -15,4 +15,8 @@ set -xe
 uname -a
 cc --version
 ${MAKE}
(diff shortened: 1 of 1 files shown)


mato:
`x86_64-Debian9-gcc630` now uses the new CI infrastructure, based on
https://github.com/roburio/vm. More nodes will migrate to this in due
course.

PR #234, opened 2018-01-26
Disable virtio tests on FreeBSD

mato:
grub-bhyve is a hack, and is unstable under nested virtualization (runs
off into 100% cpu forever). Disable the virtio tests on FreeBSD.


hannesm:
it is great that now even on FreeBSD the ukvm tests are running, thanks for setting up this service! :)

Issue #236, opened 2018-02-02
aarch64 cross compilation on x86_64


aitorhh:
I've see that there is not support for cross compilation in the current build system. Actually, there is a `TODO` item in the code.

I try to create a PR with my working cross compilation

mato:
We now have native aarch64 CI back up again, so I'm closing this as it's not worth the extra clutter in the build system, and there are no plans for "full" support for cross-compilation in any case.

PR #237, opened 2018-02-02
Linux cross compilation support with package gcc-aarch64-linux-gnu

aitorhh:
Tested on Ubuntu 16.04. Related issue #236


mato:
This goes through far too many changes to achieve the desired outcome. You should be able to get to a much smaller patch by:

1. Gating cross compilation entirely on the presence of `TARGET`, i.e. `make TARGET=aarch64-linux-gnu ...` at top-level.
2. In `configure.sh` _don't rename anything_ except the current `TARGET` to e.g. `TARGET_MACHINE`.
3. If `TARGET` is present at `configure.sh` time, set `CC`, `AS`, `LD`, `OBJCOPY` appropriately based on it and generate them into `Makeconf`.
4. At that point all that remains is propagating `UKVM_CC=$(CC)` appropriately for the places where `ukvm-configure` is called.

aitorhh:
There are extra changes to have a clear distinction between `HOSTCC` and `TARGETCC`. These are not needed but simplify the understanding of the build system.

In order to propagate properly the `UKVM_CC` to the ukvm build system it's needed to set the `TARGETCC`, otherwise it won't be propagated.

The changes were done trying to accommodate the cross compilation to the current mixed build system.

mato:
We now have native aarch64 CI back up again, so I'm closing this as it's not worth the extra clutter in the build system, and there are no plans for "full" support for cross-compilation in any case.

PR #241, opened 2018-02-14
malloc: check for clang_major >= 6 to disable null-pointer-arithmetic

hannesm:
disabling `unknown-pragmas` seems to not work on OpenBSD, as reported by @adamsteen in https://github.com/Solo5/solo5/pull/226/commits/6de9cf7a4fb9c5eb89581ed3ae63d9169ef62623

instead, check for the clang major version being `>= 6`


mato:
Yeah, I thought that might be the case :-)

Thanks for tracking it down.

PR #245, opened 2018-02-21
Solo5 public API refresh, take 2

mato:
This is a new take on the Solo5 public API refresh and cleanup (#200, now out of date take 1 in #201).  

Major changes:

- (Tersely) document all public APIs, including their semantics and error handling.
- Remove "type confusion", use the strictest possible types as appropriate throughout.
- Remove use of arch-dependent types where appropriate:
    - Introduce solo5_time_t (for time values), solo5_lba_t (for block addressing).
- Standardise on a solo5_result_t for APIs which may return an error at runtime. For all other APIs, explicitly document that the call will either succeed or abort.
- Block I/O changes:
    - Rename solo5_blk_read_sync() to solo5_block_read(), solo5_blk_write_sync() to solo5_block_write().
    - Introduce a solo5_block_info() call, which replaces the existing _sectors() and _sector_size() calls.
    - Remove the _rw() call as it's not used or implemented anywhere, and I don't think we'll ever need it at this layer.
- Network I/O changes:
    - Rename solo5_net_write() to solo5_net_send() and solo5_net_read() to solo5_net_recv().
    - Introduce a solo5_net_info() call, which replaces the existing solo5_net_mac_str() call, and also returns network device MTU.
- Rename solo5_poll() to solo5_yield().

For the full set of changes, please read through the new kernel/solo5.h (not the diff, that's unreadable as I moved things around to make it read to a human better).

This initial version includes the minimal set of changes required to support the public API change in the ukvm hypercall ABI, and the changes to all test cases. Muen changes have been compile tested only (/cc @Kensan).

I will add more changes to this PR (and/or later PRs, TBD) to further align the ukvm hypercall interface with these changes.

@djwillia @ricarkol @hannesm Please review. 

This is a bit different from what we discussed back in Marrakesh, and includes a few more changes than the old "take 1". I'm reasonably happy with everything in here with the possible exception of the formulation around "Error handling" (@hannesm I'd be especially interested in your opinion on that).

### DIFF
--- a/kernel/solo5.h
+++ b/kernel/solo5.h
@@ -139,9 +139,9 @@ solo5_time_t solo5_clock_wall(void);
  * Suspends execution of the application until either:
  *
  *   a) monotonic time reaches (deadline), or
- *   b) solo5_net_recv() would succeed.
+ *   b) solo5_net_read() would succeed.
  *
- * Returns true if solo5_net_recv() will succeed, otherwise false.
+ * Returns true if solo5_net_read() will succeed, otherwise false.
  *
  * This interface may be extended in the future to allow for selection of I/O
  * events of interest to the application.
--- a/kernel/solo5.h
+++ b/kernel/solo5.h
@@ -23,112 +23,97 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <stdbool.h>
 
 /*
  * Solo5 public APIs.
+ *
+ * These APIs are still under development and may change at any time.
+ *
+ * The documentation in this file reflects the *intent* of the API; current
+ * Solo5 implementations may deviate from the specified behaviour in the short
+ * term. If you find such cases that are not documented here, please file a
+ * bug.
  */
 
 /*
- * FIXME: These APIs are to be considered unstable and have multiple known
- * issues (type confusion, missing const-ness, in/out parameters, inconsistent
- * return values). The comments in this file document the current
- * implementation.
+ * Error handling.
+ *
+ * For APIs returning a solo5_result_t:
+ *
+ *   Unless otherwise documented below, these APIs return only SOLO5_R_OK on
+ *   success.
+ *
+ *   Application developers MUST NOT RELY on these APIs returning
+ *   SOLO5_R_EINVAL or SOLO5_R_EUNSPEC. Solo5 implementations MAY choose to
+ *   abort execution of the application in preference to returning an error
+ *   result on failure.
+ *
+ * For APIs NOT returning a solo5_result_t:
+ *
+ *   These APIs either:
+ *
+ *   a) always succeed under normal operation, or
+ *   b) abort execution of the application on failure.
+ *
+ * For the avoidance of doubt, calling APIs on devices that are not present
+ * WILL result in aborting execution of the application.
  */
+typedef enum {
+    /*
+     * The operation completed successfully.
+     */
+    SOLO5_R_OK = 0,
+    /*
+     * The operation cannot be completed at this time. Retrying an identical
+     * operation at a later time may succeed.
+     */
+    SOLO5_R_AGAIN,
+    /*
+     * Invalid argument.
+     */
+    SOLO5_R_EINVAL,
+    /*
+     * The operation failed due to an unspecified error.
+     */
+    SOLO5_R_EUNSPEC
+} solo5_result_t;
 
 /*
- * Boot-time information passed to application at entry.
+ * Application entry point.
+ *
+ * The application is passed a pointer (*info) to a statically allocated
+ * (struct solo5_start_info) with start of day information; any data in this
+ * structure should be treated as read-only.
+ *
+ * The application receives the (info->heap_start, info->heap_size) of a
+ * single, contiguous, non-executable, read/write memory region it may use for
+ * heap or stack(s) as it sees fit.  At entry, the application is provided with
+ * an initial stack growing down from (info->heap_start + info->heap_size).
+ *
+ * The application MUST NOT make any further assumptions about memory layout,
+ * including where executable code or static data are located in memory.
+ *
+ * Returning from this function is equivalent to calling solo5_exit(<return
+ * value>).
  */
-struct solo5_boot_info {
+struct solo5_start_info {
     const char *cmdline;
     uintptr_t heap_start;
     size_t heap_size;
 };
 
+int solo5_app_main(const struct solo5_start_info *info);
+
 /*
- * Application entry point.
+ * Exit the application, returning (status) to the host if possible.
  *
- * Returning from this function is equivalent to calling solo5_exit(ret).
- */
-int solo5_app_main(const struct solo5_boot_info *bi);
-
-/*
- * Network I/O.
- */
-
-/*
- * Send a single network packet (*data) of size (n). Returns -1 on failure, 0
- * on success.
- */
-int solo5_net_write_sync(uint8_t *data, int n);
-
-/*
- * Receive a single network packet into (*data) up to a maximum size of (*n).
- * Returns 0 on success and size of received packet (*n), -1 on failure or if
- * no packets are available.
- */
-int solo5_net_read_sync(uint8_t *data, int *n);
-
-/*
- * Returns a pointer to the network MAC address, formatted as a C string
- * XX:XX:XX:XX:XX:XX.
- */
-char *solo5_net_mac_str(void);
-
-/*
- * Block I/O.
- */
-
-/*
- * Writes (n) bytes from (*data) starting at the sector (sec) to the block
- * device. (n) does not need to be a multiple of the sector size. Returns 0 on
- * success, -1 on error.
- */
-int solo5_blk_write_sync(uint64_t sec, uint8_t *data, int n);
-
-/*
- * Reads (*n) bytes starting at the sector (sec) from the block device into
- * (*data). (n) does not need to be a multiple of the sector size. Returns 0 on
- * success, -1 on error.
- *
- * XXX: (*n) is not set as an output parameter. ukvm assert()s that exactly
- * (*n) bytes were written to the device.
- */
-int solo5_blk_read_sync(uint64_t sec, uint8_t *data, int *n);
-
-/*
- * Returns the block device sector size.
- */
-int solo5_blk_sector_size(void);
-
-/*
- * Returns the size of the block device in sectors.
- */
-uint64_t solo5_blk_sectors(void);
-
-/*
- * Returns true if the block device is writable.
- */
-int solo5_blk_rw(void);
-
-/*
- * Console I/O.
- */
-
-/*
- * Writes (n) bytes from (*buf) to the console. Returns number of bytes
- * actually written.
- */
-int solo5_console_write(const char *buf, size_t n);
-
-/*
- * Exits the application. The value status is used as the exit status of
- * ukvm-bin.
- *
- * Status values of 255 and above are reserved for solo5.
+ * Status values of 255 and above are reserved for use by Solo5.
  */
 #define SOLO5_EXIT_SUCCESS	0
 #define SOLO5_EXIT_FAILURE	1
 #define SOLO5_EXIT_ABORT	255
+
 void solo5_exit(int status) __attribute__((noreturn));
 
 /*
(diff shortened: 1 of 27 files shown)


mato:
Oh, and since I'm suggesting to everyone interested to read just the new `solo5.h` before looking at the diff at all, here's the link: [solo5.h@`7a4d174`](https://github.com/mato/solo5/blob/7a4d1742b0198de5fc89df48f628d7b265dff9b6/kernel/solo5.h).

hannesm:
@mato thanks for the changes.  IMHO the behaviour of `solo5_console_write` should be described more clearly (that it is not a reliable channel, but a best-effort printing).

I don't understand the introduction of `#define SOLO5_NET_HLEN 14` and its usage - I'd instead specify the `mtu` etc. including header length (i could e.g. send 802.1q tagged ethernet frames, and then the defined `HLEN` should be 16, or am I wrong here?).

are the block devices supposed to return `R_AGAIN` at any time? may be worth - similar to your text about `net` - to explicitly mention whether they do or not.

mato:
@hannesm 

>a) IMHO the behaviour of solo5_console_write should be described more clearly (that it is not a reliable channel, but a best-effort printing).

For the record, I remember having a discussion about this in Marrakesh -- the options were:

1. Keep the status quo (console is best-effort, never "fails" but can lose data).
2. Make console reliable. This would mean that a solo5_console_write() could in theory block forever, and can't be done on virtio for the "default" console anyway since that's always just an emulated serial port.
3. Abort if the host writes less data than we asked it to.

I think that the conclusion was, given the realities of the different targets and the use case of a "system console" we stick with (1), so I'll document that.

> b) I don't understand the introduction of #define SOLO5_NET_HLEN 14 and its usage - I'd instead specify the mtu etc. including header length (i could e.g. send 802.1q tagged ethernet frames, and then the defined HLEN should be 16, or am I wrong here?).

The interface `MTU` is always 1500 bytes, even with 802.1Q. What changes is the maximum frame size (+ 4 bytes) and the header format (+ 802.1Q VLAN tag). So, if we were to eventually support 802.1Q that would mean (for example, off the top of my head) adding an interface `type` to `solo5_net_info`, and a `SOLO5_NET_HLEN_8021Q` = 18 or something like that.

Also, all the host APIs follow the same rule (MTU is the data part, not including any L2 headers or footers). Hence, we should do the same otherwise things will get really confusing.

> c) are the block devices supposed to return R_AGAIN at any time? may be worth - similar to your text about net - to explicitly mention whether they do or not.

My understanding is that as long as our unit of transfer per call remains a single block the call will either succeed, reading/writing the _entire_ block_, or fail, not reading/writing anything. I can clarify the documentation to mention that.

_If_ we extend the calls to support >1 block per operation then we'd have to add a `*read_size` / `*write_size` and the application would need to deal with short reads and writes.

mato:
Pushed minor changes:

- Address @hannesm's comments (8d14fd7)
- Rename `read_size` to `recv_size` in `solo5_net_recv()` (21473b0)
- Rename `solo5_boot_info` to `solo5_start_info` which reads better (9974e8d)

mato:
Updated the link in https://github.com/Solo5/solo5/pull/245#issuecomment-367440830 to point to the tip of my branch rather than a specific commit.

hannesm:
@mato thanks for the explanations, sounds good to me!

kernel/virtio/virtio_blk.c
```diff
@@ -183,47 +183,42 @@ void virtio_config_block(struct pci_config_info *pci)
...
+solo5_result_t solo5_block_write(solo5_lba_t lba, const uint8_t *buf,
+        size_t size)
 {
     assert(blk_configured);
     /*
      * XXX: virtio_blk_op_sync does not support >1 sectors per operation.
      */
-    if (n > VIRTIO_BLK_SECTOR_SIZE)
-        return -1;
+    if (size > VIRTIO_BLK_SECTOR_SIZE)
```
ricarkol: virtio (qemu size) fails IOs smaller than SECTOR_SIZE. So for those, we end up returning an `SOLO5_R_EUNSPEC` (below). It might be simpler to check for IO's being of `SECTOR_SIZE` now, and return an `SOLO5_R_EINVAL` instead.

djwillia:
This looks really good to me, I just have 2 small comments/questions.

All of the block devices that use solo5_lba_t are using block numbers,
right?  I see the "block-addressable" in the comment [here](https://github.com/mato/solo5/blob/v2-api-refresh/kernel/solo5.h#L197), but just
wanted to make sure.  If so, perhaps the comment about what an LBA is
can be clarified.  I'm guessing that LBA stands for "logical block
address".  Would LBN "logical block number" be more clear that it is
not bytes?

I'm wondering if we want a call `solo5_clock_wall`, or whether it
would be better to just expose the `wc_epochoffset` via something like
`solo5_clock_info` (for symmetry with `solo5_net_info` and
`solo5_block_info`) to applications so that they can just do the
addition themselves with `solo5_clock_monotonic`.  I don't know, I'm
just thinking about the relationship (or not) between all the `_info`s
(`solo5_start_info`, `solo5_net_info`, `solo5_blk_info`,
`solo5_clock_info`), and whether what we're really doing is thinking
up some arbitrary functions to access a `start_info` page (as exists
in systems like Linux or Xen).

mato:
On Thursday, 22.02.2018 at 15:22, Dan Williams wrote:
> This looks really good to me, I just have 2 small comments/questions.
> 
> All of the block devices that use solo5_lba_t are using block numbers,
> right?  I see the "block-addressable" in the comment [here](https://github.com/mato/solo5/blob/v2-api-refresh/kernel/solo5.h#L197), but just
> wanted to make sure.  If so, perhaps the comment about what an LBA is
> can be clarified.  I'm guessing that LBA stands for "logical block
> address".  Would LBN "logical block number" be more clear that it is
> not bytes?

"LBA" is fairly standard terminology used in storage (well, ATA), and as
you guessed it does stand for "logical block address", so we may as well
re-use that.

However, as there seems to be quite a bit of confusion on both your and
Ricardo's part about "what units do the block APIs use", there's obviously
something unclear in the APIs as I propose them. Let me explain the thought
process here:

1. Modern storage devices are all block-addressable. The smallest unit of
I/O is always a single block. I don't want to use the term "sector" since
that evokes images of floppy disks and spinning rust, which we don't want
in a modern API.

2. However, we are dealing with **C** APIs, where the established
convention is to pass byte buffers as (_buf_, _size_) where size is
denominated in bytes. I don't want to change this convention since it will
confuse C programmers (see "POLA").

Given (2) the main question is what should the unit of (1) used by the APIs
be. I remember we had a _long_ discussion about this in Marrakesh, and
eventually settled on both units being bytes.

When I did these changes I decided *not* to use bytes as that would mean
more changes when aligning the ukvm hypercalls, which looks like it was a
mistake as that is where the confusion is coming from.

@djwillia, @ricarkol:

So, would you both be happy if all units in block calls are bytes _and_ 
to explicitly return `SOLO5_R_EINVAL` if passed arguments that are not a
multiple of `solo5_block_info.block_size`. (The current restriction on
single-block I/O sizes implies that the only valid value for _size_ under
this scheme would be _block_size_.)

Changing to bytes throughout would imply aligning the ukvm hypercall
interface to _also use bytes_, which I plan to do in a later PR.

> I'm wondering if we want a call `solo5_clock_wall`, or whether it
> would be better to just expose the `wc_epochoffset` via something like
> `solo5_clock_info` (for symmetry with `solo5_net_info` and
> `solo5_block_info`) to applications so that they can just do the
> addition themselves with `solo5_clock_monotonic`.

`wc_epochoffset` is an _implementation detail_ of how we happen to
implement the wall clock today. Our current implementation errs on the side
of implementation simplicity -- in future we may want to support targets
where the two clocks may be entirely decoupled, and/or features such as
wall clock time moving backwards due to NTP corrections on the host. In
order to do that we need to keep the `solo5_clock_wall` call as it is.

> I don't know, I'm just thinking about the relationship (or not) between
> all the `_info`s (`solo5_start_info`, `solo5_net_info`, `solo5_blk_info`,
> `solo5_clock_info`), and whether what we're really doing is thinking up
> some arbitrary functions to access a `start_info` page (as exists in
> systems like Linux or Xen).

You're right that the symmetry between the various `solo5_x_info` calls is
intentional.

However, I really really really(!) do _not_ want to end up with the
"kitchen sink" that is the `start_info` page in Xen. So, while the `info`
calls do provide an easy _extension point_ for future changes, the same
"philosophy" applies as it has until now -- we only add an API (or member
of a `struct solo5_x_info`) when we are _sure_ that we need it.

@ricarkol:

> Some comments and questions, mainly related to block  IO calls.

Responses below, where not already covered by the high-level discussion of the
units in use by the block API above.

> 
> >      }
> -
> -    return ret;
> +    else
> 
> Please change to '} else'. Our formatting rules don't seem to say that `} else {` is needed.

Noted.

> 
> >  
>  /*
> - * Blocks until monotonic time reaches until_nsecs or I/O is possible,
> - * whichever is sooner. Returns 1 if I/O is possible, otherwise 0.
> + * Reads a single block from the block device at offset (lba) into the buffer
> 
> Please mention that the unit of `solo5_lba_t` is blocks (and not bytes), and that the unit of size is bytes.
> 
> >  
>  /*
> - * Returns wall clock time in nanoseconds since the UNIX epoch.
> + * Writes a single block to the block device at offset (lba) from the buffer
> + * (*buf).
> + *
> + * (size) must be equal to solo5_block_info.block_size.
> + *
> + * On success, returns SOLO5_R_OK and writes the entire block to the block
> + * device, i.e. short writes are not possible.
> 
> Two questions:
> 
> 1. "short writes" mean writes smaller than a sector or smaller than the `size` argument?

Smaller than a single block.

> 2. What about not allowing reads/writes smaller than sector. So, what about making the unit of `size` be blocks, just like offset `lba`?

I thought this was explicit ("Writes a single block ...", "... (size) must
be equal to solo5_block_info.block_size"). Anyway, this is covered by the
high-level discussion above.

>Our `virtio-blk` fails on writes and reads smaller than a sector. Small writes would need a read first (no reason to fail reads, but we do so).

That's a bug and virtio-blk (well, the solo5_block_* calls to it) needs to
be fixed to only accept I/O operations in units of (single) blocks. I
didn't claim to fix all implementation details in this PR :-)


> -int solo5_blk_rw(void)
> -{
> -    volatile struct ukvm_blkinfo info;
> -
> -    ukvm_do_hypercall(UKVM_HYPERCALL_BLKINFO, &info);
> -
> -    return info.rw;
> +    info->block_size = bi.sector_size;
> 
> This is just a nit: given that blocks and sectors are the same for us, why not just replacing blocks with sectors? or the other way around.

My choice of "blocks" is explained above. And yes, as I mentioned in the PR
I do plan to submit followup PRs that will align the ukvm hypercall
interface to match the Solo5 public API.

> +solo5_result_t solo5_net_send(const uint8_t *buf, size_t size);
> 
> Mention that these (all net and blk) are blocking (synchronous) calls.

What do you mean? "blocking" in UNIX is generally understood to mean "will
block execution for significant amouts of time until some event happens",
so these calls do not block in that sense.  The fact that they're
synchronous should be obvious, do you see any callbacks anywhere? :-)

> 
> >  {
>      assert(blk_configured);
>      /*
>       * XXX: virtio_blk_op_sync does not support >1 sectors per operation.
>       */
> -    if (n > VIRTIO_BLK_SECTOR_SIZE)
> -        return -1;
> +    if (size > VIRTIO_BLK_SECTOR_SIZE)
> 
> virtio (qemu size) fails IOs smaller than SECTOR_SIZE. So for those, we end up returning an `SOLO5_R_EUNSPEC` (below). It might be simpler to check for IO's being of `SECTOR_SIZE` now, and return an `SOLO5_R_EINVAL` instead.

Yes, needs a fix.

mato:
> When I did these changes I decided *not* to use bytes as that would mean more changes when aligning the ukvm hypercalls, which looks like it was a mistake as that is where the confusion is coming from.

Ah, there was one other reason I used blocks (and introduced solo5_lba_t in the process) -- the existing Mirage code uses blocks, see the comments by @djs55 in "take 1" of this PR: https://github.com/Solo5/solo5/pull/201#discussion_r122971249.

However, given that we have `solo5_result_t` and my point (2) about C APIs conventionally using (_buf_, _size_) stands I'm happy to disregard that and just use bytes throughout the block APIs.

I'm not sure what type to use for the `offset` in block_read and block_write in that case, I'll probably just invent a `solo5_off_t`?

mato:
Another argument for bytes throughout is that it makes retry logic simpler if we end up extending the APIs to support >1 blocks per operation. In such a case they would gain a `*read_size` and `*write_size`, and could return `SOLO5_R_AGAIN` on incomplete operations. A (primitive) retry loop then becomes:

    do {
        result = solo5_block_write(offset, buf, size, &write_size);
        if (result == SOLO5_R_OK)
            break;
        offset += write_size;
        buf += write_size;
        size -= write_size;
    } while (size > 0 && result == SOLO5_R_AGAIN);

With `offset` expressed in blocks you'd have to recompute it each time through the loop.

djwillia:
Thanks for the explanation mato, it all sounds good to me, including the use of bytes for the block device.

ricarkol:
Thanks @mato, I also agree about using bytes offsets instead of blocks. Just to double check, offsets and sizes have to be block aligned right?

I agree with everything else as well.

mato:
On Thursday, 22.02.2018 at 13:45, Ricardo Koller wrote:
> Just to double check, offsets and sizes have to be block aligned right?

Yes, in this scheme both the `offset` and `size` must be a multiple of
`solo5_block_info.block_size` otherwise you will get a `SOLO5_R_EINVAL`
back (or abort per the comments about Error Handling in solo5.h).

Additionally, given that we only support 1 block per operation effectively
`size` must be _equal_ to `solo5_block_info.block_size`.

Is this clear?

> I agree with everything else as well.

Awesome :-)

ricarkol:
> Additionally, given that we only support 1 block per operation effectively
> `size` must be _equal_ to `solo5_block_info.block_size`.

> Is this clear?
Yes, all good from my side.

mato:
@djwillia, @ricarkol:

I've pushed a new version of this. Due to unintentionally doing a `git commit --amend` I had to rebase, squash and force-push at the same time, sorry about that.

Changes since the previous iteration:

- Change to units of bytes in the block APIs, as discussed.
- Had a change of heart and renamed `solo5_net_recv()` and `solo5_net_send()` back to `_read()` and `_write()` respectively to preserve symmetry.
- Documented that this change includes a _minimal_ ukvm hypercall ABI change to `UKVM_HYPERCALL_NETINFO` only. I've deliberately not changed any of the block hypercalls and introduced translation between the units as a _temporary_ measure. I will do the ukvm alignment in a separate PR and will raise an issue to discuss as there are a few things to think about.
- Clarifications and additions to the documentation in `solo5.h` throughout. Notably, tightened up the formulation of _Error handling_ and included the contract for memory layout described in https://github.com/Solo5/solo5/issues/223#issuecomment-366019690.

@djwillia: I think this is good to go, if you're happy with it can you do the honours and hit the big green button?

mato:
@ricarkol: I've also added more test cases to `test_blk` for invalid reads and writes, however (see comment there) I relax the rules currently to just `result != SOLO5_R_OK` intentionally as the implementations have not yet fully caught up with the `SOLO5_R_EINVAL` / `SOLO5_R_EUNSPEC` distinction.

mato:
FYI, my OPAM remote pointing to work in progress versions of the downstream Mirage/Solo5 changes is here: https://github.com/mato/opam-solo5/tree/v2-api-refresh

kernel/solo5.h
```diff
@@ -23,135 +23,248 @@
...
+solo5_time_t solo5_clock_wall(void);
 
 /*
- * Block I/O.
+ * Suspends execution of the application until either:
+ *
+ *   a) monotonic time reaches (deadline), or
+ *   b) solo5_net_recv() would succeed.
+ *
+ * Returns true if solo5_net_recv() will succeed, otherwise false.
```
hannesm: in 142 and 144, you should replace `solo5_net_recv` with `solo5_net_read`

mato: @hannesm Done, thanks for reviewing!

ricarkol:
mato: ACK on "I relax the rules currently to just result != SOLO5_R_OK" for `test_blk`. Will work on the `virtio-blk` fix for that.

PR #246, opened 2018-02-22
WIP: Port tests to bats ("Bash Automated Testing System")

ricarkol:
Just moved our code in `run-tests.sh` to `tests.bats`. `bats-core` is included as a git submodule.

Just in case, checked for these cases:
- not leaking processes in case of failure or success
- all of the console output is printed in case of error (added an `echo $output` in `teardown()` for this)
- ctrl-C works in the middle of tests (added `--foreground` to the timeout command for this)

Ran tests on a local freebsd box, but can't pass the tests in the CI as the VM doesn't have bash. @mato can I add bash into the freebsd VM image you are using? is that an option?


mato:
On Thursday, 22.02.2018 at 13:55, Ricardo Koller wrote:
> Just moved our code in `run-tests.sh` to `tests.bats`. `bats-core` is included as a git submodule.

If you don't mind, can you use "git subtree" instead of a git submodule?
Submodules require extra attention when cloning.

ricarkol:
@mato: using git subtree now. Did `git subtree add --prefix tests/bats-core bats-core master --squash`.

ricarkol:
Using subtree to just get the `libexec` portion of bats-core. Like this (future updates of bats-core should be done like this):

```
git checkout bats-core/master
git subtree split -P libexec -b temporary-bats-branch
git checkout tests-in-bats
git subtree add --squash -P tests/bats-core temporary-bats-branch
```

mato:
@rickarkol Following up on today's discussion, I've added `bash` to the FreeBSD CI VMs. You'll need to rebase this atop `master` to get the CI to work correctly.

As discussed, I'm fine with the changes here, please push to completion and merge whenever you're ready. If you merge before #258, please ping me and we'll figure out the best way forward. (Don't try and PR against #258 before you merge this, that won't work).

ricarkol:
Thanks @mato , merged. So will now will create a change for #258.

PR #247, opened 2018-02-24
avoid truncation of tsc_mult

### LINKED ISSUE #242: clock drift on FreeBSD + AMD hardware
I just deployed Solo5 master (with ukvm backend) to a FreeBSD-12 running on an AMD CPU (this board here https://pcengines.ch/apu2c4.htm).

While using FreeBSD-11/12 and Intel CPUs works out nicely, on AMD I have an immense clock drift:
- start time: `2018-02-16 17:07:16 -00:00: INF [netif] Plugging into 0 with mac ca:0a:6f:fe:a7:8b`
- host time is now `Sat Feb 17 12:48:57 UTC 2018`
- unikernel time `2018-02-16 17:09:26 -00:00: INF [dns_mirage] udp: sending 126 bytes from 53 to 192.168.42.2:32816`

the `ukvm-bin` process used so far `1:17.20` CPU minutes -- which is close to the amount of time it moved forward.

I'm not sure where to look for this issue - to me it looks like it may be either in solo5 or in FreeBSD/BHyve code (I will test a FreeBSD guest soon, and see whether I can also observe a huge clock drift there).

hannesm:
as described in #242, `tsc_mult` is a 32bit unsigned integer which truncates if `tsc_freq` is too small (and the result of `1 * 10^9 << 32 / freq` is bigger than `2^32-1`).

instead of hardcoding the scale to `32`, this patch starts with `32` and decrements until the division fits into 32 bit.  there may be better solutions (i.e. start with a heuristic for `scale` based on the `tsc_freq`).  the proposed patch may also (for very small `tsc_freq`, i.e. `1`) loop infinite.

this PR fixes #242 partially
NB: this PR is incomplete since the code seems to be copy & pasted in both muen and virtio.
EDIT: time moves on my AMD 1GHz CPU still too slow, roughly by a factor of 2 (after 20 hours uptime, the wallclock is 10 hours behind)

### DIFF
--- a/kernel/ukvm/tscclock.c
+++ b/kernel/ukvm/tscclock.c
@@ -77,11 +77,14 @@ int tscclock_init(uint64_t tsc_freq)
      *
      * (0.32) tsc_mult = NSEC_PER_SEC (32.32) / tsc_freq (32.0)
      */
-#if defined(__x86_64__)
-    tsc_mult = (NSEC_PER_SEC << 32) / tsc_freq;
-#elif defined(__aarch64__)
-    tsc_mult = NSEC_PER_SEC / tsc_freq;
-#endif
+    int shift = 32;
+    uint64_t tmp;
+    do {
+      tmp = (NSEC_PER_SEC << shift) / tsc_freq;
+      if ((tmp & 0xFFFFFFFF00000000L) == 0L)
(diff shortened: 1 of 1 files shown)


mato:
Superseded by #250.

PR #250, opened 2018-03-28
ukvm: tscclock: Dynamically calculate tsc_shift

### LINKED ISSUE #242: clock drift on FreeBSD + AMD hardware

mato:
The calculation of tsc_mult will overflow for low values of tsc_freq;
dynamically calculate a suitable tsc_shift where tsc_mult fits into 32
bits.

Additionally, add a DEBUG level log message (visible with --solo5:debug)
to tscclock_init() which prints the values in use on the current
hardware.

TODO: Virtio and muen continue to use a fixed shift of 32 for now,
fixing those targets needs a bit more refactoring to avoid code
duplication.

Thanks to @hannesm for diagnosing the problem and writing the first half
of the fix.

Tested on PCEngines APU2C4 (AMD Embedded G series GX-412TC CPU),
Raspberry Pi 3 model B+ (Broadcom BCM2837B0).

Fixes #242 and also test_time failures observed on ARM64 boards which
have the same root cause.

### DIFF
--- a/kernel/cpu_aarch64.h
+++ b/kernel/cpu_aarch64.h
@@ -36,4 +36,8 @@ static inline uint64_t cpu_cntvct(void)
     return val;
 }
 
+static inline uint64_t mul64_32(uint64_t a, uint32_t b, uint8_t s)
+{
+    return (a * b) >> s;
+}
 #endif /* !ASM_FILE */
--- a/kernel/cpu_x86_64.h
+++ b/kernel/cpu_x86_64.h
@@ -164,15 +164,15 @@ static inline uint64_t cpu_rdtsc(void)
     return ((uint64_t)h << 32) | l;
 }
 
-static inline uint64_t mul64_32(uint64_t a, uint32_t b)
+static inline uint64_t mul64_32(uint64_t a, uint32_t b, uint8_t s)
 {
     uint64_t prod;
 
     __asm__ (
         "mul %%rdx ; "
-        "shrd $32, %%rdx, %%rax"
+        "shrd %%cl, %%rdx, %%rax"
         : "=a" (prod)
-        : "0" (a), "d" ((uint64_t)b)
+        : "0" (a), "d" ((uint64_t)b), "Ic" (s)
     );
 
     return prod;
--- a/kernel/muen/muen-clock.c
+++ b/kernel/muen/muen-clock.c
@@ -50,7 +50,7 @@ uint64_t tscclock_monotonic(void)
         tsc_now = current_start = next_start;
 
     tsc_delta = tsc_now - tsc_base;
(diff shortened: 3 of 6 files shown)


mato:
@Weichen81 For the ARM64 version, I used a generic C implementation of `mul64_32()`, can you verify that this is the right thing to do? I'm not 100% sure that it's a) the most efficient implementation, b) has the same overflow semantics as the x86_64 assembly version.

Weichen81:
@mato  
>a) the most efficient implementation, 

From disassembled code, I think the GCC is smart enough to optimize such simple C code.

>b) has the same overflow semantics as the x86_64 assembly version.

Yes, it has the same overflow semantics as the x86_64.

I had tested this PR on two ARM64 servers, it works OK:
Starting tests at Thu Mar 29 13:51:56 CST 2018
test_hello.ukvm                 : PASSED
test_quiet.ukvm                 : PASSED
test_globals.ukvm               : PASSED
test_exception.ukvm             : PASSED
test_fpu.ukvm                   : PASSED
test_time.ukvm                  : PASSED
test_blk.ukvm                   : PASSED

ricarkol:
Hi @mato , everything looks good. The only piece of code that I can't comment about is the inline assembly for `mul64_32`, I can't find anywhere what does the "Lc" string constraint means. Just double checked that the right code is being generated. Just in case, rewrote the function as regular assembly and it looks like this:

```
ENTRY(mul64_32)
        pushq  %rdx
        pushq  %rcx
        movq   %rdi,%rax
        movb   %dl,%cl
        mul    %rsi
        shrd   %cl,%rdx,%rax
        popq   %rcx
        popq   %rdx
        retq
END(mul64_32)
```

mato:
@ricarkol Thanks for the review! I agree that the GCC inline assembly constraints for this particular case (use of %cl) are not particularly well documented, however, I have verified that the generated code is correct.

PR #253, opened 2018-04-30
CI: allow override of "sudo"

mato:
Some CI setups either don't need "sudo" (container based) or will use
"doas" (OpenBSD).

(This PR is also masquerading as a test of the revived `aarch64-Debian9-gcc630` CI, now with end to end tests)


mato:
Yoohoo, it all works.

PR #256, opened 2018-05-22
Feature to dump solo5 unikernel corefile using ukvm on linux

niks3089:
Please see #228  for more details

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,6 +16,7 @@ Ian Campbell (Docker, Inc.)
 Madhuri Yechuri
 Martin Lucina (robur.io / Center for the Cultivation of Technology)
 Michael LeMay
+Nikhil AP
 Ricardo Koller (IBM)
 Waldir Pimenta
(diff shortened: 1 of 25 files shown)


niks3089:
@mato : Changes are ready for review.

mato:
@ricarkol @niks3089 I started reviewing this yesterday, and in the process of trying to figure out what the right split between host-os-dependent and -independent code should be went down the rabbit hole of looking for a codified API/ABI for `NT_PRSTATUS`, which I found carefully hidden away in `<sys/procfs.h>`. An actual UNIX API I've never used, which appears to date back to the SVR4 days.

In lieu of an actual review I'm now in the process of developing a V3 using these interfaces, based on your V2, which I will PR shortly. Stay tuned.

mato:
@niks3089 @ricarkol See #258. This is not quite ready yet, but the main parts are there, and the resulting core files work fine with GDB on Debian.

PR #258, opened 2018-05-29
V3: Guest core-dump support for ukvm

mato:
This is a partial rewrite of #256 to use the `<sys/procfs.h>` APIs which codifies the _host kernel_ <-> _debugger_ interface for the `NT_PRSTATUS` content of ELF core files. This makes the host-dependent parts of the implementation much clearer, and trivial to implement for Linux/aarch64 and/or FreeBSD.

In the process of going down this rabbit hole I've also rewritten the code to write out only the pages of memory actually touched by the guest, and do that as a sparse file. This speeds up the process on my laptop from "seconds" to "instant".

This is still a work in progress, at least the following need to be addressed:

1. The dumpcore module should be entirely optional.
2. "shutdown hooks" should be called "halt hooks", the former I'd like to use for general module cleanup at tender shutdown time.
3. `ukvm_is_arch_supported()` is wrong, the module should behave similarly to GDB and refuse to init if not supported, the test scripts need to be taught about this.


mato:
@niks3089 Sorry to spoil the fun, but you really did pick one of the most obscure-to-implement features possible. I hope you're OK with me building on your PR. Thanks for getting this started, and for pointing me to a UNIX API I didn't even know existed :-)

niks3089:
@mato: no issues. Looking forward to see how this turns out.

ricarkol:
Hi @mato , you'll to rebase, sorry for that. In the meanwhile, will have a change ready for when you do.

mato:
@ricarkol I've force-pushed the result of `git rebase -X ours master`, which AFAICT did the right thing, i.e. discarded @niks3089 changes to any conflicting files (mainly `run-tests.sh`). So you should be able to do a PR against this and just reinstate a `bats` version.

You can ignore the Travis failure for now, I'll fix that in a bit.

mato:
(FreeBSD CI failure is a transient error here, now fixed)

mato:
@ricarkol Rebased again and force-pushed everything. I think this is good to go, can you give it a final once-over and ack?

@niks3089 This addresses all the points I wanted -- I've deliberately kept the commit history so that you can follow my thought process here, and hopefully have explained in the commit logs the reasoning behind each change. Again, sorry for not going over it with you in a more mentoring fashion, but that would need the two of us in a room together. Thanks again for your work on this.

ricarkol:
Thanks @mato , there is just one issue which I'm not sure is by design or not. Successful runs are also dumping cores if started with `--dumpcore`. I don't think that should be the case. Was that on purpose?

ukvm/ukvm_module_dumpcore.c
```diff
@@ -0,0 +1,252 @@
...
+#else
+
+#error Unsupported target
+
+#endif
+
+static bool use_dumpcore = false;
+
+void ukvm_dumpcore_hook(struct ukvm_hv *hv, int status, void *cookie)
+{
```
ricarkol: This should just exit if status=0. See previous comment.

ricarkol:
Besides that comment above, all LGTM.

mato:
> This should just exit if status=0. See previous comment.

Oh, sorry. I missed that in one of the cleanup passes. Thanks for 
catching it!

I'll fix it tomorrow and merge.

PR #260, opened 2018-06-01
dumpcore on OpenBSD: provide a stub (copied from ukvm_dumpcore_kvm_aarch64)

hannesm:
I tried to compiled solo5 on OpenBSD and ran into trouble, this is the minimal fix I could come up with.


mato:
@ricarkol can you please merge? I'm offline for the rest of today.

ricarkol:
Thanks @hannesm, merging.

PR #261, opened 2018-06-04
Build Muen and Virtio on OpenBSD

adamsteen:
Enable the build or Muen and Virtio on OpenBSD, these complete successfully.

I don't have an easy was of testing the results, sorry.


Kensan:
Do you have some simple instructions on how to setup an OpenBSD system/VM so I could do some tests on Muen?

mato:
I don't think it matters too much right now if the resulting Muen artifacts from an OpenBSD build are untested, as long as the build succeeds. That way it's less likely for anyone developing on OpenBSD to break the build. Merging as-is.

Issue #172, opened 2017-03-24
Renaming ukvm to a more generic name


dstolfa:
As @mato has submitted a [pull request](https://github.com/Solo5/solo5/pull/171) with abstractions that allow for generalization of the ukvm monitor towards other hypervisor APIs such as Hypervisor.Framework on Mac OS X, libvmmapi(bhyve) on FreeBSD and others, it would make sense to change the name of the monitor to something more generic to prevent confusion on non-KVM hypervisors. A possible name might be `umon` or `ukmon`, however, as @mato has pointed out, renaming ukvm has an impact on downstreams such as MirageOS.

amirmc:
A few folks suggested `uvmm`.  It also happens to sound like 'ovum', meaning 'egg', which I think is nice :)

hannesm:
I talked a bit with @mato in Brussels about renaming -- keep in mind that renaming only ever becomes more expensive if done at a later stage.

The current proposal (please correct me if I missed something, @mato) -- which sidesteps the "find a new name" issue entirely by using `solo5`:
- in mirage/mirage: rename `ukvm` to `solo5` as target (leave `muen` and `virtio` as-is)
- in opam-repository: rename `solo5-kernel-ukvm` to `solo5-kernel` (as the package name), leave others as they are
- in solo5 codebase: rename `ukvm-bin` to `solo5`

/cc @mato @djwillia @ricarkol  (please comment here if there's an argument against the above)

djwillia:
That sounds about right to me.  The way I think about it is that what we were calling solo5 are just the bindings for what we were calling ukvm.  As the solo5 api (solo5.h) and the formerly-known-as-ukvm hypercall interface converge (e.g., PR 224 moving malloc is a step toward that) , it  should be more obvious that they are the same thing, hence only needing a single name.  And the word "solo5" does not have the KVM baggage and association that the word "ukvm" does. 

The other slightly confusing part is that there are "other platforms" that can be accessed through the solo5 api (muen and virtio).  For now we are calling them "compatibility".

ricarkol:
Sounds good @hannesm.

Regarding the WIP monitors: seccomp and SGX. What about naming those `solo5-seccomp` and `solo5-sgx`? this would mean that the default `solo5` is the one that uses KVM or VMM.

mato:
The missing piece in this approach is that it leaves the "Solo5 { monitor | bindings } to hardware virtualization" without a name.

Thanking about it, what we want is concise and consistent names for:

[...]

PR #262, opened 2018-06-14
Documentation re-work for 0.3.0 release

mato:
This is a long overdue re-work of the Solo5 documentation, in preparation for the upcoming 0.3.0 release.

Most of this has already been reviewed out of band by @djwillia and @ricarkol, but am opening a PR for wider review here. Note that this does not need to be 100% complete for the release, especially since parts will change after #172.

@Kensan Could you please supply a link to the Muen-specific instructions I could refer to in `building.md`?

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,0 +1,186 @@
+# Building Solo5 and running Solo5-based unikernels
+
+As Solo5 is essentially a piece of "middleware" interfacing unikernel-style
+applications with their host systems, it is not an end-developer product as
+such.
+
+To get started as a developer with Solo5, please refer primarily to the
+instructions provided by unikernel project you intend to develop applications
+with:
+
+- MirageOS: https://mirage.io/wiki/install
+- IncludeOS: http://www.includeos.org/get-started.html
+
+That said, this document provides general information -- not specific to any
+unikernel -- about building Solo5, and running Solo5-based unikernels.
+
+## Building Solo5
+
+Solo5 itself has the following build dependencies:
+
+* a 64-bit Linux, FreeBSD or OpenBSD system (see also below under "Supported
+  targets" for further requirements),
+* a C99 compiler; recent versions of GCC and clang are supported,
+* GNU make,
+* full host system headers (on Linux, kernel headers are not always installed
+  by default).
+
+To build Solo5 from the Git repository, it is sufficient to run `make` (`gmake`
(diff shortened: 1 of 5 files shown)


mato:
Note: Best read by browsing directly to https://github.com/mato/solo5/blob/wip-docs/README.md rather than looking at the diff.

docs/building.md
```diff
@@ -0,0 +1,249 @@
...
+    ip tuntap add tap100 mode tap
+    ip addr add 10.0.0.1/24 dev tap100
+    ip link set dev tap100 up
+
+To set up vmm and the `tap100` interface on FreeBSD, run (as root):
+
+    kldload vmm
+    kldload if_tap
+    sysctl -w net.link.tap.up_on_open=1
+    ifconfig tap100 create 10.0.0.1/24 link0 up
```
hannesm: given the above `sysctl`, no need for the `up` -- and for `link0` I don't understand what that is supposed to do (`man ifconfig` tells me `link[0-2]` are device-specific, and I should look for further details in the device manpage, but `man tap` doesn't tell me any details).

adamsteen: Did you want setup instructions for OpenBSD?

To set up vmm and the `tap100` interface on OpenBSD, run (as root):

    cd /dev
    ./MAKEDEV tap100
    ifconfig tap100 inet 10.0.0.1 netmask 255.255.255.0

mato: @hannesm You're right about `up`, but `link0` is required, otherwise when a ukvm instance is re-started the interface no longer gets packets routed to it correctly (I just re-tested this on `11.1-RELEASE`). 

The documentation is a bit confusing, but `tap(4)` says:

     On the last close of the data device, the interface is brought down (as
     if with ``ifconfig tapN down'') unless the device is a VMnet device, or
     has IFF_LINK0 flag set.  All queued frames are thrown away.  If the
     interface is up when the data device is not open, output frames are
     thrown away rather than letting them pile up.

mato: @adamsteen Thanks, will include those.

Kensan:
> @Kensan Could you please supply a link to the Muen-specific instructions I could refer to in building.md

@mato Here it is on the form of a patch: 1c921519444eca8e7e9dfc6a9444648ac64eed4b

mato:
Rebased and squashed, I think this is good to go, merging.

PR #264, opened 2018-06-18
Ensure tests are run on OpenBSD

adamsteen:
The tests were not getting run on OpenBSD
9 ukvm tests pass
9 virtio tests fail
1 test not run

`abort ukvm (skipped: not implemented for amd64-unknown-openbsd6.3)`

`18 tests, 9 failures, 1 skipped`

### DIFF
--- a/tests/tests.bats
+++ b/tests/tests.bats
@@ -43,7 +43,7 @@ setup() {
     x86_64-*-freebsd*)
       # TODO, just try and run the test anyway
       ;;
-    OpenBSD)
+    amd64-unknown-openbsd*)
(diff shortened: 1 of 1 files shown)


adamsteen:
It would be nice to turn off the virtio tests, maybe not relying on the build flag and add separate run_virtio_tests flag.

mato:
Thanks. For now, can you add a case (gated on `uname -s`) to `tests.bats` to just skip the virtio tests on OpenBSD?

mato:
@adamsteen Looks good, thanks!

PR #268, opened 2018-07-16
vmmvar.h: change vm_exit from union to struct

hb9cwp:
sync with https://github.com/openbsd/src/commit/02ee787fcd78d11b22dcab625f1f53fd122b6cd9

Successfully built and run mirage-skeleton/tutorial/noop on latest snapshot of OpenBSD amd64-current from this morning, using
$ opam --version
2.0.0~rc3
$ ocaml -version
The OCaml toplevel, version 4.06.0

### DIFF
--- a/ukvm/ukvm_hv_openbsd_x86_64.c
+++ b/ukvm/ukvm_hv_openbsd_x86_64.c
@@ -157,7 +157,7 @@ int ukvm_hv_vcpu_loop(struct ukvm_hv *hv)
     if (vrp == NULL)
         err(1, "malloc vrp");
 
-    vrp->vrp_exit = malloc(sizeof(union vm_exit));
+    vrp->vrp_exit = malloc(sizeof(struct vm_exit));
     if (vrp == NULL)
         err(1, "malloc vrp_exit");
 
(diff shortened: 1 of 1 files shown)


adamsteen:
Looks good to me, all the Solo5 tests succeed here too.

When opam 2 and ocam 4.06 ports are merged into current, I will look to announce mirage/solo5 on OpenBSD on misc@.

When OpenBSD 6.4 is released, I will bump the bounds to be a minimum of OpenBSD 6.4, I was thinking Solo5 should target the latest OpenBSD stable.

mato, hannesm or ricarkol would have the final say of course.

mato:
Thanks! Not sure what the Debian CI failure is, but that's unrelated.

@adamsteen What's the timeline for the OpenBSD 6.4 release (or at least API/ABI freeze)? I should probably cut an 0.3.x of Solo5 to match.

hb9cwp:
Thanks!

@mato "The OpenBSD team makes a new release approximately every six months, with the target release dates in May and November."
 from the FAQ https://www.openbsd.org/faq/faq1.html

6.3 was released (early) April 15, 2018. As far as I know, freeze of kernel and ports is usually declared about a month before the release is cut.

PR #270, opened 2018-08-12
Update OpenBSD requirements to 6.4

adamsteen:
With the [cranking of OpenBSD to 6.4](https://marc.info/?l=openbsd-cvs&m=153393283208216&w=2), make this the minimum requirements


mato:
Looks good, Thanks! Will need a point release at some stage, but plenty of time for that until 6.4 comes out.

PR #271, opened 2018-08-15
add support for UKVM_DROP_PRIVILEGES/ukvm_hv_drop_privileges

adamsteen:
as per @mato [email](https://www.mail-archive.com/solo5@lists.h3q.com/msg00043.html) 

and  then move OpenBSD drop privileges into ukvm_hv_drop_privileges.

I am unsure about the text for `Stern Warning, not recommended for production` and function description of `void ukvm_hv_drop_privileges();` but wanted to get the PR submitted for discussion


adamsteen:
Will open a new PR with the above changes applied to after the renaming

PR #273, opened 2018-09-05
OpenBSD only needs -nopie LDFLAGS, remove -fno-pie from CFLAGS

adamsteen:
After reading [PIC, PIE and Sanitizers](https://mropert.github.io/2018/02/02/pic_pie_sanitizers/#mixing-up-binaries)

I realized we don't need to try and push -fno-pie every where and complicate all the builds.

remove -fno-pie from CFLAGS

all the solo5 tests run and succeed(for which OpenBSD should pass), i have been able to build and run (resolving failed) http-fetch(mirage-skeleton/device-usage/http-fetch), which previously failed in gmp-freestanding, zarith-freestanding and nocryto(all with hacks to get that far).

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -155,7 +155,7 @@ case $(uname -s) in
         for f in ${SRCS_AMD64}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}/amd64; done
         for f in ${SRCS}; do cp -f ${INCDIR}/$f ${HOST_INCDIR}; done
 
-        HOST_CFLAGS="-fno-pie -fno-stack-protector -nostdlibinc"
+        HOST_CFLAGS="-fno-stack-protector -nostdlibinc"
         HOST_LDFLAGS="-nopie"
         BUILD_UKVM="yes"
         BUILD_VIRTIO="yes"


mato:
Sounds reasonable. Can you summarise the downstream MirageOS hacks somewhere (e.g. on mirageos-devel@)?

adamsteen:
@mato, I don't think there are any more downstream hacks!

I will need to setup a clean environment, but it should be a matter of installing opam and ocaml, setup a switch to 4.06  and install mirage, and you are good to go.

I will post to mirageos-devel@ with the complete setup of getting mirage installed on OpenBSD, as a priority this week.

OpenBSD still lacks a decent resolver till OPAM2 makes it into ports (held up by a problem with SPARC64)

mato:
@adamsteen Oh, so did you sort out the problem with gmp-freestanding needing a GNU tar (for .xz), or does it just go away when using OPAM 2?

adamsteen:
@mato Sorry looks like i got my wires crossed, there are some fixes required for gtar, see [opam PR 3538](https://github.com/ocaml/opam/pull/3538)

PR #272, opened 2018-08-22
(WIP) V1: New terminology and structure

mato:
(WIP, DO NOT MERGE, can be rebased at any time)

This is an initial version of the new terminology and structure, as outlined in #172. The changes will be made in several parts (as individual commits in this PR):

1. [x] Introduce new directory structure and rename source files and build products.
2. [x] Rename internal symbols and APIs.
3. [x] Rename OPAM packaging. _Note: OPAM installation will not be functional until this step is completed and MirageOS dependencies are updated_
4. [ ] Sundry changes (e.g. mention of ukvm in license headers, etc.)
5. [ ] Documentation updates to reflect changes made.

See individual commits for a detailed description of changes.

/cc @ricarkol @djwillia Please follow along and shout if there's anything you'd like done differently wrt. naming. You can also review the changes themselves, but they will be big and a large chunk will be mechanical.

PR #274, opened 2018-09-12
V2: New terminology and structure

mato:
This is version 2 of the new terminology and structure, as outlined in #172 and supersedes #272 . The changes will be made in several parts (as individual commits in this PR):

1. [x] Introduce new directory structure and rename source files and build products.
2. [x] Rename internal symbols and APIs.
3. [x] Sundry changes (e.g. mention of ukvm in license headers, etc.)
4. [x] Rename OPAM packaging. _Note: OPAM installation will not be functional until this step is completed and MirageOS dependencies are updated_
5. [x] Documentation updates to reflect changes made.

See individual commits for a detailed description of changes.

Changes from V1 (#272):

* The `ukvm` target is now named `hvt` (_hardware virtualised tender_) rather than just `vt`.
* `ukvm_guest.h` is renamed to `hvt_abi.h` (a much better name), moved to `include/solo5` with adjustments to the build system to remove the use of a symlink to enable sharing this file between the `hvt` bindings and tender.
* Minor corrections to overzealous renaming in V1 (`_build-solo5-hvt` vs. `_build_solo5-vt`).
* Where the use of `hvt` would cause internal names to read `hvt_hv_X`, the extra `hv_` has been dropped. Local use of `struct hvt *hv` in tender code has been changed to `struct hvt *hvt` to match (see c67d0effad for details).
* Minor fixes to OPAM install targets (remove redundant use of `include/`).

/cc @ricarkol @djwillia


mato:
This is all ready to merge and I will do so in the next ~12 hours.

@adamsteen Unfortunately there appears to be some kind of transition going on in OpenBSD-6.4-current involving `libc++`, and thus I can't actually `pkg_add` the dependencies needed to build and test this change on a freshly-installed OpenBSD-current snapshot. If you have such a system around, would you mind testing this? If I don't hear from you by morning then I'll just merge and we'll deal with any needed fixes later. All I need is whether or not Solo5 (NOT MirageOS downstreams) builds and the tests pass.

hannesm:
@mato I can run gmake on an OpenBSD 6.4 successfully, it compiles solo5 and tests.  cannot execute tests though (missing hardware features in an emulated system)

mato:
@hannesm Thanks very much for confirming that, that's good enough for now.
@adamsteen No need for you to do anything right now, ignore my previous request.

adamsteen:
@mato Couldn't help my self! OpenBSD compiles, and all tests succeed.

Issue #281, opened 2018-10-08
Quotes in hvt boot parameters get lost


ansiwen:
If the unikernel boot parameter values contain double quotes (") they correctly end up in the Key_gen module if it is compiled for the unix target. In the hvt case the quotes get lost and must be escaped with an addition backspace (\\"). In any case I believe it should be consistent and behave in the same way for unix and hvt targets.

mato:
This is the same underlying code as that referred to in mirage/mirage-solo5#33, let's discuss there.

PR #283, opened 2018-10-10
Escape boot parameters to quoted space-separated format of Parse_argv

ansiwen:
Parse_argv.parse is expecting a space separated list of quoted backslash-escaped strings.  Until now the parameters of argv are just concatenated with spaces, which results in lost information if argv elements contain spaces themselves, and the user has to take care of the quoting and escaping of the boot parameters by themselves. Together with the escape requirements of typical shells, several layers of escaping are necessary, which is tedious and error prone. To avoid this, this change replaces `' '`, `'"'` and `'\'` with `'\ '`, `'\"'` and `'\\'` respectively, and joins them with spaces into a single string, so that Parse_argv.parse can resemble a list of strings identical to the original argv, which makes additional escaping unnecessary.

This changes the usage of boot parameters to a more natural way. For example, instead of writing
```
$ ./solo5-hvt --net=tap100 ./unikernel --json '"{\"foo\": \"bar\"}"'
```
it can now be written as
```
$ ./solo5-hvt --net=tap100 ./unikernel --json '{"foo": "bar"}'
```

Fixes: #281
Fixes: mirage/mirage-solo5#33

### DIFF
--- a/tenders/hvt/hvt_main.c
+++ b/tenders/hvt/hvt_main.c
@@ -35,23 +35,30 @@
 
 #include "hvt.h"
 
+static void push_cmdline(char c, size_t *pos, char *cmdline)
+{
+    if (*pos > HVT_CMDLINE_SIZE - 1) {
+        errx(1, "Guest command line too long (max=%d characters)",
+            HVT_CMDLINE_SIZE - 1);
+    }
+    cmdline[(*pos)++] = c;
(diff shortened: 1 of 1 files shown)


mato:
Closing as per discussion in mirage/mirage-solo5#33, this should be done elsewhere.

Issue #282, opened 2018-10-09
hvt_drop_privileges() behaviour for Linux and FreeBSD


mato:
#276 adds scaffolding for `hvt_drop_privileges()` and an OpenBSD `pledge` implementation.

`hvt_drop_privileges()` is called by the tender just before entering the VCUP loop. i.e.
* after all host resources have been acquired (all modules set up)
* guest ELF has been loaded and memory set up.

We should decide what the behaviour should be on Linux and FreeBSD, to get "secure by default" behaviour. At this time I'm not prepared to add actual `seccomp()` filtering or Capsicum here, but I think that the behaviour should include at least:

1. Droppng privileges to an unprivileged user if the tender is running as `root`.
2. `chroot()` to a suitable directory.

@hannesm: Presumably the FreeBSD case is easier, since there will be a suitable non-privileged UID always available and `/var/empty` could be used for the chroot? 

For the Linux case, I'm not sure what the best default behaviour that is guaranteed to work cross-distribution is -- could `nobody` be used as the unprivileged user? I don't know of any equivalent of `/var/empty`, but there might be something in the LSB/FHS.

A second case is when this is run out of a container in Linux -- can we guarantee that a `nobody` will always be available there?

/cc @adamsteen

hannesm:
@mato sounds sensible for FreeBSD /cc @sg2342 who may have an opinion here

sg2342:
i think for FreeBSD a call to [cap_enter(2)](https://www.freebsd.org/cgi/man.cgi?query=cap_enter&apropos=0&sektion=0&manpath=FreeBSD+11.2-RELEASE+and+Ports&arch=default&format=html) would be the way to go.

sg2342:
turns out, that cap_enter(2) will not work because because ppoll(2) is not permitted in capability mode.
However: the poll(2) code in the FreeBSD kernel is capsicum enabled and shares the relevant parts with the ppoll(2). So the only thing missing in the FreeBSD source tree is an entry for ppoll(2) in sys/kern/capabilities.conf  (and make -C sys/kern/ sysent to regen the syscall configuration).

mato:
I've been thinking about what to do here for the Linux case. There are two cases where the behaviour should be quite different:

1. _In a non-containerized setup_: unlike the BSDs there should be no need to run the hvt tender as root. The only thing the tender needs access to is `/dev/kvm`, which is easily granted through normal permission bits on the device file.
[...]

PR #284, opened 2018-10-22
Do Not Merge: Freebsd sandbox

sg2342:
POC of sandboxing the FreeBSD hvt tender with capsicum. Does not work on unpatched FreeBSD  because of [this bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=232495).
See #282

### DIFF
--- a/tenders/hvt/hvt_freebsd.c
+++ b/tenders/hvt/hvt_freebsd.c
@@ -24,6 +24,9 @@
  */
 
 #include <assert.h>
+#ifndef WITHOUT_CAPSICUM
+#include <capsicum_helpers.h>
+#endif
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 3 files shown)


mato:
Thanks for this -- it's a good start. At this stage, I would prefer that you don't try and do module-specific sandboxing, i.e. work out the minimal set of calls needed for the "non-debug" case (i.e. `core`+`net`+`blk`) and do only that in hvt_drop_privileges().

hannesm:
@mato I agree - there should be two disjoint PRs, esp. since the drop privilege & chroot can be merged as is, whereas the capsicum bits should only be merged once there's a FreeBSD release where ppoll is capsicum-enabled. I appreciate @sg2342 work on capsicum and solo5-hvt, and will deploy this soon. :D

hannesm:
note: the FreeBSD changes have been MFC'ed https://github.com/freebsd/freebsd/commit/4c4449062f3855ce395a35e5b57636d5d768739e and will be part of the next 12 release (12.1, according to https://www.freebsd.org/releng/ there's not yet a schedule for 12.1)

mato:
See #282, this is now out of date (and will be even more soon).

PR #286, opened 2018-11-05
hvt: Basic hvt_drop_privileges on FreeBSD

sg2342:
chroot to /var/empty
drop to nobody

See #282, #284

### DIFF
--- a/tenders/hvt/hvt_freebsd.c
+++ b/tenders/hvt/hvt_freebsd.c
@@ -27,6 +27,7 @@
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <pwd.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


mato:
Thanks, this looks fine. There are some minor coding style issues but those are not your fault (come from the original OpenBSD code), I'll fix those separately.

hannesm:
Since this change, the "cleaning up /dev/vmm/solo5-PID" step does no longer work (installed via `atexit(cleanup_vm)` and `atexit(cleanup_vmfd)` in `hvt_freebsd.c`), since `sysctlbyname("hw.vmm.destroy", ...` receives a permission denied (not output anywhere, see ktrace/kdump below).

I'm not sure what the best way to handle this is... given that `solo5-hvt` indeed created the device (via `hw.vmm.create`), it should be responsible for cleaning up. but in the meantime it lost sufficient privileges to clean it up. Any thoughts? The `vmfd` is no critical AFAICT since unix cleans up all open file descriptors when a process exits. A solution would be to chown `/dev/vmm/solo5-PID` to nobody to enable the dropped priv user to destroy that device.

```bash
  7753 solo5-hvt CALL  write(0x1,0x820dffe10,0x1c)
  7753 solo5-hvt GIO   fd 1 wrote 28 bytes
       "Solo5: solo5_exit(0) called
       "
  7753 solo5-hvt RET   write 28/0x1c
  7753 solo5-hvt CALL  ioctl(0x3,0xc0907601,0x800644010)
  7753 solo5-hvt RET   ioctl 0
  7753 solo5-hvt CALL  close(0x3)
  7753 solo5-hvt RET   close 0
  7753 solo5-hvt CALL  __sysctl(0x7fffffffe330,0x2,0x7fffffffe380,0x7fffffffe378,0x2015fe,0xe)
  7753 solo5-hvt SCTL  "sysctl.name2oid"
  7753 solo5-hvt RET   __sysctl 0
  7753 solo5-hvt CALL  __sysctl(0x7fffffffe380,0x3,0,0,0x800649000,0xa)
  7753 solo5-hvt SCTL  "hw.vmm.destroy"
  7753 solo5-hvt RET   __sysctl -1 errno 1 Operation not permitted
```

PR #291, opened 2018-11-20
support FreeBSD13 (12 was branched some time ago)

hannesm:


tenders/hvt/solo5-hvt-configure
```diff
@@ -113,7 +113,7 @@ case ${TARGET} in
         add_obj hvt_kvm.o hvt_kvm_x86_64.o hvt_cpu_x86_64.o
         add_header hvt_kvm.h hvt_cpu_x86_64.h
         ;;
-    x86_64-*freebsd1[12]*)
+    x86_64-*freebsd1[12-13]*)
```
mato: You want `freebsd1[123]` not `freebsd1[12-13]`.

hannesm: thx, force-pushed the commit

PR #294, opened 2018-11-22
Stack smashing protection et al

### LINKED ISSUE #293: Enable Stack Protector
Currently, we explicitly disable any stack protector that the host toolchain may be using. This is undesirable, instead we would like to compile all C code (bindings and "user") with `-fstack-protector-strong`.

Requires implementing `__stack_chk_fail()`, initialising `__stack_chk_guard` suitably in bindings `_start()` and appropriately adjusting the build system's flags.

References: Original GCC SSP paper at ftp://gcc.gnu.org/pub/gcc/summit/2003/Stackguard.pdf, this [post](https://www.elttam.com.au/blog/playing-with-canaries/) provides a good break-down of current implementations canary setup.

### LINKED ISSUE #296: Unmap zero page; forcibly disable TLS
We are currently running on both hvt and virtio with the page at VA 0x0 mapped. As a standard hardening and debugging measure, we should leave this page unmapped (not present) in the guest so that the following is enforced:

1. `NULL` pointer dereference results in `#PF`.
2. On x86_64, TLS access via `%fs` or `%gs` currently silently returns garbage (well, `0x0` on hvt) as the default values of `IA32_FS_BASE` and `IA32_GS_BASE` are zero. Until such time that we get around to supporting TLS these accesses should likewise result in `#PF`.

mato:
Work in progress, fixes #293 and fixes #296. See commit logs for details.

~~This is currently untested with any downstreams (OCaml) and mysteriously failing test_fpu on hvt (perhaps there's a stack corruption bug in the test code, need to investigate)~~.


mato:
@ehmry What would you like to do for Genode here? Something needs to provide `__stack_chk_fail()` and initialise `__stack_chk_guard`. See the references in #293 for details.

mato:
@hannesm The `clang` failure here is due to lack of support for `-mstack-protector-guard=global`. Would you mind checking for me on 12-RELEASE and -CURRENT if `clang` has decided to implement this by any chance? If not, then we'll have to go down the TLS route which is a bit annoying.

hannesm:
@mato there's still not such an option in clang (6.0.1 on FreeBSD-CURRENT). from `cc --help` output I can't find anything similar to gcc's `-mstack-protector-guard=global`

ehmry:
@mato this not a problem, I have an `ssp` branch that implement this, see afb3feaf76c205727953f4f1fe9c0b1b7ee8b52a and 5fae39bfc79579b9088dd80534fdda2c3e68cc2c.

mato:
@ehmry Okay, that might work, with some caveats which I'll comment on in your commits. However, since @hannesm confirmed `clang` lack of support for a global `__stack_chk_guard`, I need to re-work this using a TLS slot anyway.

ehmry:
Yes, well I would be surprised if I've really done enough here.

mato:
(Note to self: Replace `&= ~(uintptr_t)0xff` with `&= ~(uintptr_t)0xff00` as per suggestion from the internets to not have the \0 on canary boundary)

mato:
Depends on #296.

mato:
Rebased to include commits from #297 in its current state, and @emhry's Genode changes from mato/solo5#3. See commit logs and comments in configure.sh for details.

mato:
Force-pushed: Rebased and merged with #300. Additional tweaks since last rebase:

- moved `__stack_chk_guard` definition into its own module (crt.o)
- `__stack_chk_fail()` now makes minimal use of stack when printing a message
- cosmetics

mato:
Some more fixes for corner cases discovered, tested manually. CI is all green so merging.

Note that downstream MirageOS users of `master` after this will need to use the `gmp-freestanding` and `zarith-freestanding` packages from https://github.com/Solo5/opam-solo5 until mirage/ocaml-freestanding#47 is resolved.

Also note that our SSP implementation requires at least GCC 4.9.0 on Linux, so Solo5 will no longer build on Travis CI environments using `dist: trusty`, updates to downstream `.travis.yml` to switch to `dist: xenial` will be required.

PR #297, opened 2018-11-27
(DO NOT MERGE) virtio,hvt: x86_64: Unmap low memory (zero page+), forcibly disable TLS

mato:
This is the x86_64 implementation of the fix for #296, and a partial aarch64 implementation (test cases only). See commit logs for details.


mato:
@Kensan @ehmry This should not directly affect Muen or Genode as we don't run tests on these targets (yet), but I'd still like to know if either of these targets unmap the page at 0x0? If not, you might want to consider doing that.

tenders/hvt/hvt_cpu_x86_64.h
```diff
@@ -196,11 +196,14 @@ static const struct x86_sreg hvt_x86_sreg_unusable = {
...
+#define X86_PTE0_BASE           0x5000
+#define X86_PTE_SIZE            0x1000
+#define X86_BOOT_INFO_BASE      0x10000
 #define X86_BOOT_INFO_SIZE      0x1000
-#define X86_CMDLINE_BASE        0x6000
+#define X86_PTE0_MAP_START      X86_BOOT_INFO_BASE
+#define X86_CMDLINE_BASE        0x11000
 #define X86_CMDLINE_SIZE        0x2000
-#define X86_GUEST_MIN_BASE      0x8000
+#define X86_GUEST_MIN_BASE      0x100000
```
ricarkol: @mato: please ignore this.

I think IncludeOS is using pages below `0x100000` for something. @niks3089, would you mind checking please? You might be a bit more fresh as you just updated includeos to point to the latest solo5 version.

bindings/virtio/pagetable.S
```diff
@@ -2,11 +2,529 @@
...
+	.quad 0x1fb000 + 0x3
+	.quad 0x1fc000 + 0x3
+	.quad 0x1fd000 + 0x3
+	.quad 0x1fe000 + 0x3
+	.quad 0x1ff000 + 0x3
+
 .align 0x1000
 cpu_pd:
-	.quad 0x0000000000000000 + 0x3 + 0x80
+	.quad cpu_pt0 + 0x3
```
ricarkol: Don't know what flags are the `0x80`s adding, but shouldn't they be added as well to the `pte`s?

tenders/hvt/hvt_cpu_x86_64.c
```diff
@@ -48,23 +48,47 @@ void hvt_x86_setup_pagetables(uint8_t *mem, size_t mem_size)
...
+    {
+        /*
+         * Leave all pages below X86_PTE0_MAP_START not-present in the guest.
+         *
+         * This includes the zero page, boot GDT and the guest's page tables
+         * themselves.
+         */
+        if (paddr < X86_PTE0_MAP_START)
+            continue;
+        *pte0 = paddr | (X86_PDPT_P | X86_PDPT_RW);
```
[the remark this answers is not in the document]
mato: Good point. This will also enforce the documented API (https://github.com/Solo5/solo5/blob/master/include/solo5/solo5.h#L94) which explicitly says "The application MUST NOT make any further assumptions about memory layout".

bindings/virtio/pagetable.S
mato: Definitely not, that bit is `PAT` in a 4kB PTE.

mato: ...and in a PDE it's PS, so in cpu_pd[0] it must be zero, otherwise it'd be a 2MB page.

ehmry:
No Genode does not map the first page. We also never use %fs or %gs in userland, so we notice right away when something is compiled with TLS. NOVA leaves these registers unitialized and I think SeL4 reserves these for its own use.

mato:
@ehmry Thanks for checking, in that case there's nothing you need to do for this change.

mato:
@ricarkol I've addressed your comments, and further cleaned this up. The latest version now includes (expected to fail) tests for aarch64, and explicitly deals with the TLS registers. Could you give this another pass of review, please?

hannesm:
I tested this PR on my FreeBSD-CURRENT, both compilation and execution of all tests are passing as well. FreeBSD clang version 6.0.1, LLD 6.0.1 (FreeBSD 335540-1300000) (this is llvm-ld as well, not GNU ld)

Weichen81:
@mato  TPIDR_EL1 is used for CPU local storage (like percpu data in Linux), so TPIDR_EL0 here used for TLS is OK, just like TLS in Linux process. So I think this code is OK for Arm64.

mato:
@Weichen81 Thanks for confirming.

mato:
Merged as part of #294.

PR #305, opened 2018-12-13
fix build, 'git_t gid' should be 'gid_t gid'

adamsteen:
Solo5 fails to build on OpenBSD, quick fix changing  `git_t gid` to `gid_t gid`


mato:
Thanks, sorry about that.

PR #306, opened 2018-12-14
tenders/hvt: Arm64: Remove redundant parameter in the log message.

xiaobo55x:
It seems that the log() function doesn't care about the actual
parameter count during parsing the message string.

Signed-off-by: Haibo Xu <haibo.xu@arm.com>

### DIFF
--- a/bindings/cpu_aarch64.c
+++ b/bindings/cpu_aarch64.c
@@ -69,7 +69,7 @@ void cpu_trap_handler(struct regs *regs, int el, int mode, int is_valid)
     const uint32_t exception_cls = ESR_EC(regs->esr_el1);
 
     log(INFO, "Solo5: Trap: EL%d %s%s caught\n",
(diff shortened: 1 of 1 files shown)


xiaobo55x:
@Weichen81 @mato

Weichen81:
@xiaobo55x Thanks. I hadn't notice this needless parameter for log before.
@mato Haibo(xiaobo55x) is my team mate. I am very happy he is interesting with solo5 and want to contribute to this project. If you have anything want Arm guys to help, you can @ xiaobo55x and me : )

mato:
Thanks @xiaobo55x. Yes, the definition of `log()` should reflect that it takes a `printf()`-style format string, there is some gcc magic for this which I don't remember offhand. If you have time, maybe take a look at how `printf()`-style functions are defined in e.g. a libc and submit a followup?

xiaobo55x:
Ok, I will have a try.

PR #307, opened 2018-12-17
Enable printf-style parameter check for log() function

xiaobo55x:
Add __attribute__ ((format (printf, 2, 3))) to log() function
to enable printf-style parameter check.

Signed-off-by: Haibo Xu <haibo.xu@arm.com>

PR #308, opened 2018-12-18
Enable printf-style parameter check for log() function

xiaobo55x:
1. Add attribute ((format (printf, 2, 3))) to log() function
   to enable printf-style parameter check;
2. Fix some invocations to match the updated definition.

Signed-off-by: Haibo Xu haibo.xu@arm.com


xiaobo55x:
@mato Fix for #306 #307 is done. Please help have a look.

mato:
Looks good! Thank you!

PR #309, opened 2019-01-02
OpenBSD fixes for 'Enable printf-style parameter check for log() func…

adamsteen:
…tion #308'

OpenBSD is a little more strict, uint64_t is a unsigned long long, which requires 'llu' or 'llx' when loggin.

### DIFF
--- a/bindings/cpu_x86_64.c
+++ b/bindings/cpu_x86_64.c
@@ -164,10 +164,11 @@ static char *traps[32] = {
 
 void cpu_trap_handler(uint64_t num, struct trap_regs *regs)
 {
-    log(INFO, "Solo5: trap: type=%s ec=0x%lx rip=0x%lx rsp=0x%lx rflags=0x%lx\n",
-        traps[num], regs->ec, regs->rip, regs->rsp, regs->rflags);
+    log(INFO, "Solo5: trap: type=%s ec=0x%llx rip=0x%llx rsp=0x%llx rflags=0x%llx\n",
+        traps[num], (unsigned long long)regs->ec, (unsigned long long)regs->rip,
+        (unsigned long long)regs->rsp, (unsigned long long)regs->rflags);
     if (num == 14)
-        log(INFO, "Solo5: trap: cr2=0x%lx\n", regs->cr2);
+        log(INFO, "Solo5: trap: cr2=0x%llx\n", (unsigned long long)regs->cr2);
     PANIC("Fatal trap", regs);
 }
 
--- a/bindings/hvt/tscclock.c
+++ b/bindings/hvt/tscclock.c
@@ -97,8 +97,8 @@ int tscclock_init(uint64_t tsc_freq)
             tsc_shift--;
     } while (tsc_shift > 0 && tsc_mult == 0L);
     assert(tsc_mult != 0L);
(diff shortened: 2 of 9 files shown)


mato:
Hmm, this seems to break every other platform (see CI results), which is not helpful. Not sure how to fix this offhand, am open to ideas. Also unclear why `uint64_t` would be `unsigned long long` on OpenBSD but nowhere else.

adamsteen:
to be concise here is the first build error

```
$ gmake
gmake -C bindings hvt
gmake[1]: Entering directory '/home/asteen/devl/solo5/bindings'
cc -ffreestanding -fstack-protector-strong -fno-stack-protector -nostdlibinc -mno-red-zone -mno-sse -mno-mmx -mno-aes -mno-avx -std=gnu99 -Wall -Wextra -Werror -O2 -g -isystem /home/asteen/devl/solo5/include/crt -I/home/asteen/devl/solo5/include/solo5 -D__SOLO5_BINDINGS__ -c hvt/tscclock.c -o hvt/tscclock.o
hvt/tscclock.c:101:9: error: format specifies type 'unsigned long' but the argument has type 'uint64_t' (aka 'unsigned long long') [-Werror,-Wformat]
        tsc_freq, tsc_mult, tsc_shift);
        ^~~~~~~~
1 error generated.
gmake[1]: *** [GNUmakefile:106: hvt/tscclock.o] Error 1
gmake[1]: Leaving directory '/home/asteen/devl/solo5/bindings'
gmake: *** [GNUmakefile:39: hvt] Error 2
```
line 78 of bindings/hvt/tscclock.c where tsc_freq is defined.
```
int tscclock_init(uint64_t tsc_freq)
```

I am really not sure how to fix this, I just don't know enough.

adamsteen:
I consulted the OpenBSD misc mailing list [1], and few options were presented, the cast seemed the cleanest.

[1] [https://marc.info/?t=154674515500001&r=1&w=2](https://marc.info/?t=154674515500001&r=1&w=2)

mato:
@adamsteen Cool, thanks for investigating. The approach with casting to `(unsigned long long)` looks fine. Would you mind squashing your commits and re-pushing, and I'll merge after.

mato:
Sorry, did not get a notification that you'd force-pushed this. Thanks, merging.

cfcs:
I think the "canonical" solution is to use the `PRIx64` macro from `<inttypes.h>`.

This works on Linux and FreeBSD (I tested), and I believe also OpenBSD (quick [github search](https://github.com/openbsd/src/blob/ef80df5ce8e485ab59324a3f1550e0f35ab9379a/usr.sbin/smtpd/scheduler.c#L86)). This way you don't have to coerce the typecast and thus don't have to keep track of it if you ever want to change any of it (and you retain a tiny bit more of type safety which makes it easier to read imho):
```c
#include <stdio.h>
#include <inttypes.h>
int
main()
{
  uint64_t yo = ~0;
  printf("helo %" PRIx64 "\n", yo);
  return 0;
}
```

mato:
@cfcs I considered that, but we don't have an `inttypes.h` readily available for our appropriated host toolchain and as it's generally provided by the libc it would be hard to get one that works across all our targets. So, unfortunately it's either the cast method or dropping all format string checking.

Issue #304, opened 2018-12-11
hvt: Address space layout randomization


mato:
As part of general hardening, it is desirable to implement ASLR for the hvt target and tender. Current general purpose OSes (Linux, FreeBSD, OpenBSD) have converged on "Static PIE" as the implementation technique of choice, so we should follow their lead here.

A rough plan of the steps involved:

1. Add support to the build system (compiler toolchain checks, linker scripts) to build unikernels as "Static PIE" or equivalent (`-static-pie` itself is merely the _host toolchain's_ idea of how this should work; we need the component parts, i.e. `-fPIE` from the compiler, and the ability to build the final ELF as an `ET_DYN` executable with immediate binding).
2. Add support for loading a minimal subset of `ET_DYN` executables to the ELF loader, initially at a fixed base address equal to the current layout (`0x100000`).
3. Add support to the hvt bindings to perform the minimal amount of relocation required at startup, along the lines of e.g. musl `rcrt1.o`. This will require passing the base address from the tender to the guest.
4. At this point we have "Static PIE" (or equivalent), but no ASLR. Verify that/work with the downstream toolchains (mainly OCaml) are able to cope with this setup.
5. Decide on an address space layout and implement ASLR. This will require some thought on what the guest-virtual to guest-physical to host-virtual mapping should look like, with a view to minimizing complexity in the tender, and taking into account host hypervisor limitations on memory slots, while having a sufficient amount of address space available to randomize into.

PR #310, opened 2019-01-02
Add "spt" (seccomp) target

mato:
Add a new "spt" target ("sandboxed process tender") for x86_64 and
aarch64 Linux hosts.

This target is a userspace process running inside a strict seccomp
sandbox (up to 7 system calls with block and network I/O). The
"solo5-spt" tender serves essentially as a loader and the bindings
implement the Solo5 API by calling the host system call ABI directly.

Note that the "solo5-spt" tender is not modularized on purpose, as it
effectively ceases to operate once control passes to the Solo5
application.

Summary of changes:

tenders/spt: solo5-spt tender
bindings/spt: spt bindings
tests/test_seccomp: spt-specific test that the seccomp policy is being
enforced


mato:
@djwillia @ricarkol:

This is an initial WIP but fully working version, which I'm posting for review. I will be on vacation for 10 days starting tomorrow, but would like to merge soon after getting back if you're happy with it. So please review, I will be responding to comments in the mean time, though with some delay.

This design is simpler to the version you did for nabla containers, in that the tender treats the process as an "empty shell", initialises resources, loads the Solo5 ELF (which is a normal static binary loaded at 1MB as for hvt), applies seccomp rules and then transfers control to the Solo5 guest, after which it is no longer involved. In fact, with some trampoline trickery, it should be possible to unmap the tender from the guest process memory space entirely.

This version passes all tests (and will also pass on the CI once it is updated to have `libseccomp` available), except:
- `test_notls`: due to running as a user process -- this will probably need to be fixed by adding a "solo5_arch_set_tls_base()" or similar to the Solo5 API
- `test_zeropage`: fails due to what looks like an error in tests.bats

Also, crt_init_early() is not currently called due to the TLS issue, so `__stack_chk_guard`, while present, is not randomized.

Further, there is no OPAM/Mirage integration yet, and there are still some `TODO` items marked in the code which need to be looked at.

tenders/spt/spt_core.c
```diff
@@ -0,0 +1,139 @@
...
+
+    abort(); /* spt_launch() does not return */
+}
+
+static int setup(struct spt *spt)
+{
+    seccomp_rule_add(spt->sc_ctx, SCMP_ACT_ALLOW, SCMP_SYS(write), 1,
+            SCMP_A0(SCMP_CMP_EQ, 1));
+    seccomp_rule_add(spt->sc_ctx, SCMP_ACT_ALLOW, SCMP_SYS(exit_group), 0);
+    seccomp_rule_add(spt->sc_ctx, SCMP_ACT_ALLOW, SCMP_SYS(ppoll), 0);
```
cfcs: Since it doesn't look like we make use of the optional opportunity of supplying the fourth argument (`sigset_t sigmask`) to `ppoll`, should we make sure it's NULL?
```c
seccomp_rule_add(spt->sc_ctx, SCMP_ACT_ALLOW, SCMP_SYS(ppoll), 1,
    SCMP_A3(SCMP_CMP_EQ, NULL));
```

tenders/spt/spt_module_block.c
```diff
@@ -0,0 +1,82 @@
...
+    if (diskfile == NULL)
+        return 0; /* not present */
+
+    /* set up virtual disk */
+    diskfd = open(diskfile, O_RDWR);
+    if (diskfd == -1)
+        err(1, "Could not open disk: %s", diskfile);
+
+    spt->bi->blocki.present = 1;
+    spt->bi->blocki.block_size = 512;
```
cfcs: Would it make sense to fill this with `ioctl(diskfd, BLKSSZGET, ...)`?
I don't know much about block devices, so can't tell if that's important.

tenders/spt/spt_module_net.c
```diff
@@ -0,0 +1,270 @@
...
+static int tap_attach(const char *ifname)
+{
+    int fd;
+
+    /*
+     * Syntax @<number> indicates a pre-existing open fd, so just pass it
+     * through without any checks.
+     */
+    if (ifname[0] == '@') {
+        fd = atoi(&ifname[1]);
```
emillon: does this require error checking?

cfcs: I believe the `fcntl` below will fail if you pass it something absurd, but it might make sense to verify that `fd >`  `0` or `1` or `2`?

emillon: I was thinking of cases like `@12m` silently parsed as `12`, but in any case this shouldn't be a  security issue.

cfcs: Agreed, I guess the belt-and-suspenders solution would be to use `strtol` to ensure the whole string was consumed, then checking for fd collisions with existing FDs, but I appreciate there's a complexity trade-off that I don't have a strong opinion about.

tenders/spt/spt_elf.c
```diff
@@ -0,0 +1,227 @@
...
+    Elf64_Half ph_i;
+    Elf64_Phdr *phdr = NULL;
+    Elf64_Ehdr hdr;
+
+    /* elf entry point (on physical memory) */
+    *p_entry = 0;
+    /* highest byte of the program (on physical memory) */
+    *p_end = 0;
+
+    fd_kernel = open(file, O_RDONLY);
```
dinosaure: `fd_kernel` will be closed only if we did not go to `out_error` or `out_invalid` which is fine for me - because `err` will stop the program - but it's a file-descriptor leak.

tenders/spt/spt_core.c
```diff
@@ -0,0 +1,139 @@
...
+    memset(spt, 0, sizeof (struct spt));
+
+    /*
+     * spt->mem is addressed starting at 0, however we cannot actually map it
+     * at 0 due to restrictions on mapping low memory addresses present in
+     * modern Linux kernels (vm.mmap_min_addr sysctl). Therefore, we map
+     * spt_mem at SPT_HOST_MEM_BASE, adjusting the returned pointer and region
+     * size appropriately.
+     */
+    spt->mem = mmap((void *)SPT_HOST_MEM_BASE, mem_size - SPT_HOST_MEM_BASE,
```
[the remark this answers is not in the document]
ricarkol: Just in case, used base address of 0x20000 above, but the same happens with 0x10000.

mato:
>+ spt->mem = mmap((void *)SPT_HOST_MEM_BASE, mem_size - SPT_HOST_MEM_BASE,
>This call fails on my machine. Can't figure it out, so sharing my observations. The program segfaults right after the mmap (the line after mmap is not executed). The faulting address is memsize (like 0x4000000 for --mem=64 below):
‎
That's interesting. Might be the same thing as what's happening on Travis, so it's good you're getting that behaviour reproducibly.

>(gdb) r --mem=64 tests/test_hello/test_hello.spt
>Starting program: >/home/kollerr/solo5/tenders/spt/solo5-spt --mem=64 tests/test_hello/test_hello.spt

To debug the tender/guest transition (on mobile, from memory, so may be a bit off), try the following:

    gdb --args solo5-spt.../test_hello.spt
    b spt_launch
    r
    add-symbol-file.../test_hello.spt 0x100000
    b solo5_app_main
    finish

That should tell you if the crash is before, in, or just after spt_launch.

mato:
>Just in case, used base address of 0x20000 above, but the same happens with 0x10000.

No, that's not the problem. The mmap will fail with EACCES ‎if the base is too low.

Another thing which occurred to me is that you could try subtracting one page from the initial SP value passed to spt_launch and see if that changes anything.

tenders/spt/spt_core.c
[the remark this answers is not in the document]
ricarkol: The segfaults happens before `spt_launch`:

```
kollerr@kollerr-ThinkPad-P50:~/research/solo5$ gdb ./tenders/spt/solo5-spt
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
...
Reading symbols from ./tenders/spt/solo5-spt...done.
(gdb) b spt_launch
Breakpoint 1 at 0x401c20

(gdb) r tests/test_hello/test_hello.spt
Starting program: /home/kollerr/research/solo5/tenders/spt/solo5-spt tests/test_hello/test_hello.spt

Program received signal SIGSEGV, Segmentation fault.
0x0000000020000000 in ?? ()
(gdb) bt
#0  0x0000000020000000 in ?? ()
#1  0x00007fffffffde08 in ?? ()
#2  0x00007fffffffe1fc in ?? ()
#3  0x0000000000000000 in ?? ()
(gdb) 
```
In fact, the log below is not executed either:
```
    spt->mem = mmap((void *)SPT_HOST_MEM_BASE, mem_size - SPT_HOST_MEM_BASE,
            PROT_READ | PROT_WRITE,
            MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);
    printf("here\n"); // not seen (can't break on it after adding a breakpoint on it either)
```
And just in casce, the mmap succeeds according to `strace`.

mato:
> The segfaults happens before `spt_launch`:

Ah, then it's possible your kernel/toolchain is not applying ASLR to the
tender, which we rely on, and the mmap is somehow mucking with the mappings
of the tender itself, rather than failing which is what I expected.

What does the memory map of the tender look like just before the mmap()?
Check with e.g. a breakpoint just before the mmap call and look at
/proc/XXX/maps.

Also the output of 'ldd' on the tender would be helpful.

ricarkol:
Here it is:
```
kollerr@kollerr-ThinkPad-P50:~/research/solo5$ ldd ./tenders/spt/solo5-spt
	linux-vdso.so.1 =>  (0x00007ffca6fd8000)
	libseccomp.so.2 => /lib/x86_64-linux-gnu/libseccomp.so.2 (0x00007f363c64a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f363c280000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f363c88f000)
```

and:

```
kollerr@kollerr-ThinkPad-P50:~/research/unikernel.github.ibm.com$ cat /proc/`pgrep solo5-spt`/maps 
00400000-00403000 r-xp 00000000 08:02 3550490                            /home/kollerr/research/solo5/tenders/spt/solo5-spt
00602000-00603000 r-xp 00002000 08:02 3550490                            /home/kollerr/research/solo5/tenders/spt/solo5-spt
00603000-00604000 rwxp 00003000 08:02 3550490                            /home/kollerr/research/solo5/tenders/spt/solo5-spt
02104000-02125000 rwxp 00000000 00:00 0                                  [heap]
7f5b2fd3a000-7f5b2fefa000 r-xp 00000000 08:02 12980652                   /lib/x86_64-linux-gnu/libc-2.23.so
7f5b2fefa000-7f5b300fa000 ---p 001c0000 08:02 12980652                   /lib/x86_64-linux-gnu/libc-2.23.so
7f5b300fa000-7f5b300fe000 r-xp 001c0000 08:02 12980652                   /lib/x86_64-linux-gnu/libc-2.23.so
7f5b300fe000-7f5b30100000 rwxp 001c4000 08:02 12980652                   /lib/x86_64-linux-gnu/libc-2.23.so
7f5b30100000-7f5b30104000 rwxp 00000000 00:00 0 
7f5b30104000-7f5b30132000 r-xp 00000000 08:02 12980809                   /lib/x86_64-linux-gnu/libseccomp.so.2.3.1
7f5b30132000-7f5b30332000 ---p 0002e000 08:02 12980809                   /lib/x86_64-linux-gnu/libseccomp.so.2.3.1
7f5b30332000-7f5b30348000 r-xp 0002e000 08:02 12980809                   /lib/x86_64-linux-gnu/libseccomp.so.2.3.1
7f5b30348000-7f5b30349000 rwxp 00044000 08:02 12980809                   /lib/x86_64-linux-gnu/libseccomp.so.2.3.1
7f5b30349000-7f5b3036f000 r-xp 00000000 08:02 12980624                   /lib/x86_64-linux-gnu/ld-2.23.so
7f5b30549000-7f5b3054d000 rwxp 00000000 00:00 0 
7f5b3056e000-7f5b3056f000 r-xp 00025000 08:02 12980624                   /lib/x86_64-linux-gnu/ld-2.23.so
7f5b3056f000-7f5b30570000 rwxp 00026000 08:02 12980624                   /lib/x86_64-linux-gnu/ld-2.23.so
7f5b30570000-7f5b30571000 rwxp 00000000 00:00 0 
7fff1aea4000-7fff1aec5000 rwxp 00000000 00:00 0                          [stack]
7fff1aee9000-7fff1aeec000 r--p 00000000 00:00 0                          [vvar]
7fff1aeec000-7fff1aeee000 r-xp 00000000 00:00 0                          [vdso]
ffffffffff600000-ffffffffff601000 r-xp 00000000 00:00 0                  [vsyscall]
```

You are right, that's the problem. `0 - 00400000` is not enough for 1GB. And just to double check, it is enough for 5MBs:

```
kollerr@kollerr-ThinkPad-P50:~/research/solo5$ ./tenders/spt/solo5-spt --mem=4 tests/test_hello/test_hello.spt 

            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Memory map: 4 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x102fff)
Solo5:     rodata @ (0x103000 - 0x103fff)
Solo5:       data @ (0x104000 - 0x105fff)
Solo5:       heap >= 0x106000 < stack < 0x400000

**** Solo5 standalone test_hello ****

Hello, World
Command line is: ''
Solo5: solo5_exit(0) called
```

ricarkol:
Not the prettiest solution, but that's why we ended up using a linker script for the tender in nabla: https://github.com/nabla-containers/solo5/blob/ukvm-linux-seccomp/ukvm/ukvm.lds#L21

That was one of the nice things about loading the unikernel as a dynamic library: `dlopen` takes care of everything.

tenders/spt/spt_module_net.c
```diff
@@ -0,0 +1,270 @@
...
+    return fd;
+}
+
+static int handle_cmdarg(char *cmdarg)
+{
+    if (!strncmp("--net=", cmdarg, 6)) {
+        netiface = cmdarg + 6;
+        return 0;
+    } else if (!strncmp("--net-mac=", cmdarg, 10)) {
+        const char *macptr = cmdarg + 10;
```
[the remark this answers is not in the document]
mato: We rely on the latter: `sscanf` will return the number of items consumed successfully. If it's not exactly 6, then we warn with the "Malformed mac address" message and return `-1` which the caller takes as "setup failed" and terminates with an error.

tenders/spt/spt_module_net.c
mato: This is an "infrastructure" option intended for higher-level stacks/automation, there's not much point/no straightforward way to check at this level if the user has passed in an _incorrect_ `fd`.

OTOH, if an _invalid_ number is passed in, you will effectively get `fd=0` with the current code since there's no way to distinguish an error with `atoi` and `0` may be valid if whatever exec'd solo5-spt closed STDIN and then immediately opened a TAP device.

If you strongly care about the above case in particular please file a separate issue for it since the same code is also in `hvt` and thus best fixed in one go once this is merged.

tenders/spt/spt_elf.c
mato: Thanks, this is minor, can you please file an issue so that I can fix it in both `spt` and `hvt` after this is merged.

mato:
@cfcs and others:

Thanks for the detailed review. I have addressed the error checking for all `libseccomp` calls, and some other small fixes in c5a5ffa.

@ricarkol:

Fix/workaround for the address space conflict issue coming up shortly. I'll also comment in full on the TLS issue, in the mean time I've disabled `test_notls` for spt.

tenders/spt/spt_module_net.c
mato: One more thing: If we do decide to check for the "invalid number" case, it's probably easier to use `sscanf` with a `%u` format specifier rather than `strtoul`.

mato:
@ricarkol:

Regarding the address space conflicts: See 49c62a4 for a workaround, which uses `-Ttext-segment=0x40000000` at link time to locate the solo5-spt at 1GB on systems without PIE support in the toolchain. This version now passes all tests on Travis, which is using Ubuntu Xenial (16.04 LTS) and does not support PIE.

In the long term the true solution to this problem, especially if we want to support spt on 32-bit architectures, is to a) require PIE support on the host _and_ b) implement ASLR using static PIE on the Solo5 (guest) side as described in #304. Then we can mmap() guest memory without MAP_FIXED and avoid the problem altogether. In the short term I think this is an acceptable workaround, LMK if you think otherwise.

Regarding your other proposed solutions/comments: As you can see, a linker script for spt is not necessary for this workaround, asking the linker to set the starting address of the tender's `text` segment is sufficient. As for using the host `dlopen()` and making the guest a "shared library", I'm not in favour of that for many reasons, among others that it introduces the (unsandboxed) host dynamic linker into the TCB which is undesirable.

Regarding TLS:

> The issue, however, is that seccomp_load segfaults at gen_bpf_generate (i.e., some random place). So, I guess there is no other choice than allowing sys_arch_prctl to set FS/GS. Not sure if there should be a solo5_set_fs/gs though.

That's expected. You cannot touch the arch-specific TLS base (i.e. `%fs` on x86_64) and expect _any_ call into libc to continue to work. There are various ways to solve this, however as I mentioned _for the time being_ I've just disabled that test for `spt`. I will refactor the SSP init code in a separate commit to ensure that is done properly for spt.

As for providing an *abstract* `solo5_set_arch_tls_base` API, I think we will need one sooner or later to support language runtimes that insist on using TLS as part of their ABI. However, this is also a separate discussion.

bindings/spt/sys_linux_aarch64.c
```diff
@@ -0,0 +1,155 @@
...
+#define SYS_ppoll         73
+
+int sys_read(int fd, char *buf, size_t size)
+{
+    register long x8 __asm__("x8") = SYS_read;
+    register long x0 __asm__("x0") = fd;
+    register long x1 __asm__("x1") = (long)buf;
+    register long x2 __asm__("x2") = (long)size;
+
+    __asm__ __volatile__ (
```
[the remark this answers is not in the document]
cfcs: I guess for `clock_gettime` we could skip the context switch entirely and just read from the vDSO, but it just occurred to me that your reason for not exposing those to the guest could be that it seems likely vvar/vdso could contain "sensitive" data belonging to the tender, like PID or cpu node or UID/GID or whatever ([looks like](http://man7.org/linux/man-pages/man7/vdso.7.html) that is platform dependent).

tenders/spt/spt_core.c
```diff
@@ -56,13 +56,53 @@ struct spt_module *spt_core_modules[] = {
...
+#else
+    /*
+     * On systems where we are NOT built as a PIE executable, first assert that
+     * -Ttext-segment has been correctly passed at the link step (see
+     * configure.sh), and then check that guest memory size is within limits.
+     */
+    assert((uint64_t)&__executable_start >= (1ULL << 30));
+    if ((uint64_t)(mem_size - 1) >= (uint64_t)&__executable_start) {
+        uint64_t max_mem_size_mb = (uint64_t)&__executable_start >> 20;
+        warnx("Maximum guest memory size (%lu mB) exceeded.",
```
ricarkol: typo: `mB` to `MB`

ricarkol:
>> As for providing an abstract solo5_set_arch_tls_base API, I think we will need one sooner or later to support language runtimes that insist on using TLS as part of their ABI. However, this is also a separate discussion.

Had this issue with golang. Hopefully we don't have to open a syscall. But in any case, that's for another PR. So, all good with me.

Regarding the "shared library" trick, my main reason is gdb. It would be great to find a way of tricking gdb into loading the symbols of the unikernel without having to do a manual `symbol-file` (I guess that's the only extra step). But in any case, that's for another PR. So, all good with me on this one as well.

cfcs:
Should there be a `solo5-bindings-spt.opam` (and `.pc.in` etc) similar to `solo5-bindings-hvt.opam`?
   Here's the two packages I had to install on my systems:
   ```
   depexts: [
     "libseccomp-dev" {os-distribution = "debian"}
     "libseccomp-devel" {os-distribution = "fedora" | os-distribution = "centos" }
   ]
   ```

tests/tests.bats
```diff
@@ -236,6 +299,20 @@ teardown() {
...
+  (
+    sleep 1
+    ${TIMEOUT} 30s ping -fq -c 100000 ${NET_IP} 
+  ) &
+
+  run ${TIMEOUT} --foreground 30s ${SPT_TENDER} --net=${NET} -- ${UNIKERNEL} limit
+  [ "$status" -eq 0 ]
+}
+
 @test "abort hvt" {
```
cfcs: It seems like the `@test "abort spt"` case is missing in this file (`tests/tests.bats`), apart from that the test suite runs for me and crashes in `_exception`,`_zeropage` and `_seccomp` as expected.

mato: That's intentional. The "abort" test case is misnamed, and actually hvt-specific as it tests the functionality of the "dumpcore" module. I have it on my list to rename it and also remove the virtio invocation.

mato:
@cfcs:
> Should there be a solo5-bindings-spt.opam (and .pc.in etc) similar to solo5-bindings-hvt.opam?

Yes, those will come in a separate commit.

tenders/spt/spt_elf.c
dinosaure: Done by #311

bindings/spt/block.c
```diff
@@ -0,0 +1,79 @@
...
+    if (bi->blocki.present) {
+        block_fd = bi->blocki.hostfd;
+        block_capacity = bi->blocki.capacity;
+        block_size = bi->blocki.block_size;
+    }
+}
+
+void solo5_block_info(struct solo5_block_info *info)
+{
+    assert(block_fd >= 0);
```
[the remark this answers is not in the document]
mato: I'd rather not, this is just a weak "is the fd present" check. Depending on how the tender gets started, it's possible that stdin (fd `0`) got closed and the block fd subsequently got assigned `0`.

bindings/spt/sys_linux_aarch64.c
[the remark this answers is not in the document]
mato: I just used the convention that musl uses for aarch64. Isn't the `vsyscall` stuff ARM32 only? As for vDSO, yes, that has various issues associated with it, would prefer to avoid it for now.

tenders/spt/spt_module_block.c
mato: Maybe in future but then the tender needs to check if the underlying fd is backed by a device or a file, so, not right now.

tenders/spt/spt_core.c
```diff
@@ -0,0 +1,139 @@
...
+    spt->mem_size = mem_size;
+
+    spt->sc_ctx = seccomp_init(SCMP_ACT_KILL);
+
+    return spt;
+}
+
+void spt_bi_init(struct spt *spt, uint64_t p_end, char **cmdline)
+{
+    spt->bi = (struct spt_boot_info *)(spt->mem + SPT_BOOT_INFO_BASE);
```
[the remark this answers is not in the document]
mato: Yes, but it might not be so in the future. I'll give this another pass later.

bindings/spt/block.c
```diff
@@ -0,0 +1,79 @@
...
+    
+    /* XXX: These need hard checks on the tender/seccomp side? */
+    if(size != block_size)
+        return SOLO5_R_EINVAL;
+    if(offset & (block_size - 1))
+        return SOLO5_R_EINVAL;
+    if(offset > (block_capacity - block_size))
+        return SOLO5_R_EINVAL;
+
```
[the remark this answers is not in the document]
mato: Done in 4aaad10

tenders/spt/spt_core.c
mato: Done in 4aaad10

bindings/spt/block.c
```diff
@@ -65,7 +68,10 @@ solo5_result_t solo5_block_write(solo5_off_t offset, const uint8_t *buf,
 {
     assert(block_fd >= 0);
     
-    /* XXX: These need hard checks on the tender/seccomp side? */
+    /*
+     * Note that reads beyond capacity are additionally enforced by the
```
hannesm: `reads` -> `writes`

mato: Fixed

bindings/spt/bindings.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include "../bindings.h"
+#include "spt_abi.h"
+
+#define SYS_STDOUT 1
+
+int sys_read(int fd, char *buf, size_t size);
+int sys_write(int fd, const char *buf, size_t size);
+int sys_pread64(int fd, char *buf, size_t size, long long pos);
+int sys_pwrite64(int fd, const char *buf, size_t size, long long pos);
```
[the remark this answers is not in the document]
mato: Addressed in e6f0684 by using `long` and `void *` throughout the `sys_` functions. This matches what musl does, and seems like a good enough solution for now (64-bit archs only).

mato:
This is now almost ready to merge. I've addressed the remaining feedback, added OPAM integration and found and fixed an "interesting" Linux "feature" (see a87c50f).

I have preliminary branches with MirageOS changes required. Folks that would like to test, please ensure you have done an opam update and then install the following pins:

    opam pin add mirage-solo5 https://github.com/mato/mirage-solo5#spt
    opam pin add ocaml-freestanding.0.5.0 https://github.com/mato/ocaml-freestanding#spt
    opam pin add solo5-bindings-spt.0.5.0 https://github.com/mato/solo5#wip-spt
    opam pin add mirage https://github.com/mato/mirage#spt

If there are no major showstoppers, I will merge this over the weekend or latest on Monday, and (after adding some documentation) cut a Solo5 0.5.0 release.

tenders/spt/spt_module_block.c
```diff
@@ -0,0 +1,106 @@
...
+    spt->bi->blocki.capacity = lseek(diskfd, 0, SEEK_END);
```
cfcs: I guess we only support 64-bit platforms at the moment(?), but would it make sense to use `lseek64()` instead here to prevent running into the `-1` / `EOVERFLOW` when dealing with large files or block devices on 32bit?

bindings/spt/block.c
```diff
@@ -0,0 +1,85 @@
...
+{
+    assert(block_fd >= 0);
+
+    /*
+     * Note that writes beyond capacity are additionally enforced by the
+     * tender's seccomp policy.
+     */
+    if(size != block_size)
+        return SOLO5_R_EINVAL;
+    if(offset & (block_size - 1))
```
cfcs: Why keep this check (alignment) on the C side if not in the seccomp rules?

bindings/spt/sys_linux_aarch64.c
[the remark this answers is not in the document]
cfcs: My linux-fu turned out to be very outdated; `vsyscall` is the ancient pre-ASLR method, now probably only found on old stuff.
They split that up into an r/w section `vvar` and an executable `vdso` for data and syscall wrappers respectively.
But if vDSO is a pain and this is easier then fair enough. :)

bindings/spt/block.c
```diff
@@ -47,7 +47,10 @@ solo5_result_t solo5_block_read(solo5_off_t offset, uint8_t *buf, size_t size)
 {
     assert(block_fd >= 0);
 
-    /* XXX: These need hard checks on the tender/seccomp side? */
+    /*
+     * Note that writes beyond capacity are additionally enforced by the
+     * tender's seccomp policy.
+     */
     if(size != block_size)
```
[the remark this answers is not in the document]
mato: Not quite. The Solo5 API should return `SOLO5_R_EINVAL` if you pass invalid arguments. If this check wasn't here then it would just kill the guest with `SIGSYS` due to the seccomp filter violation. While that does the job, it makes it a pain to debug what might just be a programmer error.

tenders/spt/spt_module_block.c
mato: Good point, I'll fix that (also applies to hvt).

bindings/spt/block.c
mato: For the same reason as keeping the "invalid offset" check here (see previous comment) and additionally, the Solo5 test cases (test_blk) test for a `SOLO5_R_EINVAL` return.

mato: All good points. Unfortunately, your suggestion won't work with the current implementation as libseccomp _does not guarantee the order in which rules are applied_ (see e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1035748).

I've thought about the alternatives but don't have a clear answer yet, so will leave this as-is for now on the basis that:

1. The seccomp ruleset we apply should be as minimal as possible, and composed only of rules which either `ALLOW` something or don't match at all and thus cause the filter to return the default `SCMP_ACT_KILL`.
2. I'd like to keep the semantics of the Solo5 layer as "if you pass something which is incorrect, but not obviously _malicious_, you should get an error return back rather than crash".
3. Conversely, if possible I'd like to ensure that we don't accidentally "mask errors" from the system call (by turning `EWHATEVER` into an `EINVAL`).

I think this is good enough for an initial "experimental" release and will file an issue for this so that we don't forget about it and can re-visit it in the (near) future. It may well be that the above points are not the behaviour we want in the long term.

mato: One other important point to make here is that reading or writing to an unaligned offset is _not_ a security issue. The reason that limitation exists in the Solo5 API is twofold. First, to provide a lowest common denominator that supports porting bindings to targets which cannot deal with unaligned block access, such as _virtio_. Second, the Solo5 block API semantics are defined as "the operation will either succeed in it's entirety, or fail". In practice this currently means limiting reads or writes to one block per operation.

Regarding your argument about the security / host kernel surface impact. While it's theoretically possible to argue that an unaligned read or write on a target such as spt might execute more kernel code than purely aligned operations, in practice that is a very weak argument since the host kernel's I/O scheduler is free to do whatever it wants with the underlying operations at the device driver layer.

tenders/spt/spt_module_block.c
mato: Fixed in b9bbe8e, filed #313 for hvt fix.

bindings/spt/block.c
cfcs: **First: I don't see any problem with this patchset as it stands now either, don't let this block you from merging.** :-)

Re: `libseccomp` not letting you order the rules: what the fuck, good catch!

1) Fair enough., good point against `SCMP_ACT_ERRNO()`, would need a unique error number for `SOLO5_R_EINVAL` to avoid that.

2) Also fair, although it's pretty context-dependent what is malicious or not: Crashing on supposed malicious values can turn bugs that normally would not have any impact into DoS vectors, effectively upgrading the impact of the bug. OTOH things falling through gracefully can also be a problem, and can enable worse things than crashing, which arguably is in the low end, and easier to detect + fix.
  In this case we have duplication of the enforcement logic, and the "soft" errors hiding the "hard" errors, which makes it trickier to unit test.

4. `Succeed or fail`: Oh, speaking of that, would it make sense to open the block_fd with `O_DSYNC` or  call `fsync()` in `platform_exit()` since we don't expose a `sync` primitive and don't do direct I/O?

5) (*writing to an unaligned offset is not a security issue*): In the ideal world, you're right. In practice you're most probably also right, but in theory the more flexibility we have, the more exposed we are, and this exposure is multiplied by the number of different backends we support. See for example [CVE-2011-1750](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2011-1750).

mato:
Added some initial documentation for spt in 6fd3f90. Looking good for an initial release, so merging.

tenders/spt/spt_module_net.c
dinosaure: Just a note from what I found on my twitter feed:

> A note about the ato{i,l,f} functions – their behavior is undefined if the input cannot be parsed correctly. These functions need not set errno even. Except for behavior on error, atoi is equivalent to (int)strtol(nptr, (char **)NULL, 10). The strto{d,l,ul} functions should always be preferred, because they provide proper error reporting and choice of base.

here: https://begriffs.com/posts/2019-01-19-inside-c-standard-lib.html

PR #316, opened 2019-01-30
improve FreeBSD create / destroy behaviour

hannesm:
more specifically, hw.vmm.destroy needs privileges, so we can no longer do this
in atexit() with dropped privileges.

instead, look for solo5-YYY before hw.vmm.create, and if it exists, destroy it
(unless it is still running). Testing whether the VM is running is not possible,
we use an exclusive FS lock for this. This mechanism works under the assumption
that solo5-hvt owns the /dev/vmm/solo5-* namespace.

Contribution was developed pair-programming with @mato at Espresso Library on
Sat 26th January 2019. Code likely needs some cleanups.

### DIFF
--- a/tenders/hvt/hvt_freebsd.c
+++ b/tenders/hvt/hvt_freebsd.c
@@ -47,26 +47,14 @@
 #include "hvt.h"
 #include "hvt_freebsd.h"
 
-/*
- * TODO: To ensure that the VM is correctly destroyed on shutdown (normal or
- * not) we currently install an atexit() handler. The top-level API will need
- * to be changed to accomodate this, e.g. by introducing a hvt_shutdown(),
- * however this is incompatible with the current "fail fast" approach to
- * internal error handling.
- */
 static struct hvt *cleanup_hvt;
-
-static void cleanup_vm(void)
-{
-    if (cleanup_hvt != NULL)
-        sysctlbyname("hw.vmm.destroy", NULL, NULL, cleanup_hvt->b->vmname,
-                strlen(cleanup_hvt->b->vmname));
-}
 
 static void cleanup_vmfd(void)
 {
-    if (cleanup_hvt != NULL && cleanup_hvt->b->vmfd != -1)
+    if (cleanup_hvt != NULL && cleanup_hvt->b->vmfd != -1) {
         close(cleanup_hvt->b->vmfd);
+        cleanup_hvt->b->vmfd = -1;
+    }
 }
 
 struct hvt *hvt_init(size_t mem_size)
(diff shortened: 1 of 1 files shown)


mato:
@hannesm Where do you want me to do the cleanups to this code? OK to push to your branch so that we don't have too much PR churn?

hannesm:
@mato sure, feel free to push to my branch.

mato:
@hannesm @sg2342:

I've pushed a cleaned up version to hannesm's branch. However, after stress-testing this some more, I am **extremely unhappy** with this "fix". It does not fix the underlying problem, namely that VMs **and their resources** are not cleaned up after `solo5-hvt` exits.

In practice this means that if I run two instances of the following simple stress test:

````
$ while [ 1 ]; do tests/test_hello/solo5-hvt tests/test_hello/test_hello.hvt >/dev/null; done
````

... then I can easily bring a (nested) FreeBSD with 2GB of RAM to its knees due to virtual memory exhaustion:

````
procs  memory       page                    disks     faults         cpu
r b w  avm   fre   flt  re  pi  po    fr   sr da0 pa0   in    sy    cs us sy id
1 0 1 1.4T   14M  1525   1   1   0  1832 7054 9054   0   21  3033   248  1  2 97
````

Note the `avm` figure! The figures come back to normal after I somehow managed to kill the stress tests **and** run a `bhyvectl --destroy` on all the leaked VMs.

Now, I don't think that it is responsible of us to produce a software component that can bring the host down!

So, what to do? The least work _for me_ would be to just kill dropping privileges on FreeBSD for now. The other option is to spend half a day figuring out if a solution using `fork()` with the parent process retaining `root` privileges, doing nothing but `waitpid()` and cleaning up after the child would work. This solution, while also somewhat hacky, especially internally since a function now has a side-effect of `fork()`, would at least reliably clean up after itself.

The *real* fix is of course to fix the FreeBSD vmm APIs. However, any fix that does happen (e.g. a subset of what's in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=213689) will take time and manpower (not me!). So, I think we need a proper workaround as described in either of the options above in the mean time.

Please let me know how you'd like to proceed.

sg2342:
FreeBSD has [cap_sysctl(3)](https://www.freebsd.org/cgi/man.cgi?query=cap_sysctl&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html). It can be used to set up the proxy process for sysctls before entering capability mode (and privdrop). But since solo5  uses ppoll(3) and the capsicum releated changes will land in FreeBSD 12.1-RELEASE, it makes sense to take out the FreeBSD privdrop code and revert to the previous behaviour. I will integrate cap_sysctl(3) into #284 so once 12.1 is released, a capsicum sandboxed solo5 is ready for FreeBSD.

mato:
Ok, so, in the light of:

1. Discussion in #316, the real fix here is as described in https://github.com/Solo5/solo5/pull/316#issuecomment-464757104 (fixing the FreeBSD vmm APIs).
2. The build system refactoring in #326 changes the way hvt modules work, and the `HVT_DROP_PRIVILEGES=0` case will effectively be non-functional until I re-design that to not use compile-time flags. However, in the mean time, I need the tests to pass!

I'm going to make an executive decision here, which is: The FreeBSD privilege dropping introduced in #286 will be **reverted** on `master` shortly. 

Also, I'm going to close all issues/PRs related to this except #282, since the multiple discussions are confusing. We can discuss how to proceed forward there.

Note to self: **TODO**: The _existing_ VM cleanup code for FreeBSD and OpenBSD should be audited and at  `warn()` should be used in the atexit handler(s) if any syscalls fail.
/cc @sg2342 @hannesm

Issue #321, opened 2019-02-14
[spt] Need an executable heap for node


ricarkol:
Nodejs (rumprun) needs an executable heap for its JIT. The way around that was by cheating a bit and use "PROT_READ implies PROT_EXEC". But that's forbidden in spt (which is the right thing to do), so really don't know what to do. One option would be to add something like an optional EXTRA_MMAP_FLAG to solo5 to be used with care and a big warning.

mato:
Right, since you need to run the node JIT, you need to get executable pages from somewhere. And you are correct, in the Solo5 design there is no way to get "more" executable pages after loading the unikernel, this is deliberate.

So, what to do? I can see two options:

1. A hack, which would be to provide a run-time option to the spt tender (and eventually maybe also hvt for consistency) to add `PROT_EXEC` to the `mmap()` call mapping guest memory. Unfortunately, this also means you will get an **executable stack**, as the stack is not separate from the heap. So you will lose quite a bit of your security posture.

2. Have you considered AOT compiling all your node.js code instead? There are projects around, such as https://github.com/lastmjs/node-compiler that claim to be able to compile any node.js application down to a single ELF executable. Obviously you'd have to port that to NetBSD/rumprun, but it is by far the better approach to running node.js code in a Solo5 type environment. (/cc @djwillia!)

djwillia:
I don't know the answer to the question of what to do about JIT and it seems like it's a complicated thing.  It would be nice to be able to push everything to compile time, but I don't think we can do that for some use-cases that are otherwise a good fit for Solo5.  I would be in favor of doing the hack suggested in number 1 with a warning so that the rump-on-solo5 stuff (used by nabla containers) doesn't have to fork Solo5.

It sounds like an interesting longer-term project to think a bit more about in what cases you can get away without a JIT for things like node.JS...

mato:
> It would be nice to be able to push everything to compile time, but I don't think we can do that for some use-cases that are otherwise a good fit for Solo5. 

Why not?

> I would be in favor of doing the hack suggested in number 1 with a warning so that the rump-on-solo5 stuff (used by nabla containers) doesn't have to fork Solo5.

In that case, please submit a PR for `spt` which adds a run-time option with a suitably dangerous name e.g. `--x-exec-heap` and prints the warning `Running with an executable guest heap and stack, security posture will be reduced.` at run time.

ricarkol:
> Why not?
> 
> >
[...]

PR #324, opened 2019-02-19
spt: Add the --x-exec-heap argument

ricarkol:
More at #321 

Can add tests and an hvt implementation if needed.


mato:
Looks reasonable, notwithstanding my reservations in #321. Don't bother with hvt unless you're actually going to use it.

ricarkol:
Thanks!

PR #339, opened 2019-03-22
For discussion only: Expose solo5_mem_alloc and solo5_mem_lock

minad:
This would open the door for multiple separate randomized stack and heap memory regions. Given separate regions the stack would not be able to overwrite parts of the heap.

Downsides:
* More API and exposed complexity
* Requires the application to lock the memory layout after an initialization phase
* The initial stack size is fixed, however applications could also allocate larger memory regions and use those as stack.

Mailing list thread: https://www.mail-archive.com/solo5@lists.h3q.com/msg00064.html. Please discuss there, not here, to avoid splitting the discussion.

### DIFF
--- a/bindings/bindings.h
+++ b/bindings/bindings.h
@@ -1,4 +1,4 @@
-/* 
+/*
  * Copyright (c) 2015-2018 Contributors as noted in the AUTHORS file
  *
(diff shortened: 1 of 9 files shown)


minad:
Close this since it doesn't seem compatible with the security model where every is statically preallocated.

PR #342, opened 2019-03-26
 OpenBSD: add -mno-retpoline to HOST_CFLAGS to allow MirageOS Unikernels to run

### LINKED ISSUE #336: OpenBSD Current: MirageOS Unikernels fail to run on OpenBSD Current (retpoline)
retpoline was [enabled by default](https://marc.info/?l=openbsd-cvs&m=154621129329314&w=2) at the end of December last year.

The fix for now is to add ` -mno-retpoline` to `HOST_CFLAGS` see [try without retpoline]( https://github.com/Solo5/solo5/commit/01749ac6f95bcc953b2277a17e651268f4236a9c)

I will open a PR when https://github.com/Solo5/solo5/issues/326 has been completed.

adamsteen:
closes #336

retpoline was [enabled by default](https://marc.info/?l=openbsd-cvs&m=154621129329314&w=2) at the end of December last year.

The fix for now is to add ` -mno-retpoline` to `HOST_CFLAGS`

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -203,7 +203,7 @@ case $(uname -s) in
         # we don't support yet. Unfortunately LLVM does not support
         # -mstack-protector-guard, so disable SSP on OpenBSD for the time
         # being.
-        HOST_CFLAGS="-fno-stack-protector -nostdlibinc"
+        HOST_CFLAGS="-fno-stack-protector -mno-retpoline -nostdlibinc"
         warn "Stack protector (SSP) disabled on OpenBSD due to toolchain issues"
         HOST_LDFLAGS="-nopie"
         BUILD_HVT="yes"


mato:
I've integrated this change into #345, so closing this one.

minad:
What is the issue with retpolines and stack guards on openbsd? Wouldn't it be nice if those would work?

adamsteen:
@minad It would be fantastic if those worked, that is the end goal, but having MirageOS/solo5 working now is more important.

* SSP on OpenBSD has a different ABI to everyone else, so until someone looks at this, its turned off (I do have a working SSP branch but am not sure how to integrate it into the mainline)
* retpoline I am less sure on, I just haven't had the time to look further, I think the problem occurs when the ocaml runtime is launch, but have not investigate, besides a quick read of the source code on the bus.

Please feel free to create a PR to fix one or both of these issues, all help is appreciated.

Issue #326, opened 2019-02-20
Remove specialization of hvt tender at unikernel compile time


mato:
A long-standing design [issue](https://www.mail-archive.com/solo5@lists.h3q.com/msg00024.html) in Solo5 is the specialization of the hvt tender being done at unikernel compile time, which is a hold-over from the days of ukvm.

With the introduction of the spt tender the problem has become even more pronounced, so the time has come to remove this concept from the codebase.

Among other things (see linked mailing list post for more), this change is blocking:

- A sane supply chain, allowing the _operator_ (not the _developer_) to supply the solo5-hvt tender. This is a pre-requisite for a better deployment story, e.g. integration with libvirt-based systems for deployment.
- Sharing code between the hvt and spt tenders (e.g. ELF loader, common host device/module setup code).
- Support for multiple instances of block/network devices, and support for "future" devices (e.g. host-only communication channels).

The following sub-tasks are part of this work:

1. [x] Build system re-factoring (see comment below) and removing hvt compile-time specialization with minimal changes to the current codebase. (#345)
2. [x] Share common code between tenders. (#364)
Details:

1. As a start, we will build ~~one `solo5-hvt`~~ two tenders, `solo5-hvt` with all non-debug modules and `solo5-hvt-debug` including debug modules. the tender will allow any combination of devices to be enabled at run time, i.e. post this step it will no longer _enforce any contract_ between the unikernel and its host on what resources (network, block devices) are required by the unikernel to run.
2. Tenders need to be able to share common code, as the upcoming support for multiple devices will introduce more common functionality.

After this is complete I will raise a separate issue to deal with the actual implementation of multiple device support and adding back the _enforcement_ of the host/unikernel contract by the tender as the two are ultimately intertwined.

mato:
Progress update: Working on this has led to an overdue refactoring of the build system, with the following overall goals:

1. Correct dependency handing. i.e. `make` must always rebuild what has changed, with a minimal amount of dependencies specified manually in Makefiles.
2. Clear data flow throughout the build system. The build system needs to produce four groups of artifacts (bindings for _A_ targets, _B_ tenders for the host, _C_ tests for _A_ targets and _D_ .pc files for use by OPAM), using different compilers (or flags). As things currently stand, the number of combinations leads to unmaintainable spaghetti code.
[...]

Issue #292, opened 2018-11-22
Remove -m-no-red-zone on AMD64


mato:
We have been carrying `-m-no-red-zone` for AMD64 for all code (bindings themselves and "user" code) since the dawn of time. Since then, the CPU interrupt handling and trap handling code has been refactored to use ISTn, i.e. separate stacks for interrupt and trap handlers.

We should consider removing `-m-no-red-zone`. There are two options:

1. Remove it from "user" code only, i.e. for the MirageOS case, do not pass it down to OPAM via pkg-config.
2. Remove it from the bindings as well.

The first option seems entirely safe -- "user" code will never run in an interrupt context, and any trap/interrupt handlers in bindings currently run on a separate stack via ISTn.

The second option should be possible with a minor tweak to our usage of ISTn to accomodate virtio (hvt does not use hardware interrupts). Currently, we have (from bindings/cpu_x86_64.c):

````
IST1 (cpu_intr_stack): All hardware interrupts except #NMI (virtio only).
IST2 (cpu_trap_stack): All traps except #DF.
IST3 (cpu_nmi_stack): #NMI, #DF.
````

Given that on virtio it is theoretically possible for IRQ0 (timer) to nest with IRQx (virtio_net), I think we want something like this:

````
IST1 (cpu_intr1_stack): IRQ0 (timer, virtio only)
IST2 (cpu_intr2_stack): All other hardware interrupts (i.e. only virtio-net).
IST3 ... IST5 (unused, reserved for future expansion)
IST6 (cpu_trap_stack): All traps except #DF.
IST7 (cpu_nmi_stack): #NMI, #DF.
````

References: [System V Application Binary Interface x86-64 Architecture Processor Supplement, section 3.2.2](http://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf), also [this](https://stackoverflow.com/a/42585495) stackoverflow answer.

Any opinions? @djwillia @ricarkol?

djwillia:
Interesting, the red zone business seems very subtle, which is why I guess most kernels just avoid it.  But you are right, we certainly don't need it in the hvt case.  I would lean towards removing it entirely (option 2 above).  It seems the only issue is the virtio case.

[...]

PR #345, opened 2019-03-27
Refactor build system, remove hvt compile-time specialization

mato:
This is step 1 of #326. Note that as I'll be offline for about two weeks starting this Friday, I am going to merge this at the latest by tomorrow morning. As far as I know there are no major showstoppers and the shortcomings can be fixed later.

/cc @ricarkol @djwillia @hannesm (FreeBSD) @adamsteen (OpenBSD) @Kensan (Muen) @ehmry (Genode).

Note that I have not tested Muen or Genode beyond build-testing. I have also not tested the result of this with downstream Mirage/Solo5 on anything other than Debian (which is also being done by the E2E CI now).

@ehmry There was some confusion in "which set of CFLAGS / LD / LDFLAGS" was used for Genode in the old build system, I think I've picked the right ones in the new version but you might want to double-check.

----

This is a large change, and the majority of it is effectively a full
re-write of the build system. The actual removal of hvt compile-time
specialization is fairly straightforward.

User-visible changes:

- 'configure.sh' now needs to be run manually before running 'make',
  this is consistent with POLA.
- conversely, 'make clean' no longer cleans Makeconf. Use 'distclean' or
  'clobber' for that.
- 'configure.sh' will now print the targets that can (will) be built on
  this system. The strategy is still "build everything we can", however
  I have disabled Genode on all systems except Linux due to toolchain
  issues.
- You can now build a subset of targets from the top-level 'make', by
  specifying 'CONFIG_XXX=' (disable) or 'CONFIG_XXX=1' (enable) either
  on the command line, or editing the generated Makeconf.
- Makefiles use silent rules by default. To get the old verbose ones
  back, use 'make V=1'.
- The 'solo5-hvt' tender is no longer "specialized" to the unikernel.
  We build two tenders, 'solo5-hvt' with all non-debug modules
  configured and 'solo5-hvt-debug' with additional debug modules (gdb,
  dumpcore where available).
- 'solo5-hvt-configure' is kept around for now for backward
  compatibility with OPAM/MirageOS but is essentially a NOP.

Developer-visible changes:

- The build system now has proper support for auto-generation of
  dependencies. This means you can safely edit source files, run make
  and be sure you will get a complete incremental build.
- Makefiles have been refactored to use common best practices, remove
  repetition, consistent variable names and clear interfaces between
  configure.sh/Makeconf/Makefiles, all the while keeping them simple
  enough to understand for me on a Monday morning before coffee. I.e.
  limit use of macros, eval, etc.
- hvt tender modules are no longer defined by compile-time flags,
  instead a dynamic array is placed into a special ELF section
  (.modules).  This means that a hvt tender binary can be combined from
  an arbitrary set of hvt_module_XXX object files, which is the right
  way to do things going forward and also simplifies the build system
  (not needing to build multiple targets from the same set of sources).

Shortcomings / TODOs:

- Dependency files (*.d) are stored in-tree. I spent several days on
  trying to figure out how to get them to work out of tree, but in
  combination with the non-recursive use of subdirectories in 'bindings'
  I could not figure out the required Makefile magic.
- HVT_DROP_PRIVILEGES=0 is non-functional with the new modules
  arrangement, but needs a re-design anyway.

Other changes included as part of this PR:

- Revert privilege dropping on FreeBSD (see discussion in #282).
- The build system changes effectively implement option 1 in #292, i.e.
  on x86_64 -m no-red-zone is only used for bindings, not for
  application code.
- tests/tests.bats has been refactored for DRY as it was getting totally
  unmaintainable.


mato:
Note that the failures for `13-basic-x86_64-Debian10` and `14-basic-x86_64-OpenBSD64` are in the tests, I am not going to attempt to fix those at this time. Will open separate issues for those once merged.

mato:
(This PR is intended mainly as a heads up and to document the changes made, I'm not actually asking people to review the build system changes, unless you feel particularly masochistic...)

tenders/hvt/hvt.h
```diff
@@ -160,23 +160,26 @@ extern hvt_vmexit_fn_t hvt_core_vmexits[];
...
     int (*handle_cmdarg)(char *cmdarg);
     char *(*usage)(void);
 };
 
-/*
- * Array of compiled-in modules. NULL terminated.
- */
-extern struct hvt_module *hvt_core_modules[];
+struct hvt_module {
+    const char *name;
```
ricarkol: If you use `const char name[32];` (the compiler will complain if 32 is too small) then the modules section can contain the name like below:

 ```
[kollerr@oc6638465227 solo5]$ readelf -x modules tenders/hvt/solo5-hvt
Hex dump of section 'modules':
  0x00606180 636f7265 00000000 00000000 00000000 core............
  0x00606190 90184000 00000000 00000000 00000000 ..@.............
  0x006061a0 00000000 00000000 00000000 00000000 ................
  0x006061b0 00000000 00000000 00000000 00000000 ................
  0x006061c0 626c6b00 00000000 00000000 00000000 blk.............
  0x006061d0 202e4000 00000000 e02d4000 00000000  .@......-@.....
  0x006061e0 d02d4000 00000000 00000000 00000000 .-@.............
  0x006061f0 00000000 00000000 00000000 00000000 ................
  0x00606200 6e657400 00000000 00000000 00000000 net.............
  0x00606210 40344000 00000000 20384000 00000000 @4@..... 8@.....
  0x00606220 30324000 00000000                   02@.....
```

File: configure.sh [hunk 2]
```diff
@@ -67,18 +79,25 @@ ld_is_lld()
...
-    x86_64-*|amd64-*)
-        TARGET_ARCH=x86_64
+# Determine HOST and ARCH based on what the toolchain reports.
+case ${CC_MACHINE} in
+    x86_64-linux*)
+        CONFIG_ARCH=x86_64 CONFIG_HOST=Linux
         ;;
-    aarch64-*)
-        TARGET_ARCH=aarch64
+    aarch64-linux*)
```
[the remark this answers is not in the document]
mato: Fixed, thanks!

tenders/hvt/hvt.h
mato: Uh, yeah, the `.modules` stuff is a bit raw. I'm not going to change it now, will revisit when I get back.

PR #351, opened 2019-04-07
SSP for OpenBSD, generic naming of symbols

adamsteen:
OpenBSD use different symbols for __stack_chk_*

See the following mappings
__stack_chk_guard => __guard_local
__stack_chk_fail => __stack_smash_handler

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -202,7 +202,7 @@ case "${CONFIG_HOST}" in
         # clang-provided headers for compiler instrinsics. We copy the rest
         # (std*.h, cdefs.h and their dependencies) from the host.
         cc_is_clang || die "Only 'clang' is supported on OpenBSD"
-	[ "${CONFIG_ARCH}" = "x86_64" ] ||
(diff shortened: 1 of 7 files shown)


configure.sh
```diff
@@ -276,6 +277,8 @@ CONFIG_HOST=${CONFIG_HOST}
 MAKECONF_CC=${CC}
 MAKECONF_LD=${LD}
 CONFIG_SPT_NO_PIE=${CONFIG_SPT_NO_PIE}
+SSP_GUARD=${SSP_GUARD}
+SSP_FAIL=${SSP_FAIL}
```
mato: Rather than passing `SSP_GUARD` and `SSP_FAIL` all the way up from `configure.sh` into the bindings Makefile, could you instead please use an `ifeq ($(CONFIG_HOST), OpenBSD) ... else ... endif` block in `bindings/GNUmakefile`, just before the `define LINK.bindings`?

adamsteen:
@mato fixed, rebased and squashed!

mato:
Looks good, thanks!

Issue #343, opened 2019-03-26
API redesign


minad:
@mato I am also leaving for vacation and I wanted to finish a prototype before. This is a proposal for an API redesign, which should make it possible to solve some of the outstanding feature requests. Things should probably go step by step and the discussion should be on the mailing, I know ;)
But I wanted to back my uneducated ramblings on the mailing list with some code. We can continue the discussion on the mailing list as soon as you are back and the more important issues are sorted out.

My prototype works only on Linux x86_64 with seccomp (for now). I can also port it to aarch64 if desired. It is similar to spt, however it is "tenderless" (#341) and contains all the startup code directly in the binary. There is no libc and no libseccomp. This works only for spt obviously, but in that case I actually prefer that approach in contrast to having the separate loader. (**Edit**: If you want to execute untrusted spt images, one still needs the spt tender however).

So what does it have?

* Block operations for more than one block #325. The checking is done by hand written BPF asm.
* Multiple block devices, network devices and stream devices (all statically allocated in the beginning). For simplicity there is a fixed maximal number but this could be increased easily.
* Adds discard operation #333 
* Adds sync operation #330 
* yield returns bitmask for the network devices
* No calls for block_info and net_info, everything is available at startup
* Separate stack (by just mmaping the separate heap)
* No error result type. Maybe not a good idea. For example the block calls could fail if the disk is full, but I don't know what the app should do in that case instead of termination if there is no possibility to save state.

Furthermore there are other bikesheds, which you can ignore:

* argv/argc instead of cmdline. **EDIT**: I am zeroing out the environment and argv and auxv to avoid leaking information.
* startup info is globally available, the pointer could also be passed to main
* yield gets nanoseconds instead of deadline, "decouples" yield and monotonic calls
* removed the abort call since it seems useless (?) and it just calls exit

The code is here https://gist.github.com/minad/c6f44a245f1ea481ded2535a96778be5

minad:
Hi @hannesm, just saw your mail after creating this issue. I will answer to it when I am back. Just letting you know I got your mail :)

minad:
I updated my prototype to support input stream devices. IO streams can be used for example with named FIFOs. Output streams can be simple files. This extensions leads to a more complicated `yield` call supporting read/write events on the streams.

Issue #325, opened 2019-02-19
Enable block operations for more than 1 block


ricarkol:
This discussion started in PR #315 

The issue is that the solo5 interface is limited to a block at a time and the block is too small. The issue is that it's inefficient as it involves too many exits for hvt, or too many syscalls for spt. The following experiment quantifies that. This is the time in seconds to read a 1GB file sequentially for an increasing block size on both hvt and spt (*):

```
block		hvt		spt
512		5.690s		0.463s
1024		2.881s		0.473s
4096		0.776s		0.171s
8192		0.434s		0.126s
16384		0.259s		0.103s
```

spt is already pretty fast at 512, but it can be 4x faster by increasing the block size to 8192. This experiment's point is not to increase the block size necessarily, but to allow for multiblock requests instead.

(*) details of the experiment. This is the unikernel: https://gist.github.com/ricarkol/b66c899edd96fd7f8fb2fbaeabad0694#file-solo5-blksize-test-c. This is how the blk size was changed: https://gist.github.com/ricarkol/b66c899edd96fd7f8fb2fbaeabad0694#file-blk-size-diff. Used an Ubuntu 18.04.1 running on an Intel(R) Xeon(R) CPU E3-1270 v6 @ 3.80GHz. The disk was stored as a file on an SSD formatted with ext4 (caches were not dropped before each experiment).

mato:
> The issue is that the solo5 interface is limited to a block at a time and the block is too small. And the problem with that is that it's inefficient as it usually involves too many exits for hvt, or too many syscalls for spt. The following experiment quantifies that. [...]

The following comments from #315 are also relevant to this discussion:

* @ricarkol https://github.com/Solo5/solo5/pull/315#issuecomment-457950289
* @cfcs https://github.com/Solo5/solo5/pull/315#issuecomment-458069807

Regarding atomicity guarantees and the "ideal block size" at the solo5_block API level: I'd have to study the problem in depth, so I can't give any informed opinion right now.

So, concentrating on the core of the issue which could be immediately addressed, which is **allowing block I/O in multiples of the block size per request**: In order to do that, the following needs to be done:

1. `hvt`, `spt`: If the solo5 virtual block device is backed by a _file_ on the host, we need to ensure that the guest cannot write beyond the end of the file. I.e. in any `pwrite(fd, buf, count, offset)` ensure that `(offset + count) <= capacity` holds, accounting for overflow(!).
[...]

PR #352, opened 2019-04-10
Bpf seccomp

minad:
Taken from #343. The bpf assembly is not yet adapted for solo5 yet, since it supports multiple block devices etc.

This removes the dependency on libseccomp and it lifts the restriction of single block writes. Blocks must still be aligned however (related #325).


minad:
Ping @mato. Is there interest in this?

mato:
@minad Yes, I think this is worth doing, but needs some more work.

One thing we lose with this approach is that currently the seccomp filter is built up by calls in each module that gets activated, so the filter is effectively "tailored" to the modules in use. This is useful for future modules that might want to punch different (but again, optional) holes in the seccomp filter, and also follows the principle of least privilege.

Further, I'm not sure how the "one big filter" approach would work in the presence of multiple instances of modules (devices/resources). This I think we won't know for certain until I have a prototype in place.

minad:
> One thing we lose with this approach is that currently the seccomp filter is built up by calls in each module that gets activated, so the filter is effectively "tailored" to the modules in use. This is useful for future modules that might want to punch different (but again, optional) holes in the seccomp filter, and also follows the principle of least privilege.

This is right. But I don't think there is a solution which is as simple as this one if you want to build the seccomp filter step by step from each module. I don't see the seccomp filter as part of the device modules, but as something sitting below, like a firewall. Therefore I think it is also justified from a design point of view, to do it in a monolithic step. The filter builder should just iterate over all the active modules.

> Further, I'm not sure how the "one big filter" approach would work in the presence of multiple instances of modules (devices/resources). This I think we won't know for certain until I have a prototype in place.

What do you mean by that? The filter I have written already supports multiple devices. There is support for up to 8 block devices, multiple IO stream devices (like serial ports) and multiple network devices. The filter blob is parametrized by the number of devices before loading. It is hard to make that more flexible than that. The limitation in the number of devices comes from the encoding of lookup tables in BPF, which is very limited. I made a testbed/prototype to test some things concerning multiple devices and the filter script proposed in this PR is just taken from there. For solo5 it would definitely need some tweaking.

Alternatively you could look for a full blown BPF compiler solution, which allows you to build the filter up step by step. But this would involve much complexity. Basically libseccomp is such a thing, but it seems insufficient.

For the scope of a highly restricted sandbox like solo5 I would rather prefer I very short hand written filter script, written in BPF ASM directly, which can be checked carefully. Also note that eBPF is not yet available for seccomp and I have no idea when it will be. There are at least some kernel patches floating around afaik.

mato:
@minad:

I've thought about your approach [of a single monolithic BPF] some more, especially in the context of the evolving support for multiple devices (#372 / #373).

> What do you mean by that? The filter I have written already supports multiple devices. There is support for up to 8 block devices, multiple IO stream devices (like serial ports) and multiple network devices. The filter blob is parametrized by the number of devices before loading. It is hard to make it more flexible than that.

The thing is, in the presence of multiple devices, the monolithic approach generates a weaker filter than the per-module gradual approach we have now.

For example, in your BPF in `block_call:`, you're making the assumption that the file descriptors on which `pread` / `pwrite` may be called are part of the contiguous range `$BPF_BLOCK_MIN ... $BPF_BLOCK_MAX`, and said range contains _no other file descriptors_. There's no way to guarantee that, as the devices get attached in whichever order the user may have specified them on the tender's command line. Contrast this to the libseccomp-based approach which adds a rule _specific_ to each file descriptor.

Also, if we want to keep at least some checks to disallow/limit the impact of block writes past the end of the backing file (to enable #325), those checks need to be parametrized _per_ file descriptor as each may have a different device capacity (i.e. limit for the `pos` argument to the `pwrite` call). I don't see how to achieve that without generating at least some of the filter at tender run-time.

> For the scope of a highly restricted sandbox like solo5 I would rather prefer I very short hand written filter script, written in BPF ASM directly, which can be checked carefully. 

YMWV, but my opinion on this as maintainer is that the one-monolithic-filter approach will have a much higher maintenance cost further down the line. Any change to a module which might affect the filter requires careful coordination between all modules, and patching the monolith itself. With the libseccomp-based filter this is much reduced, as each module is responsible for it's "piece" of the filter.

Personally, I find the libseccomp-based rules easier to follow from the tender's perspective than hand-written BPF. Granted, we are trusting libseccomp to generate correct rules, which may not always be the case (see http://www.paul-moore.com/blog/d/2019/03/libseccomp_v240.html), but that is a trade-off I'm willing to keep until we have an option which is convincingly better.

One such option worth investigating is a compromise, where, given the relatively small set of seccomp rules we need, the fact that we always operate as a strict whitelist and the fact that the parametrization needed is straightforward (basically only changing literals, not control flow), we use a combination of 

a) modules being able to define (parametrized) BPF "snippets", each of which is self-contained and must either `ret ALLOW` or continue to the next instruction.
b) the core then combining all snippets into a single BPF filter, with the arch-specific header and a `ret KILL` at the end.

...however, I'm not sure if implementing the above wouldn't just lead to a half-assed implementation of what libseccomp already does for us today.

minad:
Ok, I guess this can be closed then.

PR #353, opened 2019-04-24
tests: Increase timeout, virtio start delay

### LINKED ISSUE #346: virtio test_ping_serve flaky/failing on Debian 10 / QEMU 3.1
As can be seen in e.g. https://gist.github.com/Solo5-CI/b470d17f6ee7ca3f6f5b0d70c50f5e59 and other builds since the introduction of the `Debian10` builder, test_ping serve on virtio is failing due to:

```
# PING 10.0.0.2 (10.0.0.2) 56(84) bytes of data.
# 
# --- 10.0.0.2 ping statistics ---
# 100000 packets transmitted, 99955 received, 0.045% packet loss, time 408ms
# rtt min/avg/max/mdev = 0.213/0.216/4.081/0.018 ms, ipg/ewma 0.233/0.216 ms
```

This might be due to several reasons:

1. A bug in our virtio code, exposed by the update to QEMU 3.1 in Debian 10.
2. A bug in QEMU 3.1 virtio-net (unlikely).
3. Scheduling pressure / startup time / other transient issue, i.e. things are not set up quickly enough.

I've no time to investigate this right now, but given the general flakiness of test_ping_serve, which relies on an inherently _unreliable_ transport (ICMP), apart from verifying that it's not a bug on our end an entirely acceptable fix would be to add tolerance for say, up to 0.5% packet loss using bash arithmetic on `XXXXXX` received from the `ping` output.

mato:
- increase test timeout to 60 seconds as test_ping_serve may take >30s on the test infrastructure.
- increase delay before pinging virtio variant of test_ping_serve as qemu 3.x takes ages to start up.

Fixes #346 and also the OpenBSD test_ping_serve failures (/cc @adamsteen).

### DIFF
--- a/tests/tests.bats
+++ b/tests/tests.bats
@@ -75,15 +75,15 @@ teardown() {
 }
 
 hvt_run() {
-  run ${TIMEOUT} --foreground 30s ${HVT_TENDER} "$@"
+  run ${TIMEOUT} --foreground 60s ${HVT_TENDER} "$@"
 }
 
 spt_run() {
-  run ${TIMEOUT} --foreground 30s ${SPT_TENDER} "$@"
+  run ${TIMEOUT} --foreground 60s ${SPT_TENDER} "$@"
 }
 
 virtio_run() {
-  run ${TIMEOUT} --foreground 30s ${VIRTIO} "$@"
+  run ${TIMEOUT} --foreground 60s ${VIRTIO} "$@"
 }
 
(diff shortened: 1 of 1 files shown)


adamsteen:
@mato just wanted to note here, when I run the tests manually, I have never had a problem with a timeout, but since it’s a timeout, it shouldn’t be a problem.

PR #354, opened 2019-04-25
OpenBSD atexit cleanup - add warnings

adamsteen:
Warn if we are unable to close the vmm file descriptor or ioctl fails to terminate the vm(unikernel), but still exit.


mato:
Thanks!

PR #355, opened 2019-04-25
spt/hvt: limit default memory to 32MB (instead of 512MB)

hannesm:
in my opinion this is more reasonable and allows me to not have to pass `--mem=xx` during development. changing a default value is always a breaking change, but I think it is worth it.

I run several solo5-based unikernels with memory ranging between 16 and 128MB in production -- a more conservative would be to use 64 or 128MB memory...

### DIFF
--- a/tenders/hvt/hvt_main.c
+++ b/tenders/hvt/hvt_main.c
@@ -107,7 +107,7 @@ static void usage(const char *prog)
     fprintf(stderr, "KERNEL is the filename of the unikernel to run.\n");
     fprintf(stderr, "ARGS are optional arguments passed to the unikernel.\n");
     fprintf(stderr, "Core options:\n");
-    fprintf(stderr, "  [ --mem=512 ] (guest memory in MB)\n");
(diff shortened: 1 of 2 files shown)


mato:
> in my opinion this is more reasonable and allows me to not have to pass --mem=xx during development.

Why do you care about the default during _development_? It's just virtual memory, the actual RSS depends on how much the guest uses. During development I'd prefer to err on the side of random "MirageOS + Irmin + whatever" unikernel will run without hitting memory limits, otherwise I can see people complaining "it doesn't work".

> I run several solo5-based unikernels with memory ranging between 16 and 128MB in production -- a more conservative would be to use 64 or 128MB memory...

It makes sense in production, but IMO in a production setting you should be doing proper resource allocation in your orchestration layer based on your operational preferences, and thus always passing a `--mem` to the tender anyway.

hannesm:
ok, yes in production I already always pass through `--mem`. for development and demonstration purposes - when I intend to show "small memory usage", I need to pass `--mem=32` manually... but I see your point.

PR #366, opened 2019-05-28
capsicum(4) sandbox solo5 on FreeBSD

sg2342:
- cap_sysctl(3) provides access to the sysctl namespace (restricted
  to hw.vmm.destroy)
- hvt_module_dumpcore was changed to use openat(2} and file creation
  restricted to the current working directory
- block, net, stdio file descriptors and ioctls have been restricted
  as in bhyve(8)

the ppoll(2) capsicum whitelist change was included in the 12-STABLE
branch and will be part of the FreeBSD-12.1 release (scheduled for
Nov 2019)

### DIFF
--- a/tenders/hvt/hvt_freebsd.c
+++ b/tenders/hvt/hvt_freebsd.c
@@ -82,7 +82,6 @@ static void cleanup_vm(void)
     }
 }
 
-
 static void cleanup_vmfd(void)
 {
     if (cleanup_hvt != NULL && cleanup_hvt->b->vmfd != -1)
--- a/tenders/hvt/hvt_freebsd.c
+++ b/tenders/hvt/hvt_freebsd.c
@@ -25,9 +25,7 @@
 
 #include <assert.h>
 #include <err.h>
-#include <errno.h>
 #include <fcntl.h>
-#include <pwd.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
--- a/tenders/hvt/hvt_freebsd.c
+++ b/tenders/hvt/hvt_freebsd.c
@@ -44,6 +44,16 @@
 #include <sys/cpuset.h>
 #include <machine/vmm_dev.h>
 
+#include <osreldate.h>
+
+#if HVT_FREEBSD_ENABLE_CAPSICUM
+#include <sys/capsicum.h>
+#include <sys/nv.h>
+#include <libcasper.h>
+#include <casper/cap_sysctl.h>
+#include <capsicum_helpers.h>
+#endif
+
 #include "hvt.h"
 #include "hvt_freebsd.h"
 
@@ -56,12 +66,24 @@
  */
 static struct hvt *cleanup_hvt;
 
+#if HVT_FREEBSD_ENABLE_CAPSICUM
+static cap_channel_t *capsysctl;
+#endif
+
 static void cleanup_vm(void)
 {
-    if (cleanup_hvt != NULL)
-        sysctlbyname("hw.vmm.destroy", NULL, NULL, cleanup_hvt->b->vmname,
+    if (cleanup_hvt != NULL) {
+#if HVT_FREEBSD_ENABLE_CAPSICUM
+       cap_sysctlbyname(capsysctl, "hw.vmm.destroy", NULL, NULL,
+                cleanup_hvt->b->vmname, strlen(cleanup_hvt->b->vmname));
+       cap_close(capsysctl);
+#else
+       sysctlbyname("hw.vmm.destroy", NULL, NULL, cleanup_hvt->b->vmname,
                 strlen(cleanup_hvt->b->vmname));
+#endif
+    }
 }
+
 
 static void cleanup_vmfd(void)
 {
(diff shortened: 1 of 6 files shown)


mato:
> hvt_module_dumpcore was changed to use openat(2} and file creation restricted to the current working directory

Using `openat(2)` would be useful for all hvt host systems. How about we make this behaviour explicit, by changing dumpcore to take an option of `--dumpcore=DIR` (and verifying that `DIR` exists and is writable...).

> block, net, stdio file descriptors and ioctls have been restricted as in bhyve(8)

Where are the stdio (well, stdout) fds restricted? I don't see that in the diff.

Also, can you please do the rights restrictions in the relevant module's `setup()` rather than in the common `tap_attach()` and `block_attach()`? Those functions should not have "extra" side effects.

> the ppoll(2) capsicum whitelist change was included in the 12-STABLE branch and will be part of the FreeBSD-12.1 release (scheduled for Nov 2019)

At the moment this does not compile on 11-STABLE, we should continue to support that. Also, if I understand you correctly, the new code will _require_ **12.1**-RELEASE or later to run. Therefore, can you gate the code on that suitably? If so, how? (i.e. I'd like to know how this would be done before you actually implement it)

sg2342:
> > hvt_module_dumpcore was changed to use openat(2} and file creation restricted to the current working directory
> 
> Using `openat(2)` would be useful for all hvt host systems. How about we make this behaviour explicit, by changing dumpcore to take an option of `--dumpcore=DIR` (and verifying that `DIR` exists and is writable...).

ok, will do.

> 
> > block, net, stdio file descriptors and ioctls have been restricted as in bhyve(8)
> 
> Where are the stdio (well, stdout) fds restricted? I don't see that in the diff.

calls to caph_limit_stdin(3) and caph_limit_stderr(3) in hvt_drop_privileges().
> 
> Also, can you please do the rights restrictions in the relevant module's `setup()` rather than in the common `tap_attach()` and `block_attach()`? Those functions should not have "extra" side effects.

ok.

> 
> > the ppoll(2) capsicum whitelist change was included in the 12-STABLE branch and will be part of the FreeBSD-12.1 release (scheduled for Nov 2019)
> 
> At the moment this does not compile on 11-STABLE, we should continue to support that. Also, if I understand you correctly, the new code will _require_ **12.1**-RELEASE or later to run. Therefore, can you gate the code on that suitably? If so, how? (i.e. I'd like to know how this would be done before you actually implement it)

my preferred solution would be:

```
#if __FreeBSD_version >= 1200503
/* capsicum enabled code */
#else 
/* previous implementation */
#endif  
```
in the hvt code (and only add -lnv to  HOSTLDLIBS if uname -K is not less than 1200503)

mato:
> calls to caph_limit_stdin(3) and caph_limit_stderr(3) in hvt_drop_privileges().

Ah, ok, I missed that. Those should probably get done in core module's `setup()`, see below.

I'm not familiar with the capsicum APIs, have read through the man pages a bit, it's not obvious to me which APIs do what and how they inter-relate. So... here's what my intent with moving the bits of code around is -- in high level terms, if possible:

1. The various modules modules build up their respective parts of the sandbox in `setup()`. This includes the core module.
2. Likewise, the actual VMM interface code does so in a single place. `hvt_init()` where you're currently doing some of this seems like the right place.
3. Then, `hvt_drop_privileges()` should be the function that actually applies the sandbox.

Does that make sense? Right now `hvt_drop_privileges()` seems to be doing bits of (1) and (2), is that intentional?

> my preferred solution would be: [...gating on __FreeBSD_version, uname, etc]

Hmm. I'd really like to minimise `#ifdefs`, with the exception of separating host-dependent implementations. Do we really need to `#ifdef` out the code on FreeBSD < 12.1-RELEASE? I.e. will the code not compile at all on 11.x?

sg2342:
> > calls to caph_limit_stdin(3) and caph_limit_stderr(3) in hvt_drop_privileges().
> 
> Ah, ok, I missed that. Those should probably get done in core module's `setup()`, see below.
> 
> I'm not familiar with the capsicum APIs, have read through the man pages a bit, it's not obvious to me which APIs do what and how they inter-relate. So... here's what my intent with moving the bits of code around is -- in high level terms, if possible:
> 
> 1. The various modules modules build up their respective parts of the sandbox in `setup()`. This includes the core module.
> 2. Likewise, the actual VMM interface code does so in a single place. `hvt_init()` where you're currently doing some of this seems like the right place.
> 3. Then, `hvt_drop_privileges()` should be the function that actually applies the sandbox.
> 
> Does that make sense? Right now `hvt_drop_privileges()` seems to be doing bits of (1) and (2), is that intentional?

i addressed this in d67a5705c22b23c6e780a6e2f9e576e66137a257 . It was not intentional (code evolved this way after i removed the chroot/droppriv code).

> 
> > my preferred solution would be: [...gating on __FreeBSD_version, uname, etc]
> 
> Hmm. I'd really like to minimise `#ifdefs`, with the exception of separating host-dependent implementations. Do we really need to `#ifdef` out the code on FreeBSD < 12.1-RELEASE? I.e. will the code not compile at all on 11.x?

I have to research the capsicum API situation on 11.x. At least for 12.0-RELEASE it is possible to find the OS version (using getosreldate(3)) and then only enter the sandbox if the OS version is new enough).

sg2342:
on 11.x 
- the capsicum_helpers(3) API is missing: easy to solve since caph_enter(3), caph_limit_stout(3) and caph_limit_stderr(3) are only convince wrappers.
- code needs to be linked against libcasper (in addition to libnv): OS version has to be detected in configure.sh
- the sysctl proxy is not there: i don't see an alternative to compile time OS version detection and preprocessor use here

mato:
Thanks for researching. In that case, here's what I would do:

1. Do all the checks to determine if we can use capsicum in configure.sh (based on the outputs of `uname` presumably), and put the result in a `CONFIG_HVT_FREEBSD_ENABLE_CAPSICUM`.
2. In the tender Makefile, apart from adding `-lnv` if this config item is defined, also add `-DHVT_FREEBSD_ENABLE_CAPSICUM=1`.
3. In the tender source code itself, gate the capsicum code on the presence of the above.

WDYT? This way it's clear where and how the decision gets made.

sg2342:
i put the dumpcore openat(2) changes in #371

sg2342:
latest version implements the suggested capsicum detection/gate. Capsicum init code was moved to setup() function of block and net modules. I also merged the dumpcore-openat changes.

sg2342:
did a test by merging current #372. hvt tests completed.  Since ppoll(2) was replaced by kqueue(2), CONFIG_HVT_FREEBSD_ENABLE_CAPSICUM can be set FreeBSD-12.0 and later (uname -K >= 1200086)

tenders/hvt/hvt_freebsd.c
```diff
@@ -44,6 +44,16 @@
 #include <sys/cpuset.h>
 #include <machine/vmm_dev.h>
 
+#include <osreldate.h>
```
hannesm: this include is no longer needed

hannesm:
I tested this PR, and it works fine on my FreeBSD-12 stable. thanks!

tenders/hvt/hvt_freebsd.c
sg2342: done

mato:
Ok to defer this for now ,until we have a release with the multiple devices and associated changes out? Coming early next month, fingers crossed.

mato:
@sg2342 Would you mind rebasing this on latest master, so that I can re-review and consider it for upcoming release(es)?

sg2342:
> @sg2342 Would you mind rebasing this on latest master, so that I can re-review and consider it for upcoming release(es)?

done

tenders/hvt/hvt_freebsd.c
```diff
@@ -129,15 +148,51 @@ struct hvt *hvt_init(size_t mem_size)
...
 
 #if HVT_DROP_PRIVILEGES
 void hvt_drop_privileges()
 {
-    /*
-     * This function intentionally left blank for now (see #282).
-     */
+#if HVT_FREEBSD_ENABLE_CAPSICUM
+    if (caph_enter() == -1)
+        err(1, "cap_enter() failed");
```
mato, line 194: Possibly clearer to say: `"Unable to enter sandbox: caph_enter() failed"`?

sg2342:
> You have a bunch of checks like:
> 
> ```c
> if (cap_/*<something>*/(/*...*/) == -1 && errno != ENOSYS) {
>    /* fail */
> }
> ```
> 
> When would this fail with `ENOSYS`? I'm not too happy about exceptions like this, since the code flow doesn't actually stop, just skips that particular call. That looks like a maintenance nightmare further down the line.
> 
> If the `ENOSYS` means "the running kernel was configured w/o Capsicum", perhaps test for that conclusively up-front in a single place, skip _all_ the Capsicum-related code and drop the `&& errno != ENOSYS` from all of it?

You are right about this. And the use of the `ENOSYS` pattern is also inconsistent. I will remove it, since run-time detection will just add more complexity for the very unusual use cases (FreeBSD 12 or later without `options CAPABILITY_MODE`). If someone really wants this, they can always patch the one line in configure.sh.

mato:
> If someone really wants this, they can always patch the one line in configure.sh.

In fact, they don't need to touch configure.sh. Editing the generated `Makeconf`, or running `echo CONFIG_HVT_FREEBSD_ENABLE_CAPSICUM= >> Makeconf` after configure.sh is sufficient to override the setting, I do that all the time.

PR #367, opened 2019-05-28
Port Solo5 SPT to PowerPC64

stefanberger:
This series of patches ports Solo5 SPT to PowerPC64 (ppc64).

The biggest difference between ppc64 and the other supported architectures (x86_64, aarch64) is the usage of thread-local storage (tls). tls basically has to always be available and contain valid pointers since practically any function can access it using register r13.


tests/test_notls/test_notls.c
```diff
@@ -37,6 +37,8 @@ int solo5_app_main(const struct solo5_start_info *si __attribute__((unused)))
                        "add x0, x0, #0x10; "
                        "ldr w1, [x0]"
                        : : : "x0", "w1");
+#elif defined(__powerpc64__)
+    __asm__ __volatile("ld 3,0(13)" : : : "r3", "r13");
```
stefanberger: This test will not be run, so probably just leaving this empty should be ok.

stefanberger:
Following a discussion with @ricarkol we found that the thread-local storage (TLS) accesses are due to accesses to the the stack_guard described in tcbhead_t [here](http://openpowerfoundation.org/wp-content/uploads/resources/leabi/content/dbdoclet.50655241_50444.html). This can be avoided by compiling the applications with `-mstack-protector-guard=global`. So far I have been able to avoid having to use 'special compile time options' by copying from the original TLS (the one the tender is using) into the one prepared by the tender for the app as well as defining the TLS structures in the test cases in a special way.

It's not clear to me under what circumstances all the other fields in tcbhead_t could be accessed if the mission was to now prevent the stack guard access using that compile time option. Maybe it would be safer to work with a TLS copied from the original TLS rather than having to choose 'special compile time options' to make it work, also for applications and libraries later on. So, as the code is now I would require that a TLS for ppc64 looks like this:

struct tls {
    unsigned char pad[0x1000];
    unsigned char thread_local_data_go_here[0];  // determine the size of this part here from the ELF
};

The insight is also that the tcbhead_t seems to not be aligned at pad[0] but further up so that its last byte is in pad[0x0fff] (it's right-aligned rather than left-aligned to the pads). The stack guard is accessed via -0x7010(r13) where r13 is the pointer to the tls but set to point 0x8000 beyond the base of the above tls structure. Pointing it that far beyond the base address was done to be able to easily address fields ranging from roughly -0x8000 (= base of tls) to +0x8000, so 64kb, restricted by addressing modes of the processor.

mato:
@stefanberger:

> Following a discussion with @ricarkol we found that the thread-local storage (TLS) accesses are due to accesses to the the stack_guard described in tcbhead_t here. This can be avoided by compiling the applications with -mstack-protector-guard=global.

The use of TLS for `__stack_chk_guard` is indeed the default behaviour on (mainly) Linux toolchains on our other architectures. This is why I made the decision to use `-mstack-protector-guard=global` for Solo5 (and thus any downstream libOS C code). So, compiling with these flags is the correct approach.

> The biggest difference between ppc64 and the other supported architectures (x86_64, aarch64) is the usage of thread-local storage (tls). tls basically has to always be available and contain valid pointers since practically any function can access it using register r13.

Given the above and your previous comment, idoes this still hold? If not, can we drop the changes to TLS initialization and associated bits in this PR?

stefanberger:
@mato I unfortunately do not know under what other circumstances the thread-local storage may be accessed. It looks like these fields here make up the tls:

```
typedef struct {
    /* Reservation for HWCAP data. */
        unsigned int hwcap2;
        unsigned int hwcap; /* not used in LE ABI */
 
    /* Indicate if HTM capable (ISA 2.07). */
        int tm_capable;
        int tm_pad;
 
    /* Reservation for dynamic system optimizer ABI. */
        uintptr_t dso_slot2;
        uintptr_t dso_slot1;
 
    /* Reservation for tar register (ISA 2.07). */
        uintptr_t tar_save;
 
    /* GCC split stack support. */
        void *__private_ss;
 
    /* Reservation for the event-based branching ABI. */
        uintptr_t ebb_handler;
        uintptr_t ebb_ctx_pointer;
        uintptr_t ebb_reserved1;
        uintptr_t ebb_reserved2;
        uintptr_t pointer_guard;
 
    /* Reservation for stack guard */
        uintptr_t stack_guard;
 
    /* DTV pointer */
        dtv_t *dtv;
    } tcbhead_t;
```

We know how to eliminate access to `pointer_guard` using compiler options. I am not sure about the other fields, though.

mato:
There are two separate issues here, which I think are a source of confusion:

1. The use of TLS for SSP in _default_ Linux toolchains. This is accounted for by `-mstack-protector-guard=global` as you found yourself.
2. Whether or not a libOS / application / unikernel / the thing consuming the Solo5 APIs, toolchain and "runtime environment" uses TLS.

I think you're asking about (2) -- that is not Solo5's responsibility. All Solo5 provides is the basic primitive required to set the TLS register at runtime (`solo5_set_tls_base()`). The actual TCB and TLS setup should be done by the libOS itself, not by Solo5.

stefanberger:
I now opened PR https://github.com/Solo5/solo5/pull/368, which eliminates access to the stack_guard from the TLS. If this is the right approach, then I guess those two patches there are the right ones to merge and we can close this PR.

stefanberger:
Here's a bit more explanation from glibc about the ppc64 TLS:

```
/* The stack_guard is accessed directly by GCC -fstack-protector code,
   so it is a part of public ABI.  The dtv and pointer_guard fields
   are private.  */
typedef struct
{
  /* Reservation for HWCAP data.  To be accessed by GCC in
     __builtin_cpu_supports(), so it is a part of public ABI.  */
  uint64_t hwcap;
  /* Reservation for AT_PLATFORM data.  To be accessed by GCC in
     __builtin_cpu_is(), so it is a part of public ABI.  Since there
     are different ABIs for 32 and 64 bit, we put this field in a
     previously empty padding space for powerpc64.  */
#ifndef __powerpc64__
  /* Padding to maintain alignment.  */
  uint32_t padding;
  uint32_t at_platform;
#endif
  uint32_t __unused;
  /* Reservation for AT_PLATFORM data - powerpc64.  */
#ifdef __powerpc64__
  uint32_t at_platform;
#endif
  /* Reservation for Dynamic System Optimizer ABI.  */
  uintptr_t dso_slot2;
  uintptr_t dso_slot1;
  /* Reservation for tar register (ISA 2.07).  */
  uintptr_t tar_save;
  /* GCC split stack support.  */
  void *__private_ss;
  /* Reservation for the Event-Based Branching ABI.  */
  uintptr_t ebb_handler;
  uintptr_t ebb_ctx_pointer;
  uintptr_t ebb_reserved1;
  uintptr_t ebb_reserved2;
  uintptr_t pointer_guard;
  uintptr_t stack_guard;
  dtv_t *dtv;
} tcbhead_t;
```
I am not sure under what circumstances gcc will call into ` __builtin_cpu_supports()` or `__builtin_cpu_is()`?

The gcc man page doesn't seem to indicate we can eliminate access to these functions:

```
       -fno-builtin
       -fno-builtin-function
           Don't recognize built-in functions that do not begin with __builtin_ as prefix.

           GCC normally generates special code to handle certain built-in functions more
           efficiently; for instance, calls to "alloca" may become single instructions which
           adjust the stack directly, and calls to "memcpy" may become inline copy loops.  The
           resulting code is often both smaller and faster, but since the function calls no
           longer appear as such, you cannot set a breakpoint on those calls, nor can you change
           the behavior of the functions by linking with a different library.  In addition, when
           a function is recognized as a built-in function, GCC may use information about that
           function to warn about problems with calls to that function, or to generate more
           efficient code, even if the resulting code still contains calls to that function.  For
           example, warnings are given with -Wformat for bad calls to "printf" when "printf" is
           built in and "strlen" is known not to modify global memory.

           With the -fno-builtin-function option only the built-in function function is disabled.
           function must not begin with __builtin_.  If a function is named that is not built-in
           in this version of GCC, this option is ignored.  There is no corresponding
           -fbuiltin-function option; if you wish to enable built-in functions selectively when
           using -fno-builtin or -ffreestanding, you may define macros such as:

                   #define abs(n)          __builtin_abs ((n))
                   #define strcpy(d, s)    __builtin_strcpy ((d), (s))
```

stefanberger:
Closing this PR now in favor of the other one...

mato:
> I am not sure under what circumstances gcc will produce code that calls into __builtin_cpu_supports() or __builtin_cpu_is()?

> The gcc man page doesn't seem to indicate we can eliminate access to these functions:

That's what `-ffreestanding` does.

PR #369, opened 2019-06-06
Allow for differing guest page size

mato:
Some architectures do not/cannot use 4k pages, so allow for changing the target-specific "guest page size", i.e. the value passed as -z max-page-size to the linker. Update the linker scripts to take this value from the linker rather than hard-coding it.

NOTE: This is a test to see if this approach is portable to the BSDs, specifically if the LLVM linker is happy with `CONSTANT(MAXPAGESIZE)`.


mato:
This looks like it does the right thing, verified by dumping the binaries during CI in a temporary commit and alignment is being passed through correctly. Merging.

PR #371, opened 2019-06-07
hvt_dumpcore: use openat(2), change option to --dumpcore=DIR

sg2342:


mato:
@sg2342 This looks good, thanks! Merging.

Issue #79, opened 2016-08-19
Google Compute Engine VMs only support virtio-scsi disk devices


ricarkol:
GCE has only one type of disks: virtio-scsi (from https://cloud.google.com/compute/docs/tutorials/building-images). We currently don't support these. This issue is tracking that effort.

mato:
Marking this as "help wanted" in the sense that its up for grabs for someone that needs the functionality. Current goals are concentrated on targets other than _virtio_.

mbacarella:
I've offered a $100 bounty for this. Perhaps others will join me: https://www.bountysource.com/orders/133856

It might be better to implement the virtio SCSI device in pure OCaml and introduce hooks into Solo5 to support it. But your call really.

reynir:
The URL seems to have moved slightly: https://cloud.google.com/compute/docs/images/building-custom-os

PR #377, opened 2019-07-19
Update virtio net and block to handle multiple devices

ricarkol:
This PR is really 2 changes: virtio changes to support multiple devices, and dealing with the application manifest. Dealing with the application manifest is tricky as PCI devices have no names.
The easiest solution would be to have the user explicitely map devices to PCI slots in the manifest: `{ ..."type": "NET_BASIC", "PCI":"bus1:0x4:0"}`. But this would be a pain as not much people use the virtio backend. Moreover, most cloud providers don't give enough control to users to define what buses and PCI addresses to use. They just allow users to define the ordering of devices. So, that's what this PR uses: the ordering of devices. For example, `virtio-run -n tap1 -d /p/D1 -n tap2` maps to the following manifest like this:

```
    "devices": [
        { "name": "service1", "type": "NET_BASIC" },	--> tap1
        { "name": "disk1", "type": "BLK_BASIC" },	--> /p/D1
        { "name": "service2", "type": "NET_BASIC" }	--> tap2
    ]
```

A second PR after this will add a new virtio-run tool which would take the same arguments as the tenders (`--net:service1=tap1 ...`). This tool would read the manifest from the application and use it to define the PCI slots of all the devices passed by the user.

These are all the changes:

- virtio/pci.c and virtio/start.c:
    PCI devices found in pci_enumerate are assigned to devices from the application
    manifest in the order in which they are found. This order is given by the PCI
    slot numbers which are controlled in solo5-virtio-run.sh.  The application
    fails (during PCI enumeration) if the type of a device just found does not
    match the type of the next expected device in the manifest.  Devices whose type
    (e.g., virtio_scsi) are unknown are just skipped. We do not fail in this case
    because GCE always includes virtio_scsi devices (can't control that).
    
- virtio/virtio_net.c and virtio_blk.c:
    Moved all device specific objects, like the virtio queues and the PCI base
    address, to `virtio_[blk|net]_desc` structures.  These are initialized at config
    time and stored in a table (bd_table and nd_table) indexed by their respective
    solo5_handle's hostfd.  Also moved all virtio specific code to `static
    virtio_[blk|net]_*` functions.  These functions only deal with
    `virtio_[blk|net]_desc` structures.  All the other ones, like `solo5_block_read`,
    translate `solo5_handles` using the hostfd field in the handle.

- solo5-virtio-run.sh:
    As a first version, add the ability to have multiple -d DISK and -n NETIF
    arguments. The order of the arguments is used as the PCI slot number (starting
    at 2 as those are reserved).

### DIFF
--- a/bindings/virtio/bindings.h
+++ b/bindings/virtio/bindings.h
@@ -58,15 +58,9 @@ struct pci_config_info {
     uint8_t irq;
 };
 
-void pci_enumerate(void);
+int pci_enumerate(struct mft *mft);
 
-/* virtio.c: mostly net for now */
-void virtio_config_network(struct pci_config_info *);
-void virtio_config_block(struct pci_config_info *);
-
-uint8_t *virtio_net_pkt_get(size_t *size);  /* get a pointer to recv'd data */
-void virtio_net_pkt_put(void);      /* we're done with recv'd data */
-int virtio_net_xmit_packet(const void *data, size_t len);
-int virtio_net_pkt_poll(void);      /* test if packet(s) are available */
+int virtio_config_network(struct pci_config_info *, solo5_handle_t, struct mft *mft);
+int virtio_config_block(struct pci_config_info *, solo5_handle_t, struct mft *mft);
 
 #endif /* __VIRTIO_BINDINGS_H__ */
--- a/bindings/virtio/pci.c
+++ b/bindings/virtio/pci.c
@@ -53,44 +53,57 @@
 } while (0)
 
 
-static uint32_t net_devices_found;
-static uint32_t blk_devices_found;
-
+static solo5_handle_t mft_index = 0;
 #define PCI_CONF_SUBSYS_NET 1
 #define PCI_CONF_SUBSYS_BLK 2
 
-static void virtio_config(struct pci_config_info *pci)
+/*
+ * mft_index tracks the order in which devices are found (given by the
+ * PCI slots). This is used to map found devices to devices in the manifest.
+ */
+static int virtio_config(struct pci_config_info *pci, struct mft *mft)
 {
     /* we only support one net device and one blk device */
     switch (pci->subsys_id) {
     case PCI_CONF_SUBSYS_NET:
         log(INFO, "Solo5: PCI:%02x:%02x: virtio-net device, base=0x%x, irq=%u\n",
             pci->bus, pci->dev, pci->base, pci->irq);
-        if (!net_devices_found++)
-            virtio_config_network(pci);
-        else
-            log(WARN, "Solo5: PCI:%02x:%02x: not configured\n", pci->bus,
-                pci->dev);
+        if (virtio_config_network(pci, mft_index, mft) != 0)
+            return -1;
+        mft_index++;
         break;
     case PCI_CONF_SUBSYS_BLK:
         log(INFO, "Solo5: PCI:%02x:%02x: virtio-block device, base=0x%x, irq=%u\n",
             pci->bus, pci->dev, pci->base, pci->irq);
-        if (!blk_devices_found++)
-            virtio_config_block(pci);
-        else
-            log(WARN, "Solo5: PCI:%02x:%02x: not configured\n", pci->bus,
-                pci->dev);
+        if (virtio_config_block(pci, mft_index, mft) != 0)
+            return -1;
+        mft_index++;
         break;
     default:
+        /* Most likely this is something like a virtio_scsi device
+           that we can not handle. Let's not handle nor count it
+           in pci_virtio_index. */
         log(WARN, "Solo5: PCI:%02x:%02x: unknown virtio device (0x%x)\n",
             pci->bus, pci->dev, pci->subsys_id);
-        return;
     }
+    return 0;
 }
 
 #define VENDOR_QUMRANET_VIRTIO 0x1af4
 
-void pci_enumerate(void)
+/*
+ * Every PCI device found in pci_enumerate is attempted to be assigned to a
+ * device from the application manifest in the order in which they are found. This
+ * order is given by the PCI slot number which is controlled in
+ * solo5-virtio-run.sh.
+ * 
+ * This funtion can fail if the type of a device just found does not match the
+ * type of the next expected device in the manifest. Devices whose type (e.g.,
+ * virtio_scsi) is unknown is just skipped.
+ * 
+ * Returns 0 on success, -1 otherwise.
+ */
+int pci_enumerate(struct mft *mft)
 {
     uint32_t bus;
     uint8_t dev;
@@ -116,8 +129,10 @@ void pci_enumerate(void)
                 PCI_CONF_READ(uint16_t, &pci.base, config_addr, IOBAR);
                 PCI_CONF_READ(uint8_t, &pci.irq, config_addr, IRQ);
 
-                virtio_config(&pci);
+                if (virtio_config(&pci, mft) != 0)
+                    return -1;
             }
         }
     }
+    return 0;
 }
--- a/bindings/virtio/start.c
+++ b/bindings/virtio/start.c
@@ -49,6 +49,23 @@ void _start(void *arg)
 
 extern struct mft_note __solo5_manifest_note;
 
+/* Will abort the program if any of the devices is not acquired. */
+static void mft_check_all_acquired(struct mft *mft)
+{
+    bool fail = false;
+    for (unsigned i = 0; i != mft->entries; i++) {
+        if (!mft->e[i].attached) {
+            log(WARN, "Solo5: Device '%s' of type %s not attached.",
+                    mft->e[i].name, mft_type_to_string(mft->e[i].type));
+            fail = true;
+        }
+    }
+    if (fail) {
+        log(ERROR, "Solo5: All declared devices must be attached. Aborting.\n");
+        solo5_abort();
+    }
+}
+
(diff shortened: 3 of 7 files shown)


ricarkol:
Rebased on top of latest master (including #380).

mato:
@ricarkol:

> Dealing with the application manifest is tricky as PCI devices have no names. [...] Moreover, most cloud providers don't give enough control to users to define what buses and PCI addresses to use. 

Indeed, this is a problem with virtio devices. To be honest, I wasn't planning to support multiple devices for virtio at all, but your approach is interesting. Not sure if it holds together though.

Meta-question: Do you actually _need_ this functionality, or have your implemented it just for completeness/fun?

> They just allow users to define the ordering of devices. So, that's what this PR uses: the ordering of devices. For example, virtio-run -n tap1 -d /p/D1 -n tap2 maps to the following manifest like this:

    "devices": [
        { "name": "service1", "type": "NET_BASIC" },	--> tap1
        { "name": "disk1", "type": "BLK_BASIC" },	--> /p/D1
        { "name": "service2", "type": "NET_BASIC" }	--> tap2
    ]

I've read through both your description of the design and the code changes, and I don't quite follow how the mapping of PCI devices to manifest entries is intended to work. Using the above manifest as an example, what happens if:

1. The order of the devices presented by the hypervisor changes? Say we have: virtio-net@slot2, virtio-block@slot3, virtio-net@slot4. What happens if this changes to: virtio-net@slot2, virtio-net@slot3, virtio-block@slot4?
2. Conversely, what if the order of the devices in the manifest changes? Note that a JSON object is technically unordered, so it would be "within the spec" for the Solo5 tooling to randomly shuffle the order of devices in the manifest.

Thinking about the alternatives, you write:

> The easiest solution would be to have the user explicitely map devices to PCI slots in the manifest: { ..."type": "NET_BASIC", "PCI":"bus1:0x4:0"}.

That doesn't work, since the manifest is explicitly meant to be agnostic of any "target" wiring, and you'd effectively introduce a virtio-specific interpretation of it.

What we could do (bar a bunch of annoying parsing code), is define the mapping of devices to PCI slots _at run time_ on the command line. Using the above manifest, here are a couple of straw man examples:

Fixed slot numbers:

    --solo5:virtio:map:service1=virtio-net@pci:2 --solo5:virtio:map:service2=virtio-net@pci:3 --solo5:map:disk1=virtio-block@pci:4

Ordering based on PCI detection order _within device types_, i.e. virtio-net@0 is "the first virtio-net device":

    --solo5:virtio:map:service1=virtio-net@0 --solo5:virtio:map:service2=virtio-net@1 --solo5:map:disk1=virtio-block@0

Thinking about it, the latter feels like the best compromise between mapping things correctly and at the same time not relying too much on the specifics of how the hypervisor lays out the entire PCI bus.

However, I will ask my meta-question again... is this worth the extra complexity? Are you planning to use this functionality?

ricarkol:
Thanks for the review @mato.

> However, I will ask my meta-question again... is this worth the extra complexity? Are you planning to use this functionality?

I usually use the virtio backend for comparison purposes (e.g., performance experiments). Not sure if anybody is running mirage at some public cloud (like GCP). I remember @sgrove used to (I used to have a small mirage demo service in GCP). Another reason is that it would be nice to eventually support firecracker, so keeping the virtio backend up to date could become useful.

> 2. Conversely, what if the order of the devices in the manifest changes? Note that a JSON object is technically unordered, 

The current manifest JSON uses an array to list the devices. Arrays are ordered: "JSON arrays are ordered lists of values" [0].

> 1. (What happens if) The order of the devices presented by the hypervisor changes?

Tested this assumption in qemu and bhyve for virtio-blk and net devices and a small number of devices (~8). Couldn't find this "contract" in their respective documentation, so not sure. My missing doubt is what happens with a larger number of devices.

> Ordering based on PCI detection order within device types, i.e. virtio-net@0 is "the first virtio-net device":

This seems like the right assumption. So, considering that `2.` can be safely assumed (the manifest has an ordered list of devices). We could merge the two without needing the runtime args. Will work on this approach.

[0] https://restfulapi.net/json-array/

mato:
> I usually use the virtio backend for comparison purposes (e.g., performance experiments). Not sure if anybody is running mirage at some public cloud (like GCP). I remember @sgrove used to (I used to have a small mirage demo service in GCP). Another reason is that it would be nice to eventually support firecracker, so keeping the virtio backend up to date could become useful.

Yes, having the virtio backend working is useful. I was just pointing out that we don't necessarily need to support multiple devices until such a point as someone explicitly asks for it. In my opinion, what would be much more useful would be implementing the minimum required to get virtio-scsi working (as without that we have no support for persistence on GCP (#79).

> The current manifest JSON uses an array to list the devices. Arrays are ordered: "JSON arrays are ordered lists of values" [0].

This is true.

> Tested this assumption in qemu and bhyve for virtio-blk and net devices and a small number of devices (~8). Couldn't find this "contract" in their respective documentation, so not sure. My missing doubt is what happens with a larger number of devices.

I don't think we can count on any such contract. With cloud hypervisors all we can depend on is ordering within a device group (i.e. "first network card", "second network card" implies ordering within the PCI slot(s)/function(s) for virtio-net).

> So, considering that 2. can be safely assumed (the manifest has an ordered list of devices). We could merge the two without needing the runtime args. Will work on this approach.

We _can't assume_ that the order of the devices in the manifest has side-effects (for virtio). We _could specify_ it that way, but that seems like a bad idea as it places lots of "surprising" constraints on the downstream build system. For example, in MirageOS, the device declarations come from _code_ in `config.ml`, and are transformed by Functoria into the generated implementation. Imposing ordering constraints on that would be hard, if not impossible, without introducing hacks like sorting on the device names themselves.

ricarkol:
> We can't assume that the order of the devices in the manifest has side-effects (for virtio). We could specify it that way, but that seems like a bad idea as it places lots of "surprising" constraints on the downstream build system. For example, in MirageOS, the device declarations come from code in config.ml, and are transformed by Functoria into the generated implementation. Imposing ordering constraints on that would be hard, if not impossible, without introducing hacks like sorting on the device names themselves.

I agree that the unikernel build system should not care about the ordering. Luckily, there is no need. The only place where the ordering constraint has to be maintained is at the manifest library/tool level: if the manifest has `[netB, netA]`, then the MFT library and tool should list devices in that same order (more on why below). The solo5 build system ensures this by storing devices in the ELF note in the same order as the manifest [1]. If the user passes devices in that same order, then the virtio bindings can map the first device they get to netB and the second to netA, according to the list obtained from the MFT library [2].

Let me go in more detail with that example. Let's say the mirage tool builds a unikernel with a configuration JSON like this: `[netB, netA]`. The order doesn't matter at this point: the unikernel could have generated the other order. Before starting this unikernel on the virtio backend, the user would check the configuration json (or use solo5-mfttool) and see the list `[netB, netA]`. He would then start it while passing devices in that order `virtio-run -n netB -n netA`. The bindings would get `[PCI:1, PCI:2]` from the enumeration, and the list `[netB, netA]` from the MFT library. Assuming they have the same order (the only constraint we need), the bindings can map `PCI:1` to `netB` and `PCI:2` to `netA` (and vice-versa). If the unikernel `acquires` and does IO on `netB`, then the bindings would do it on `PCI:1`.

[1] Generating the ELF note is always done by solo5 right @mato? mirageOS/Functoria just creates the JSON right?
[2] The MFT library has a `get_by_index` function which can be used to enumerate all devices in order.

mato:
@ricarkol:

> I agree that the unikernel build system should not care about the ordering. Luckily, there is no need. 

You misunderstand me. What I meant in this context was that the unikernel _developer_ should not care about the ordering. In the case of someone developing a Mirage unikernel, they don't (and shouldn't) have control of whether or not a manifest gets generated in any particular order. Their code should only care about the _name_ of a particular device.

Further, the other party involved, i.e. the _operator_ (be that a person or system), should be able to specify how devices get wired in an easy to understand, robust and deterministic fashion _on all targets where this is possible_. On targets where this is _not_ possible, such as virtio, we should still strive for behaviours that are easy to understand. You write:

> The only place where the ordering constraint has to be maintained is at the manifest library/tool level: if the manifest has [netB, netA], then the MFT library and tool should list devices in that same order (more on why below). The solo5 build system ensures this by storing devices in the ELF note in the same order as the manifest [1]. If the user passes devices in that same order, then the virtio bindings can map the first device they get to netB and the second to netA, according to the list obtained from the MFT library [2].

Notwithstanding (1) and (2) which happen to be true at the moment, _"If the user passes devices in that same order"_ (where by user I presume you mean the _operator_) is a big if, especially where there are two parties involved as I describe above. The above behaviour doesn't pass my "easy to understand" or "robust in the event of unexpected changes" test.

So, sorry, but I strongly prefer either the current situation (one device of each max, grab the first one), or am open to any ideas that don't rely on subtle side-effects in the manifest like ordering. Even then, I'm note sure my previous proposal (mapping on the command line) is robust enough, but it's _just_ good enough that I'd take something like that while squinting and chanting "it's just virtio". :-)

ricarkol:
Will give the proposal of parsing in the command line a try (the example below). If the code looks too messy too quick, will just close this PR.

```
--solo5:virtio:map:service1=virtio-net@0 --solo5:virtio:map:service2=virtio-net@1 --solo5:map:disk1=virtio-block@0
```

mato:
No activity for >6 months and no clear pull from users, so closing.

Issue #372, opened 2019-06-12
Support for multiple devices


mato:
This issue sets out the high-level design of upcoming support for multiple devices in Solo5, from the point of view of the unikernel application ("developer") and deployment operator ("user"). It is intended to be read/discussed in conjunction with the work in progress implementation in PR #373.

Steps to completion:

1. [x] High-level design document (this issue).
2. [x] Implementation for _hvt_, _spt_, _virtio_ (partial), _genode_ (~~#373~~ #375).
3. [x] Implementation for _muen_ (#402).
4. [x] MirageOS integration (see mirage/mirage#992).

## Application manifest

We introduce the concept of an _application manifest_, which is defined by the developer at application build time, using the following JSON format and customarily named `manifest.json`:

```
{
    "version" = 1,
    "devices" = [
        { "name" = "<NAME>", "type" = "<TYPE>" },
        ...
    ]
}
```

_NAME_ is a human-readable unique identifier for the device. The intention is that the name conveys some meaning of the intended use ("wiring") of the device, eg. `frontend` or `storage`. _NAME_ must be composed of alphanumeric characters only, and within 1..67 characters in length.

_TYPE_ is the type of device being declared, currently `BLOCK_BASIC` or `NET_BASIC`.

Note that _NAME_ must be unique, and there is a maximum limit of 64 devices in the manifest.

At build time, `manifest.json` is pre-processed by the newly introduced `solo5-mfttool`, generating a C source file with a binary representation. This source file is then compiled using the normal Solo5 tool chain, and linked into the unikernel binary, where it is represented as an ELF NOTE.

## API changes

We introduce two new data types, `solo5_handle_t` to represent an opaque "handle" to a device, and `solo5_handle_set_t` to represent a set of up to 64 handles represented as a bitmask.

The Solo5 public API will change as follows (abridged):

```
void solo5_yield(solo5_time_t deadline, solo5_handle_set_t *ready_set);
solo5_result_t solo5_net_acquire(const char *name, solo5_handle_t *handle,
        struct solo5_net_info *info);
solo5_result_t solo5_net_write(solo5_handle_t handle, const uint8_t *buf,
        size_t size);
solo5_result_t solo5_net_read(solo5_handle_t handle, uint8_t *buf,
        size_t size, size_t *read_size);
solo5_result_t solo5_block_acquire(const char *name, solo5_handle_t *handle,
        struct solo5_block_info *info);
[...]

PR #379, opened 2019-08-07
Reserve first manifest entry, use MFT_DEV_XXX for device types

mato:
Reserve the first manifest entry, to prevent the use of `0` as a valid handle. Namespace manifest device types under MFT_DEV_XXX.

@cfcs, @ricarkol any comments / objections to this approach? For extra paranoia/bounds checking in the manifest validation, is it worth also making the _last_ manifest entry mandatory in a similar way, with a different reserved value?


ricarkol:
No objections. Looks good. Importantly, all the users of the manifest API (like the bindings) stay the same, which I checked partially with PR #377.

> For extra paranoia/bounds checking in the manifest validation, is it worth also making the last manifest entry mandatory in a similar way, with a different reserved value?

Not sure. There are already lots of checks against `mft->entries` and `MFT_MAX_ENTRIES`.

mato:
> Not sure. There are already lots of checks against mft->entries and MFT_MAX_ENTRIES.

Those would stay as they are, with no more checks added. All I was thinking of was adding a `MFT_RESERVED_LAST` as a terminator, and checking in `mft_validate()` only that:

1. mft->entries is at least 2.
2. mft->e[mft->entries].type contains `MFT_RESERVED_LAST`.

This would provide extra assurance that we did indeed load the _entire_ manifest correctly.

ricarkol:
> This would provide extra assurance that we did indeed load the entire manifest correctly.

Yes, that makes sense. Plus, 62 devices vs. 63 is not a bit deal.

mato:
Ok, done, rebased and force-pushed. I also re-ordered the enum to place the reserved values at a high starting point.

mfttool/mfttool.c
```diff
@@ -66,18 +66,21 @@ static void jexpect(enum jtypes t, jvalue *v, const char *loc)
...
     "\n"
     "MFT_NOTE_BEGIN\n"
     "{\n"
     "  .version = MFT_VERSION, .entries = %d,\n"
-    "  .e = {\n";
+    "  .e = {\n"
+    "    { .name = \"\", .type = MFT_RESERVED_ZERO },\n";
 
 static const char out_entry[] = \
     "    { .name = \"%s\", .type = MFT_%s },\n";
```
cfcs: similarly here I guess something like `%-64s` would force initialization?

include/solo5/mft_abi.h
```diff
@@ -39,11 +39,13 @@
...
 /*
- * Supported device types.
+ * Supported manifest entry types.
  */
 typedef enum mft_type {
-    MFT_BLOCK_BASIC,
-    MFT_NET_BASIC
+    MFT_BLOCK_BASIC = 1,
+    MFT_NET_BASIC,
+    MFT_RESERVED_ZERO = (1U << 31),
```
cfcs: Kind of pedantic, but would you mind making this `30` instead of `31` to avoid having to deal with unsigned/signed uncertainties?

tenders/hvt/hvt_main.c
```diff
@@ -54,6 +54,8 @@ static void setup_modules(struct hvt *hvt, struct mft *mft)
 
     bool fail = false;
     for (unsigned i = 0; i != mft->entries; i++) {
```
cfcs: I guess these loops can now also start at `1`?

tenders/spt/spt_main.c
```diff
@@ -53,6 +53,8 @@ static void setup_modules(struct spt *spt, struct mft *mft)
 
     bool fail = false;
     for (unsigned i = 0; i != mft->entries; i++) {
```
cfcs: same here re: starting at `1` instead of `0`

cfcs:
1) I think it makes sense with a dummy entry at the beginning.
2) I'm not a fan of having both a terminator entry AND a `entries` count since it increases the number of "representable illegal states" (trailing stuff after the terminator now has to be checked). It seems to me that `entries` can be derived from `mft_size` which is passed as an argument to the validation function regardless, but that complicates the size calculation elsewhere.
3) I wonder if we can fit some kind of useful information into the entry rather than having it be a dummy placeholder. The Solo5 ABI version, perhaps?
4)  The names `MFT_RESERVED_ZERO` and `MFT_RESERVED_LAST` seem a bit weird to me, I would have expected `FIRST`/`LAST`.

mato:
@cfcs:

> I'm not a fan of having both a terminator entry AND a entries count since it increases the number of "representable illegal states" (trailing stuff after the terminator now has to be checked). 

I don't follow why we would need to check any trailing data after the terminator, or what we would check for?

> It seems to me that entries can be derived from mft_size which is passed as an argument to the validation function regardless, but that complicates the size calculation elsewhere.

I'd prefer not to depend on/interpret `mft_size` too much as it represents the structure size including packing, and as such is dependent on various whims of the toolchain.

If we want to choose between having `mft->entries` OR a terminator, which is better? Implementation-wise, the majority of the code needs to iterate over the entire manifest. With a `mft->entries`, this is generally done by:

    for (unsigned i = 0; i != mft->entries; i++)
        /* do something with mft->e[i], and/or use i as an (index to) a handle */

With a terminator, this would become:

    for (unsigned i = 0; mft->e[i].type != MFT_RESERVED_LAST; i++)
        /* do something with mft->e[i], and/or use i as an (index to) a handle */

The former is simpler, so that is in favour of keeping `mft->entries` and not having a terminator.

WDYT?

> I wonder if we can fit some kind of useful information into the entry rather than having it be a dummy placeholder. The Solo5 ABI version, perhaps?

Good point, let's discuss this in #372 as it concerns the manifest design in general.

> The names MFT_RESERVED_ZERO and MFT_RESERVED_LAST seem a bit weird to me, I would have expected FIRST/LAST.

I can change those, once we decide what goes where. I'll also change the device types to be `MFT_DEV_<something>`.

mfttool/mfttool.c
```diff
@@ -66,18 +66,21 @@ static void jexpect(enum jtypes t, jvalue *v, const char *loc)
...
+    "/* Generated by solo5-mfttool, do not edit */\n\n"
     "#define MFT_ENTRIES %d\n"
     "#include \"mft_abi.h\"\n"
     "\n"
     "MFT_NOTE_BEGIN\n"
     "{\n"
     "  .version = MFT_VERSION, .entries = %d,\n"
-    "  .e = {\n";
+    "  .e = {\n"
+    "    { .name = \"\", .type = MFT_RESERVED_ZERO },\n";
```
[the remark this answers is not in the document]
mato: `static` and global variables are always initialised to zero implicitly (by the linker/loader), so there's no point in explicitly doing that.

include/solo5/mft_abi.h
mato: Sure, will do. Good point since `enum` is signed.

tenders/hvt/hvt_main.c
mato: I added

        if (mft->e[i].type >= MFT_RESERVED_ZERO)
            continue;

in the loop body, it's more explicit than a `1`-based loop.

mato:
Ok, after some more discussion the consensus is to not have a MFT_RESERVED_LAST as a termination, `mft->entries` is enough. I've rebased this atop latest master, and in an additional commit have also namespaced the manifest device types as `MFT_DEV_XXX`.

Will merge shortly.

PR #375, opened 2019-07-11
Multiple devices support, application manifest

mato:
Squashed and rebased version of #373, plus Genode support from @ehmry. Muen is currently disabled, pending multiple-devices support from @Kensan.

Heads up: This will be merged into master by tomorrow at the latest.

PR #380, opened 2019-08-09
solo5_yield() should return void

mato:
Part of #372 / clean-up of #375 :

@hannesm pointed out to me that having `solo5_yield()` return both a `ready_set` and `bool` is redundant, so change the return type to `void`.

@ehmry I've updated the Genode bindings but not tested them.
@ricarkol #377 will need some changes to account for this.


mato:
I've not seen any objections here, so merging.

PR #385, opened 2019-08-15
ELF loader: Refactor and fix multiple issues.

### LINKED ISSUE #382: ELF loader: memsz < filesz memset length underflow
(moving here from [original comment](https://github.com/Solo5/solo5/pull/375#discussion_r306338363) so we can keep track of this)

At the moment,  `filesz` (`phdr[ph_i].p_filesz`) is subtracted from `memsz` (`phdr[ph_i].p_memsz`) for each `Elf64_Phdr` without underflow checking during the loading of unikernel ELF files, and the resulting value is used as the `length` argument of a `memcpy` call:
https://github.com/Solo5/solo5/blob/master/tenders/common/elf.c#L187

It is [guaranteed (by the code)](https://github.com/Solo5/solo5/blob/master/tenders/common/elf.c#L161-L162) that `paddr + files < mem_size` (all `uint64_t`).
It is [also guaranteed](https://github.com/Solo5/solo5/blob/master/tenders/common/elf.c#L164-L165) that `paddr + memsz < memsz`.

It is however not guaranteed that `memsz - filesz >= 0`,  which means the `size_t` result can underflow when `filesz > memsz`, resulting in a very large `length` argument being passed to the `memset(daddr + filesz, 0, memsz - filesz);` call. Since the value of `filesz` is bounded by the `mem_size` allocated to the unikernel by the administrator, it is unlikely that a malicious ELF file can avoid eventually crashing with a segmentation fault, but the intermediary writes could have security implications (and segfaults are never nice), so I think we should fix this.

I've attached a test vector (I tried to write a unit test and a PR to add it as a regression test to the the test framework, but I got lost in the Makefile/shell script soup and ultimately failed to produce a negative test for a specific tender):
```c
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>
#include <stdio.h>
#include <elf.h>
#include <string.h>
#include "../include/solo5/mft_abi.h"
int
main(){
  Elf64_Ehdr hdr;

  /* make ehdr_is_valid happy: */
  hdr.e_ident[EI_MAG0] = ELFMAG0;
  hdr.e_ident[EI_MAG1] = ELFMAG1;
  hdr.e_ident[EI_MAG2] = ELFMAG2;
  hdr.e_ident[EI_MAG3] = ELFMAG3;
  hdr.e_ident[EI_CLASS] = ELFCLASS64;
  hdr.e_type = ET_EXEC;
  hdr.e_machine = EM_X86_64;

  /* set ph params */
  hdr.e_phoff = sizeof(Elf64_Ehdr);
  hdr.e_phentsize = sizeof (Elf64_Phdr);
  hdr.e_phnum = 0x2;

  for(size_t i = 0; i < sizeof(Elf64_Ehdr); i++){
    printf("%c", ((char *)&hdr)[i]);
  }

  /////////////////// solo5 manifest  
  Elf64_Phdr phdr = {0};

  phdr.p_type = PT_NOTE; // satisfy check in loop

  phdr.p_paddr = 0;
  phdr.p_align = 0;
  phdr.p_offset = sizeof(Elf64_Ehdr) + 2*sizeof(Elf64_Phdr);
  // ^-- at the end of this file

  /* make mft_validate() happy */
  phdr.p_filesz = sizeof(struct mft_note);
  struct mft_note nhdr = {0};
  memcpy(&nhdr.h.name, SOLO5_NOTE_NAME, sizeof(SOLO5_NOTE_NAME));
  nhdr.h.type = SOLO5_NOTE_MANIFEST;
  nhdr.h.descsz = 0;
  nhdr.m.version = MFT_VERSION;
  nhdr.m.entries = 0;

  for(size_t i = 0; i < sizeof(Elf64_Phdr); i++){
    printf("%c", ((char *)&phdr)[i]);
  }

  /////////////////// memsz < filesz 
  memset(&phdr, 0, sizeof(phdr));

  phdr.p_type = PT_LOAD; // satisfy check in loop

  phdr.p_addr = 0x100000; // hmm, got confusing results without this

  phdr.p_memsz = 0;
  phdr.p_filesz = sizeof(Elf64_Ehdr) + sizeof(Elf64_Phdr);
  // memsz < filesz, to underflow in the memset


  for(size_t i = 0; i < sizeof(Elf64_Phdr); i++){
    printf("%c", ((char *)&phdr)[i]);
  }

  for(size_t i = 0; i < sizeof(struct mft_note); i++){
    printf("%c", ((char *)&nhdr)[i]);
  }

};
```

ping @mato

### LINKED ISSUE #383: ELF loader: ehdr.ph_entsz < sizeof(Elf64_phdr) out-of-bounds reads
(moved from [original comment](https://github.com/Solo5/solo5/pull/375#discussion_r302975349) so we can keep track of this)

If an ELF file contains an `Elf64_Ehdr` with a `ph_entsz` < sizeof(Elf64_Phdr)` we end up reading out-of-bounds when accessing the members of the phdr struct: https://github.com/Solo5/solo5/blob/master/tenders/common/elf.c#L131-L154

- I believe we should validate that `ph_entsz >= sizeof(Elf64_Phdr)` in `ehdr_is_valid()`.

Here is a test vector:
```c
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>
#include <stdio.h>
#include <elf.h>
int
main(){
  Elf64_Ehdr hdr;

  /* make ehdr_is_valid happy: */
  hdr.e_ident[EI_MAG0] = ELFMAG0;
  hdr.e_ident[EI_MAG1] = ELFMAG1;
  hdr.e_ident[EI_MAG2] = ELFMAG2;
  hdr.e_ident[EI_MAG3] = ELFMAG3;
  hdr.e_ident[EI_CLASS] = ELFCLASS64;
  hdr.e_type = ET_EXEC;
  hdr.e_machine = EM_X86_64;

  /* set ph params */
  hdr.e_phoff = 0;
  hdr.e_phentsize = 0;
  hdr.e_phnum = 0xffff; /* as many as possible to get a segfault */

  for(size_t i = 0; i < sizeof(Elf64_Ehdr); i++){
    printf("%c", ((char *)&hdr)[i]);
  }
};
```

And the result:
```shell
~/c/solo5 gcc trigger.c > /dev/shm/trigger.elf
~/c/solo5 ./tenders/hvt/solo5-hvt /dev/shm/trigger.elf
Segmentation fault
~/c/solo5 gdb --args ./tenders/hvt/solo5-hvt-debug /dev/shm/trigger.elf
GNU gdb (Debian 8.2.1-2) 8.2.1
This GDB was configured as "x86_64-linux-gnu".
Reading symbols from ./tenders/hvt/solo5-hvt-debug...done.
(gdb) run
Starting program: /home/user/c/solo5/tenders/hvt/solo5-hvt-debug /dev/shm/trigger.elf

Program received signal SIGSEGV, Segmentation fault.
elf_load_mft (file=0x7fffffffe3eb "/dev/shm/trigger.elf", mft=0x7fffffffdec0, mft_size=0x7fffffffdec8) at common/elf.c:269
269	        if (phdr[ph_i].p_type != PT_NOTE)
(gdb) 
```

### LINKED ISSUE #384: MFT: mft_validate() out-of-bounds read
(copied from [original comment](https://github.com/Solo5/solo5/pull/375#discussion_r306280518) so that we can keep track of this)

We [currently have](https://github.com/Solo5/solo5/blob/master/tenders/common/mft.c#L39-L41):

```c
int mft_validate(struct mft *mft, size_t mft_size)
{
    if (mft->version != MFT_VERSION)
        return -1;
    if (mft->entries > MFT_MAX_ENTRIES)
        return -1;
```

If the buffer read (`*mft`) is smaller than `sizeof(struct mft)`, `mft->version` could read past the end of `*mft`. It doesn't seem to be guaranteed(?) that this can't happen.

I think adding a check like the code below to the top of this function would make sense:
```c
    if (mft_size < sizeof(*mft))
        return -1;
```

Here's a test vector that caused `solo5-spt` to end up calling `mft_validate()` with a `mft_size` of 0:

```c
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>
#include <stdio.h>
#include <elf.h>
#include "./include/solo5/mft_abi.h"
int
main(){
  Elf64_Ehdr hdr = {0};

  /* make ehdr_is_valid happy: */
  hdr.e_ident[EI_MAG0] = ELFMAG0;
  hdr.e_ident[EI_MAG1] = ELFMAG1;
  hdr.e_ident[EI_MAG2] = ELFMAG2;
  hdr.e_ident[EI_MAG3] = ELFMAG3;
  hdr.e_ident[EI_CLASS] = ELFCLASS64;
  hdr.e_type = ET_EXEC;
  hdr.e_machine = EM_X86_64;

  /* set ph params */
  hdr.e_phoff = sizeof(Elf64_Ehdr);
  hdr.e_phentsize = sizeof (Elf64_Phdr);
  hdr.e_phnum = 0x1; /* our MFT note */

  for(size_t i = 0; i < sizeof(Elf64_Ehdr); i++){
    printf("%c", ((char *)&hdr)[i]);
  }

  /* MFT note */
  Elf64_Phdr phdr = {0};

  phdr.p_type = PT_NOTE;

  /* ensure that p_filesz - offsetof(mft_n_h, m) >= 0 so the malloc(note) won't fail: */
  phdr.p_filesz = sizeof(struct mft_note_header) + 4;
  /* +5 would cause note_size and mft_size to be 1, and so on. */

  phdr.p_offset = sizeof(Elf64_Ehdr) + sizeof(Elf64_Phdr); /* pread_in_full */

  for(size_t i = 0; i < sizeof(Elf64_Phdr); i++){
    printf("%c", ((char *)&phdr)[i]);
  }

  struct mft_note_header mftnh = {
  /* strncmp(nhr.name, SOLO5_NOTE_NAME, ..) == 0 */
    .name = SOLO5_NOTE_NAME,
  /*  nhdr.type = SOLO5_NOTE_MANIFEST */
    .type = SOLO5_NOTE_MANIFEST,
  /* nhdr.descs <= MFT_NOTE_MAX_SIZE */
    .descsz = 0
  };
  
  for(size_t i = 0; i < sizeof(mftnh); i++){
    printf("%c", ((char *)&mftnh)[i]);
  }  
};
```

```shell
~/solo5/ cc mft_entries.c -o gen && ./gen > trigger.spt && solo5-spt ./trigger.spt
```

mato:
@cfcs found multiple issues in our ELF loader (#382, #383). This code is quite old and has been in need of review for some time, so here is a first cut of refactoring along with fixes for the reported issues.

Fixes #382, fixes #383, fixes #384.

### DIFF
--- a/tenders/common/elf.c
+++ b/tenders/common/elf.c
@@ -131,93 +131,98 @@ void elf_load(const char *file, uint8_t *mem, size_t mem_size,
 {
     int fd_kernel = -1;
     ssize_t nbytes;
-    size_t ph_size;
-    Elf64_Off ph_off;
-    Elf64_Half ph_entsz;
-    Elf64_Half ph_cnt;
-    Elf64_Half ph_i;
     Elf64_Phdr *phdr = NULL;
-    Elf64_Ehdr hdr;
-
-    /* elf entry point (on physical memory) */
-    *p_entry = 0;
-    /* highest byte of the program (on physical memory) */
-    *p_end = 0;
+    Elf64_Ehdr *ehdr = NULL;
+    Elf64_Addr e_entry;                 /* Program entry point */
+    Elf64_Addr e_end;                   /* Highest memory address occupied */
 
     fd_kernel = open(file, O_RDONLY);
     if (fd_kernel == -1)
         goto out_error;
 
-    nbytes = pread_in_full(fd_kernel, &hdr, sizeof(Elf64_Ehdr), 0);
+    ehdr = malloc(sizeof(Elf64_Ehdr));
+    if (ehdr == NULL)
+        goto out_error;
+    nbytes = pread_in_full(fd_kernel, ehdr, sizeof(Elf64_Ehdr), 0);
     if (nbytes < 0)
         goto out_error;
     if (nbytes != sizeof(Elf64_Ehdr))
         goto out_invalid;
-    if (!ehdr_is_valid(&hdr))
+    if (!ehdr_is_valid(ehdr))
         goto out_invalid;
+    /*
+     * e_entry must be non-zero and within range of our memory allocation.
+     */
+    if (ehdr->e_entry == 0 || ehdr->e_entry >= mem_size)
+        goto out_invalid;
+    e_entry = ehdr->e_entry;
 
-    ph_off = hdr.e_phoff;
-    ph_entsz = hdr.e_phentsize;
-    ph_cnt = hdr.e_phnum;
-    ph_size = ph_entsz * ph_cnt;
+    size_t ph_size = ehdr->e_phnum * ehdr->e_phentsize;
     phdr = malloc(ph_size);
     if (!phdr)
         goto out_error;
-    nbytes = pread_in_full(fd_kernel, phdr, ph_size, ph_off);
+    nbytes = pread_in_full(fd_kernel, phdr, ph_size, ehdr->e_phoff);
     if (nbytes < 0)
         goto out_error;
     if (nbytes != ph_size)
         goto out_invalid;
 
     /*
-     * Load all segments with the LOAD directive from the elf file at offset
-     * p_offset, and copy that into p_addr in memory. The amount of bytes
-     * copied is p_filesz.  However, each segment should be given
-     * p_memsz aligned up to p_align bytes on memory.
+     * Load all program segments with the PT_LOAD directive.
      */
-    for (ph_i = 0; ph_i < ph_cnt; ph_i++) {
-        uint8_t *daddr;
-        uint64_t _end;
-        size_t offset = phdr[ph_i].p_offset;
-        size_t filesz = phdr[ph_i].p_filesz;
-        size_t memsz = phdr[ph_i].p_memsz;
-        uint64_t paddr = phdr[ph_i].p_paddr;
-        uint64_t align = phdr[ph_i].p_align;
-        uint64_t result;
-        int prot;
+    e_end = 0;
+    for (Elf64_Half ph_i = 0; ph_i < ehdr->e_phnum; ph_i++) {
+        Elf64_Addr p_vaddr = phdr[ph_i].p_vaddr;
+        Elf64_Xword p_filesz = phdr[ph_i].p_filesz;
+        Elf64_Xword p_memsz = phdr[ph_i].p_memsz;
+        Elf64_Xword p_align = phdr[ph_i].p_align;
+        Elf64_Addr p_vaddr_end;
 
         if (phdr[ph_i].p_type != PT_LOAD)
             continue;
 
-        if ((paddr >= mem_size) || add_overflow(paddr, filesz, result)
-                || (result >= mem_size))
+        if ((p_vaddr >= mem_size)
+                || add_overflow(p_vaddr, p_filesz, p_vaddr_end)
+                || (p_vaddr_end >= mem_size))
             goto out_invalid;
-        if (add_overflow(paddr, memsz, result) || (result >= mem_size))
+        /* 
+         * Each segment should be given p_memsz aligned up to p_align bytes of
+         * memory, compute this as p_vaddr_end.
+         */
+        if (add_overflow(p_vaddr, p_memsz, p_vaddr_end)
+                || (p_vaddr_end >= mem_size))
             goto out_invalid;
         /*
-         * Verify that align is a non-zero power of 2 and safely compute
-         * ((_end + (align - 1)) & -align).
+         * Verify that p_align is a non-zero power of 2 and safely compute
+         * ((p_vaddr_end + (p_align - 1)) & -p_align).
          */
-        if (align > 0 && (align & (align - 1)) == 0) {
-            if (add_overflow(result, (align - 1), _end))
+        if (p_align > 0 && (p_align & (p_align - 1)) == 0) {
+            if (add_overflow(p_vaddr_end, (p_align - 1), p_vaddr_end))
                 goto out_invalid;
-            _end = _end & -align;
+            p_vaddr_end = p_vaddr_end & -p_align;
         }
-        else {
-            _end = result;
-        }
-        if (_end > *p_end)
-            *p_end = _end;
+        /*
+         * Keep track of the highest byte of memory occupied by the program.
+         */
+        if (p_vaddr_end > e_end)
+            e_end = p_vaddr_end;
 
-        daddr = mem + paddr;
-        nbytes = pread_in_full(fd_kernel, daddr, filesz, offset);
+        /*
+         * Load the segment into host memory at host_vaddr and ensure any BSS
+         * (p_memsz - p_filesz) is initialised to zero.
+         */
+        uint8_t *host_vaddr = mem + p_vaddr;
+        nbytes = pread_in_full(fd_kernel, host_vaddr, p_filesz,
+                phdr[ph_i].p_offset);
         if (nbytes < 0)
             goto out_error;
-        if (nbytes != filesz)
+        if (nbytes != p_filesz)
             goto out_invalid;
-        memset(daddr + filesz, 0, memsz - filesz);
+        if (p_memsz < p_filesz)
+            goto out_invalid;
+        memset(host_vaddr + p_filesz, 0, p_memsz - p_filesz);
 
-        prot = PROT_NONE;
+        int prot = PROT_NONE;
         if (phdr[ph_i].p_flags & PF_R)
             prot |= PROT_READ;
         if (phdr[ph_i].p_flags & PF_W)
@@ -229,27 +234,31 @@ void elf_load(const char *file, uint8_t *mem, size_t mem_size,
                   file, ph_i);
             goto out_invalid;
         }
-        if (mprotect(daddr, _end - paddr, prot) == -1)
+        if (mprotect(host_vaddr, p_vaddr_end - p_vaddr, prot) == -1)
             goto out_error;
     }
 
-    free (phdr);
-    close (fd_kernel);
-    *p_entry = hdr.e_entry;
+    free(ehdr);
+    free(phdr);
+    close(fd_kernel);
+    *p_entry = e_entry;
+    *p_end = e_end;
     return;
 
 out_error:
     warn("%s", file);
-    free (phdr);
+    free(ehdr);
+    free(phdr);
     if (fd_kernel != -1)
-        close (fd_kernel);
+        close(fd_kernel);
     exit(1);
 
 out_invalid:
     warnx("%s: Exec format error", file);
-    free (phdr);
+    free(ehdr);
+    free(phdr);
     if (fd_kernel != -1)
-        close (fd_kernel);
+        close(fd_kernel);
     exit(1);
 }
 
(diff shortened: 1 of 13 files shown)


tenders/common/elf.c
```diff
@@ -73,120 +74,155 @@ static ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
...
+        if (p_vaddr_end > e_end)
+            e_end = p_vaddr_end;
 
-        daddr = mem + paddr;
-        nbytes = pread_in_full(fd_kernel, daddr, filesz, offset);
+        /*
+         * Load the segment into host memory at host_vaddr and ensure any BSS
+         * (p_memsz - p_filesz) is initialised to zero.
+         */
+        uint8_t *host_vaddr = mem + p_vaddr;
```
cfcs: 1) if `mem_size` is large (in turn permitting a large `p_vaddr`), and `mem` (allocated by the caller) is near the end of the address space, doesn't this risk overflowing?

2) Is it intentional that the start address is not adjusted for alignment, and that the only influence that `p_align` has is on `p_vaddr_end` (and in turn for the range for the `mprotect()` call)?

tenders/common/elf.c
```diff
@@ -73,120 +74,155 @@ static ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
...
-            if (add_overflow(result, (align - 1), _end))
+        if (p_align > 0 && (p_align & (p_align - 1)) == 0) {
+            if (add_overflow(p_vaddr_end, (p_align - 1), p_vaddr_end))
                 goto out_invalid;
-            _end = _end & -align;
-        }
-        else {
-            _end = result;
+            p_vaddr_end = p_vaddr_end & -p_align;
         }
```
cfcs: I'm not 100% sure I'm following the bit hacking here, but it seems to me that if `p_align` is a multiple of 2 then: 
```c
Elf64_Addr  p_vaddr_end = ((p_vaddr + p_memsz) + (p_align -1)) & (-p_align);
```
It has been verified previously that `p_vaddr + p_filesz < mem_size` and that `p_vaddr + p_memsz < mem_size`, but I'm wondering if a large `p_align` can lead us change the memory protection flags of a range that goes beyond `mem_size` ?

tenders/common/elf.c
mato: > If mem_size is large (in turn permitting a large p_vaddr), and mem (allocated by the caller) is near the end of the address space, doesn't this risk overflowing?

I don't see how the caller could allocate a valid memory range (mem, memsize) where (mem + memsize) overlaps the address space. In any case, that would seem to be the caller's problem, not the ELF loader's?

> Is it intentional that the start address is not adjusted for alignment, and that the only influence that p_align has is on p_vaddr_end (and in turn for the range for the mprotect() call)?

No idea, I should re-read that part of the ELF spec.

tenders/common/elf.c
mato: Good question. Also not obvious if we shouldn't be doing the `memset()` using the size adjusted for alignment. WIll check.

tenders/common/elf.c
cfcs: 1) Yes, that is the caller's problem. But `mem_size` comes from the user, and `mem` is allocated by `malloc`, so I wonder if we should have something to handle this in place somewhere? (I don't know if we do). It seems like a failure mode that would be worth either checking for here, or documenting in `elf.h`?

tenders/common/elf.c
```diff
@@ -73,120 +74,155 @@ static ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
...
+        Elf64_Addr p_vaddr_end;
 
         if (phdr[ph_i].p_type != PT_LOAD)
             continue;
 
-        if ((paddr >= mem_size) || add_overflow(paddr, filesz, result)
-                || (result >= mem_size))
+        if ((p_vaddr >= mem_size)
+                || add_overflow(p_vaddr, p_filesz, p_vaddr_end)
+                || (p_vaddr_end >= mem_size))
```
cfcs: Hmm, re-reading this, shouldn't it be `> mem_size` instead of `>= mem_size`?
Consider the case where someone has a mapping of 2MB - or a single byte, for simplicity:
```c
p_vaddr = 0;
p_filesz = p_memsz = 1;
mem_size = 1;
/* it follows that */
!( p_vaddr >= mem_size ); /* good, within range */
p_vaddr_end = 0 + 1; /* good, no overflow */
p_vaddr_end >= 1; /* rejected since p_vaddr_end == 1 */
```
Now p_vaddr_end is

mato: Done in abeca34.

tenders/common/elf.c
mato: (1) Done in abeca34 using `assert()`s since this is a caller error.

mato:
@cfcs Ok, I've done another pass in abeca34. This makes the ELF loader even more strict -- see the commit message for details. I've also implemented adjusting the segment start address for alignment.

I also deliberately disallow overlapping segments in the ELF loader. As a side effect of this, I found that clang (FreeBSD) was generating such an executable for `test_tls` due to missing TLS sections in our linker script, so I fixed that -- I think -- at least enough to make the test pass. I'm still not convinced that we are actually doing enough/the right thing regarding TLS but don't want to get too distracted by that particular rabbit hole right now.

Please review.

mato:
Note that the OpenBSD test is actually successful, but is showing as pending due to random surf- build/node.js brokenness.

tenders/common/elf.c
```diff
@@ -73,120 +75,237 @@ static ssize_t pread_in_full(int fd, void *buf, size_t count, off_t offset)
...
+         * Memory protection flags should be applied to the aligned address
+         * range (p_vaddr_start .. p_vaddr_end). Before we apply them, also
+         * verify that the address range is aligned to the architectural page
+         * size.
+         */
+        if (p_vaddr_start & (EM_PAGE_SIZE - 1))
+            goto out_invalid;
+        if (p_vaddr_end & (EM_PAGE_SIZE - 1))
+            goto out_invalid;
+        host_vaddr = mem + p_vaddr_start;
```
cfcs: I find the reuse of this variable (`host_vaddr`) a bit confusing.
In the blocks above, `host_vaddr = mem + p_vaddr` is being used to load `p_filesz` bytes (unaligned).
Then additional stuff after that is `memset` up until length `mem + p_memsz`.

In this line here it is reassigned as `host_vaddr = mem + p_vaddr_start` (aligned, rounded down), after which the mapping between `mem + v_addr_start` and `mem + align_up(p_memsz, p_align)` is `mprotect()'ed` using the desired flags.
What is the intended contents of `mem + align_down(p_vaddr, p_align)` through `mem + p_vaddr` at this point? Should that also be cleared, or is the caller expected to have cleared `mem ... mem+mem_size`?

mato: > I find the reuse of this variable (host_vaddr) a bit confusing.

Fair enough. How about I introduce a `host_vaddr_start` for the second block?

> What is the intended contents of mem + align_down(p_vaddr, p_align) through mem + p_vaddr at this point?

The [ELF specification](http://refspecs.linuxfoundation.org/elf/elfspec.pdf) discusses this on page 2-8:

> ‘‘Impurities’’ in the other three pages are not logically part of the process image; whether the system expunges them is unspecified.

I.e. we don't care. In practice, given that the host memory comes from `mmap()` it will on all modern systems be initialised to zero.

Issue #303, opened 2018-12-11
hvt: Enforce non-executable code (W^X)


mato:
As part of general hardening, we should enforce both in the ELF loader and at the guest-physical to host-virtual translation layer that any executable code in the guest is not also writable (W^X). This has been in-place and known to work for some time on Linux/KVM, but not actively enforced. Manual testing on FreeBSD/vmm shows that this is _not_ functional there (see third point below).

There are several sub-issues to this:

1. `hvt_elf_load()` currently warns if a P`T_LOAD` PHDR requests both `PF_X` and `PF_W`. This should be changed to a fatal error instead.
2. `hvt_elf_load()` should enforce that `PT_LOAD` PHDRs cannot overlap, otherwise an attacker could construct a malicious ELF defeating the logic in the previous point. While overlapping `PT_LOAD` PHDRs are not explicitly mentioned in the ELF specification, there is no reason to support this behaviour. This check can be further simplified by enforcing that `PT_LOAD` PHDRs must be sorted by `p_vaddr` in the ELF file, which is within the [specification](http://refspecs.linuxbase.org/elf/gabi4+/ch5.pheader.html).
3. `hvt_elf_load()` currently applies `PF_x` flags from a `PT_LOAD` PHDR by calling `mprotect()`. This results in the correct behaviour (enforcement) _only_ on Linux/KVM which updates the guest-physical to host-virtual translation (i.e. EPT on x86_64) as expected. On FreeBSD/vmm, the ELF loader needs to issue a suitable `VM_MMAP_MEMSEG` call at the same point. On OpenBSD/vmm the situation is TBD.
4. A test case needs to be added, verifying that `.text` as seen by the guest is not writable.

ricarkol:
The warning for "if a PT_LOAD PHDR requests both PF_X and PF_W." was because of IncludeOS which creates a single XWR segment (at least when using the solo5 platform). Don't know if that is still happening. In any case, I think the IncludeOS guys are interested in W^X code, so this is worth fixing on that side.

@alfred-bratterud ^

adamsteen:
@mato OpenBSD will not allow an mprotect call with both write and execute enabled, W^X has been enforced at OS level since September 2016. I initially hit this in porting efforts.

See [OpenBSD Innovations](https://www.openbsd.org/innovations.html) **W^X**

mato:
> @mato OpenBSD will not allow an mprotect call with both write and execute 
[...]

PR #386, opened 2019-08-27
ELF loader: Generic NOTE loading

mato:
The ultimate motivation for this is to allow us to use different ELF notes for different purposes (see https://github.com/Solo5/solo5/issues/372#issuecomment-525272670). This will be useful for ABI versioning. From the commit log:

Make the NOTE loading functionality generic, rather than coupled to loading just a "MFT1" note. elf_load_mft(...) is now renamed as elf_load_note(...). The code is also intentionally decoupled from any dependency on mft_abi.h.

In the process, various NOTE structure alignment issues which were "papered over" in the initial iteration of this code have been fixed now that I have a correct understanding of how (and why) the structures are laid out.

Additional fixes:

- place .note.solo5.manifest in .data (as opposed to .rodata) for virtio and muen, as these bindings are expected to need to modify the structure in-place (as opposed to getting a copy from a tender).
- improve virtio built-in manifest initialisation, and introduce a global variable for access to the validated struct mft.
- document the ELF functions some more.

@cfcs I would very much appreciate your thorough review :-)

/cc @ricarkol (virtio changes)
/cc @Kensan (see above re muen placement of note structure)
/cc @Ehmry what about Genode? I presume the bindings there also need to modify mft in place, so you may need to account for that in your linker script. OR it may be easier just to take a writable copy of the struct?

### DIFF
--- a/include/solo5/mft_abi.h
+++ b/include/solo5/mft_abi.h
@@ -28,6 +28,7 @@
  */
 
 #include <stdbool.h>
+#include <stddef.h>
 #include <stdint.h>
 
 #ifndef MFT_ABI_H
@@ -67,7 +68,7 @@ struct mft_net_basic {
 #define MFT_NAME_MAX  67        /* Characters */
 
 /*
- * Manifest entry (mft.e[]).
+ * Manifest entry (struct mft.e[]).
  */
 struct mft_entry {
     char name[MFT_NAME_SIZE];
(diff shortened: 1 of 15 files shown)


include/solo5/mft_abi.h
```diff
@@ -99,55 +100,76 @@ struct mft {
...
+/*
+ * Defines an Elf64_Nhdr with n_name filled in and padded to a 4-byte boundary,
+ * i.e. the common part of a Solo5-owned Nhdr.
+ */
+struct mft1_nhdr {
+    uint32_t n_namesz;
+    uint32_t n_descsz;
+    uint32_t n_type;
+    char n_name[(sizeof(SOLO5_NOTE_NAME) + 3) & -4];
 };
```
cfcs: Since this is an on-disk wire format, would it make sense to have something here to guarantee that padding etc makes sense?
Self-alignment for the 32bit members makes sense, but it seems to me that whether or not padding is applied to the `struct mft1_nhdr` would depend on `SOLO5_NOTE_NAME`.
I'm thinking either adding `__attribute__((packed))` or
```c
_Static_assert((sizeof(struct mft1_ndr) == (4+4+4+8)), "mft1_nhdr alignment issue");
```
(or both?)

Same re: `mft1_note` below, ensuring that
```c
_Static_assert(sizeof(struct mft1_note) == sizeof(struct mft1_nhdr) + sizeof(struct mft),
        "struct mft1_note would be padded");
```
(because we have occurrences of `sizeof(struct mft1_note) - sizeof(struct mft1_nhdr)` which rests on the variant that there's no padding between the two)

include/solo5/mft_abi.h
```diff
@@ -99,55 +100,76 @@ struct mft {
...
+/*
+ * Defines the entire note (header, descriptor content).
+ */
+struct mft1_note {
+    struct mft1_nhdr h;
+    /*
+     * (m) will be aligned to a MFT_NOTE_ALIGN boundary.
+     */
     struct mft m;
 };
```
cfcs:
```suggestion
 };
_Static_assert(offsetof(struct mft1_note, m) == MFT1_NOTE_ALIGN,
        "(m) should be aligned to a MFT1_NOTE_ALIGN boundary");
```
Then we can get rid of the comment as well and ensure it statically, but it would require moving the definition of `MFT1_NOTE_ALIGN` above this struct definition.

include/solo5/mft_abi.h
```diff
@@ -99,55 +100,76 @@ struct mft {
...
 
 /*
- * Manifest ELF note headers.
+ * Maximum supported number of manifest entires.
+ */
+#define MFT_MAX_ENTRIES 64
```
cfcs: Would it make sense to make this `MFT1_MAX_ENTRIES` ?

include/solo5/mft_abi.h
```diff
@@ -99,55 +100,76 @@ struct mft {
...
-    struct mft_note_header h;
+/*
+ * Defines the entire note (header, descriptor content).
+ */
+struct mft1_note {
+    struct mft1_nhdr h;
+    /*
+     * (m) will be aligned to a MFT_NOTE_ALIGN boundary.
+     */
     struct mft m;
```
cfcs: Are we sure that we never want to change the `mft` format, i.e. should this also be `struct mft1`?

cfcs:
Re: writable `mft`: Is the `mft` exposed to the unikernel itself? If so, are we ensuring that the unikernel's mapping is read-only?

include/solo5/mft_abi.h
mato: Thanks for the suggestions. I've merged #388 which switches compilation to `-std=c11`, enabling us to use `_Static_assert()`.

The first `_Static_assert()` makes sense, though I would do:

```
_Static_assert((sizeof(struct mft1_nhdr)) ==
        (sizeof(uint32_t) + sizeof(uint32_t) + sizeof(uint32_t) + 8),
        "struct mft1_nhdr alignment issue");
```

The 2nd one will not pass, try it yourself. Dealing with the padding between `h` and `m` in a `struct mft1_note` is precisely why this change introduces `note_align`, `MFT1_NOTE_ALIGN` and related. Thus, `sizeof(struct mft1_note) - sizeof(struct mft1_nhdr)` must include the padding.

include/solo5/mft_abi.h
mato: > Then we can get rid of the comment as well and ensure it statically, but it would require moving the definition of MFT1_NOTE_ALIGN above this struct definition.

I don't follow, what does this gain us?

include/solo5/mft_abi.h
mato: > Would it make sense to make this MFT1_MAX_ENTRIES ?

Related:

> Are we sure that we never want to change the mft format, i.e. should this also be struct mft1?

I'm not sure. `struct mft` has its own inner versioning, and I'm hoping that we don't need to change it in an incompatible way. I'm also lazy to go changing `mft` to `mft1` throughout the codebase. But if you think it's worth it, I'll do that. Let me know.

mato:
> Re: writable mft: Is the mft exposed to the unikernel itself? If so, are we ensuring that the unikernel's mapping is read-only?

That depends on the backend. `hvt` will enforce `.text` and `.rodata` as read-only (but not non-executable, due to #303). `spt` will enforce all combinations. `virtio` enforces nothing as someone would have to write proper page table builders for it, not worth the work. `genode` and `muen`, not sure.

I should probably summarize the above in documentation before a release.

include/solo5/mft_abi.h
cfcs: I found it a bit odd that some of them dealing with MFTv1 were versioned and some not, but I think your rationale makes sense.

I think the general MFT structure seems extensible enough (`struct mft` with version, count, `mft_entry` VLA) - for some applications it might make sense to be able to adjust the `MAX_MFT_ENTRIES` constant, but that seems like a backwards compatible change going forward too, if need be.

To sum up, I think `struct mft_entry` seems like the most likely candidate for changes. Not sure either if it's worth changing everything for that at the moment.

cfcs:
@mato My concern was that if the tender and the unikernel are sharing this `struct mft` and the unikernel can write to it, the tender can't rely on the data being valid (in which case you might want to operate on a tender-only copy?)

include/solo5/mft_abi.h
cfcs: It would statically ensure what is currently asserted in the comment?

include/solo5/mft_abi.h
cfcs: Ah, okay, my bad!

include/solo5/mft_abi.h
mato: To do that we want the following, placed after the definition of MFT1_NOTE_ALIGN:

```
_Static_assert((offsetof(struct mft1_note, m) & (MFT1_NOTE_ALIGN - 1)) == 0,
        "struct mft1_note.m is not aligned to a MFT1_NOTE_ALIGN boundary");
```

Right?

mato:
@cfcs:

> My concern was that if the tender and the unikernel are sharing this struct mft and the unikernel can write to it, the tender can't rely on the data being valid (in which case you might want to operate on a tender-only copy?)

Tender-based targets don't share `struct mft`, and never operate on the built-in manifest in place, only on copies. The initial tender copy comes from `elf_load_note()` in `main()`. This is filled out during module setup and then again copied into the guest's "low memory" before launch, see `hvt_boot_info_init()` and `spt_boot_info_init()`.

include/solo5/mft_abi.h
cfcs: Yes, that would do the trick!

include/solo5/mft_abi.h
mato: > To sum up, I think struct mft_entry seems like the most likely candidate for changes. Not sure either if it's worth changing everything for that at the moment.

I agree. I do want to make some small changes to `struct mft_entry`, to allow for storing something other than a `hostfd`. Will do so in a separate PR, as that doesn't directly concern the NOTE loading.

mato:
@cfcs:

I've rebased onto latest master, added the suggested `_Static_asserts` we've discussed and made some minor cleanups to mft_abi.h to only define names under MFT[1].

I think this is good to go?

tenders/common/elf.c
```diff
@@ -384,81 +409,111 @@ void elf_load_mft(const char *file, struct mft **mft, size_t *mft_size)
...
-    note_offset = phdr[ph_i].p_offset + offsetof(struct mft_note, m);
-    note_size = phdr[ph_i].p_filesz - offsetof(struct mft_note, m);
-    note = malloc(note_size);
-    if (note == NULL)
+    assert(note_align > 0 && (note_align & (note_align - 1)) == 0);
+    note_offset = (sizeof nhdr + (note_align - 1)) & -note_align;
+    assert(note_offset >= sizeof nhdr);
+    note_pad = note_offset - sizeof nhdr;
+    note_size = nhdr.h.n_descsz - note_pad;
+    assert(note_size > 0);
```
cfcs: I believe this will always be true when `note_size != 0` since `note_size` is a `size_t`, so if it underflows it will be (very) large at this point. I didn't verify that such an underflow is achievable, but it seems reasonable given `nhdr.h.n_desc` can be e.g. `2`, and `note_align` is just required to be a non-zero power of 2?

If `!= 0` is the intention, we should have that instead.

But it looks to me like we might want to do the `nhdr.h.n_desc - note_pad` subtraction in a fashion that checks for underflow, since that would also guarantee that it's smaller than `max_note_size`?

mato: Got it. How about:

    assert(note_size != 0 && note_size <= nhdr.n_descsz)

?

cfcs: Perfect!

PR #370, opened 2019-06-07
[WIP] set hvt_guest_mprotect for "hvt: Enforce that guest executable code is not also writable (W^X) #303"

adamsteen:
Hi @mato and @hannesm 

just looking for feedback about the setup of hvt_guest_mprotect and elf_load.

PR #391, opened 2019-09-03
Introduce t_guest_mprotect() in ELF loader, fix hvt implementation

### LINKED ISSUE #390: hvt: Guest-executable pages should not be marked PROT_EXEC on the host
During discussion of interaction between host `mprotect()` and KVM behaviour, Sean Christopherson [pointed out to me](https://marc.info/?l=kvm&m=155872602111445&w=2) that our current implementation results in guest-executable pages also being executable on the host side. This is a security issue, since were the hvt tender compromised, it would make it much easier for an attacker to execute code in the tender process.

mato:
This allows for tender (spt, hvt) control over how page protection is applied to guest memory. This infrastructure is then used to ensure that guest-executable pages are not also host-executable.

Fixes #390.

/cc @cfcs #370 #303

### DIFF
--- a/tenders/common/elf.c
+++ b/tenders/common/elf.c
@@ -43,6 +43,7 @@
 #include <unistd.h>
 
 #include "cc.h"
+#include "elf.h"
 
 /*
  * Define EM_TARGET, EM_PAGE_SIZE and EI_DATA_TARGET for the architecture we
@@ -194,7 +195,8 @@ static int align_up(Elf64_Addr addr, Elf64_Xword align, Elf64_Addr *out_result)
 }
 
 void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
-       uint64_t p_min_loadaddr, uint64_t *p_entry, uint64_t *p_end)
+        uint64_t p_min_loadaddr, guest_mprotect_fn_t t_guest_mprotect, void *t,
+        uint64_t *p_entry, uint64_t *p_end)
 {
     ssize_t nbytes;
     Elf64_Phdr *phdr = NULL;
@@ -329,11 +331,6 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
             goto out_invalid;
         if (p_vaddr_end & (EM_PAGE_SIZE - 1))
             goto out_invalid;
-        uint8_t *host_vaddr_start = mem + p_vaddr_start;
-        /*
-         * Double check result for host (caller) address space overflow.
-         */
-        assert(host_vaddr_start >= (mem + p_min_loadaddr));
         int prot = PROT_NONE;
         if (phdr[ph_i].p_flags & PF_R)
             prot |= PROT_READ;
@@ -346,7 +343,8 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
                   bin_name, ph_i);
             goto out_invalid;
         }
-        if (mprotect(host_vaddr_start, p_vaddr_end - p_vaddr_start, prot) == -1)
+        assert(t_guest_mprotect != NULL);
+        if (t_guest_mprotect(t, p_vaddr_start, p_vaddr_end, prot) == -1)
             goto out_error;
     }
 
--- a/tenders/common/elf.c
+++ b/tenders/common/elf.c
@@ -195,8 +195,8 @@ static int align_up(Elf64_Addr addr, Elf64_Xword align, Elf64_Addr *out_result)
 }
 
 void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
-        uint64_t p_min_loadaddr, guest_mprotect_fn_t t_guest_mprotect, void *t,
-        uint64_t *p_entry, uint64_t *p_end)
+        uint64_t p_min_loadaddr, guest_mprotect_fn_t t_guest_mprotect,
+        void *t_guest_mprotect_arg, uint64_t *p_entry, uint64_t *p_end)
 {
     ssize_t nbytes;
     Elf64_Phdr *phdr = NULL;
@@ -344,7 +344,8 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
             goto out_invalid;
         }
         assert(t_guest_mprotect != NULL);
-        if (t_guest_mprotect(t, p_vaddr_start, p_vaddr_end, prot) == -1)
+        if (t_guest_mprotect(t_guest_mprotect_arg, p_vaddr_start, p_vaddr_end,
+                    prot) == -1)
             goto out_error;
     }
 
--- a/tenders/common/elf.h
+++ b/tenders/common/elf.h
@@ -26,9 +26,21 @@
 #define COMMON_ELF_H
 
 /*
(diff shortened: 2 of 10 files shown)


tenders/common/elf.c
```diff
@@ -194,7 +195,8 @@ static int align_up(Elf64_Addr addr, Elf64_Xword align, Elf64_Addr *out_result)
 }
 
 void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
-       uint64_t p_min_loadaddr, uint64_t *p_entry, uint64_t *p_end)
+        uint64_t p_min_loadaddr, guest_mprotect_fn_t t_guest_mprotect, void *t,
```
cfcs: 1) Hmm, would you be okay with changing the name to be something a bit more specific, like `void *tender_handle`
 or `*t_guest_mprotect_handle` whatever that is easier to recognize at the call site of `t_guest_mprotect()`?

2) Later on this will be casted to one of:
    ```c
    struct hvt *hvt = t;
    struct spt *spt = t;
    ```
    I wonder if it would make sense with a target-specific pointer type so we avoid having `void *` casts passed around?

tenders/common/elf.h
```diff
@@ -25,10 +25,22 @@
...
+ * guest_mprotect_fn() is called by the ELF loader to request that the page
+ * protection flags (prot) as used by the system mprotect(), i.e. PROT_X from
+ * sys/mman.h be applied to the guest memory range (addr_start .. addr_end).
+ * (t) is the backend-specific argument.
+ *
+ * Returns 0 on success, -1 and errno set on failure.
+ */
+typedef int (*guest_mprotect_fn_t)(void *t, uint64_t addr_start,
+        uint64_t addr_end, int prot);
+
```
cfcs: This lacks a description of `t` (see previous comment).

tenders/hvt/hvt.h
```diff
@@ -107,6 +107,13 @@ void hvt_vcpu_init(struct hvt *hvt, hvt_gpa_t gpa_ep);
 void hvt_boot_info_init(struct hvt *hvt, hvt_gpa_t gpa_kend, int cmdline_argc,
         char **cmdline_argv, struct mft *mft, size_t mft_size);
 
+/*
+ * Apply page protections to guest memory. See guest_mprotect_fn_t in elf.h for
+ * a full description.
+ */
+int hvt_guest_mprotect(void *t, uint64_t addr_start, uint64_t addr_end,
```
cfcs: Would it be tricky to just do `struct hvt *t` directly?
```suggestion
int hvt_guest_mprotect(struct hvt *t, uint64_t addr_start, uint64_t addr_end,
```

cfcs: It seems like `addr_start` and `addr_end` could additionally be const? (To decrease risk of confusing `vaddr_start` and `addr_start` when the former is assigned)

tenders/spt/spt_core.c
```diff
@@ -146,6 +146,28 @@ struct spt *spt_init(size_t mem_size)
...
+    assert(vaddr_start >= spt->mem);
+    size_t size = addr_end - addr_start;
+    assert(size > 0 && size <= spt->mem_size);
+
+    /*
+     * On spt, there is no distinction between host-side and guest-side memory
+     * protection, so just pass through to mprotect() directly, which will do
+     * the right thing.
+     */
+    return mprotect(vaddr_start, size, prot);
```
cfcs: Do we have use-cases for `W && X` mappings, or would it make sense to prevent the creation of such mappings?

tenders/common/elf.h
cfcs: It it correctly understood that in most cases `guest_mprotect_fn_t` primarily applies memory protection flags to the *host* view of the guest's memory mappings?

tenders/common/elf.c
mato: 1. I can do `void *t_guest_mprotect_arg`.
2. Will reply separately re function pointers.

tenders/common/elf.h
mato: > This lacks a description of t (see previous comment).

That's because I don't know what to call `t` in prose. "core tender struct" doesn't add any information.

> It it correctly understood that in most cases guest_mprotect_fn_t primarily applies memory protection flags to the host view of the guest's memory mappings?

Not quite. On KVM, at least the R/W part of the flags is applied also (indirectly) to the _guest_ mappings. On OpenBSD, if at some point if `vmm` gains support for `VMM_IOC_MPROTECT_EPT` then this function will also be responsible for directly manipulating the guest mappings via EPT.

tenders/hvt/hvt.h
mato: > It seems like addr_start and addr_end could additionally be const? (To decrease risk of confusing vaddr_start and addr_start when the former is assigned)

That doesn't work well in C for primitive types, since if you have e.g.

    int foo(const int bar);

then whatever `int` the _caller_ passed to `foo()` must also be `const` on the caller's side. So it's not very practical. Not worth it here.

tenders/spt/spt_core.c
mato: > Do we have use-cases for W && X mappings, or would it make sense to prevent the creation of such mappings?

Already done by the caller (ELF loader) at https://github.com/Solo5/solo5/blob/master/tenders/common/elf.c#L344, I don't want to do that check here.

mato:
> I wonder if there's a not-overly-complex way to avoid function pointers being passed around; they constitute a natural target for ROP-style exploits and pose challenges for control-flow integrity systems.

I agree that fiunction pointers are not great for the reasons you mention, but I don't see a way around them.

We're trying to keep the ELF loader decoupled from other code, and remember it's shared between different tenders, so the function has different types, expecting a `struct hvt` or `struct spt` depending on the tender.

> 1. I guess one option could be to write tuples of (`prot, addr_start , addr_end`) somewhere from `elf_load` and have a specialized `elf_load` wrapper for each target that looped over that list and applied the protections, but I'm not sure that's cleaner.

That's more code and passing data around, likely more complexity and bugs.

> 2. Another option could be to merge this as-is, and keep it in mind for any future restructuring of these things that we end up doing. Without a good solution I feel that is probably the way to go, but I wanted to leave my concern here in case you had a good idea for how to avoid the function pointer.

I prefer to merge and get the fix, rather than trying to find "clever" ways of avoiding the function pointer now. However, for the record, there is a third option:

We could use preprocessor magic and make the decision on what to call at compile time, thus avoiding the function pointer. However, we still have the issue of the argument `t_guest_mprotect_arg` which needs to be passed through `elf_load()`. So there would be quite a bit of `#ifdef` magic around _and_ in order to get the build system to grok different `elf.o`'s for hvt and spt, we would also have to `#include "../common/elf.c"` into a different module. So, doable, but not exactly the least-surprising way to do things.

tenders/spt/spt_core.c
cfcs: Ah, I had missed that that took place just above the call site in `elf_load`.

tenders/common/elf.h
cfcs: `t_arg` is fine with me, it's easier to search for and harder to confuse than `t`.

cfcs:
Cool, thanks for answering my questions. I say LGTM then :)

PR #394, opened 2019-09-06
ppc64: Fix syscall interface for syscall error case

stefanberger:
In case of a syscall error the syscall returns the (positive) errno as
the result and has the summary overflow flag in cr0 set. In our case we
set the result value to -errno (=-r3).

Signed-off-by: Stefan Berger <stefanb@linux.ibm.com>

### DIFF
--- a/bindings/spt/sys_linux_ppc64le.c
+++ b/bindings/spt/sys_linux_ppc64le.c
@@ -42,13 +42,17 @@ long sys_read(long fd, void *buf, long size)
     register long r3 __asm__("r3") = fd;
     register long r4 __asm__("r4") = (long)buf;
     register long r5 __asm__("r5") = size;
+    long cr;
 
     __asm__ __volatile__ (
-            "sc"
-            : "=r" (r3)
+            "sc\n\t"
+            "mfcr %1"
+            : "=r" (r3), "=&r" (cr)
             : "r" (r0), "r" (r3), "r" (r4), "r" (r5)
             : "memory", "cc"
     );
+    if (cr & CR0_SO)
+        r3 = -r3;
 
     return r3;
 }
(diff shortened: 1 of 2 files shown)


mato:
`solo5_yield()` and `solo5_net_read()` in bindings/spt/net.c both rely on `-errno` being returned by the syscall for `EAGAIN` and `EINTR`, so this will break things.

stefanberger:
> `solo5_yield()` and `solo5_net_read()` in bindings/spt/net.c both rely on `-errno` being returned by the syscall for `EAGAIN` and `EINTR`, so this will break things.

Fixed. In this case we need to negate the r3 using `r3 = -r3;` (for all syscalls).

bindings/spt/sys_linux_ppc64le.c
```diff
@@ -42,13 +42,17 @@ long sys_read(long fd, void *buf, long size)
     register long r3 __asm__("r3") = fd;
     register long r4 __asm__("r4") = (long)buf;
     register long r5 __asm__("r5") = size;
+    long cr;
 
     __asm__ __volatile__ (
-            "sc"
-            : "=r" (r3)
+            "sc\n\t"
```
mato: Nit: Superfluous `\n\t` here and in other occurrences of `sc`?

stefanberger: we need either `\n\t` or a ';' after the `sc`, otherwise we get:
```
CC spt/sys_linux_ppc64le.c
/tmp/ccDPEtcA.s: Assembler messages:
/tmp/ccDPEtcA.s:21: Error: unrecognized opcode: `scmfcr'
GNUmakefile:67: recipe for target 'spt/sys_linux_ppc64le.o' failed
make[1]: *** [spt/sys_linux_ppc64le.o] Error 1
GNUmakefile:69: recipe for target 'bindings' failed
```

Fixed a type in the header file, though...

mato:
> we need either \n\t or a ';' after the sc, otherwise we get:

Ok, thanks for the explanation.

PR #398, opened 2019-09-11
Rename mfttool to elftool, dumo manfest as JSON, add ability to dump ABI.

mato:
Rename `solo5-mfttool` to `solo5-elftool`.

Change `solo5-elftool dump` to use JSON as the format (e.g. for integration or orchestration systems). Add `solo5-elftool abi` to dump the ABI target and version from the ABI1 note.
/cc @cfcs @Kensan @hannesm


elftool/elftool.c
```diff
@@ -210,18 +213,66 @@ static int mfttool_dump(const char *binary)
...
+	default:
+	    return "unknown";
+    }
+}
+
+static int elftool_abi(const char *binary)
+{
+    int bin_fd = open(binary, O_RDONLY);
+    if (bin_fd == -1)
+    if (bin_fd == -1)
```
hannesm: looks like a duplicate line `if (bin_fd == -1)`...

elftool/elftool.c
```diff
@@ -210,18 +213,66 @@ static int mfttool_dump(const char *binary)
...
+        err(1, "%s: Could not open", binary);
+
+    struct abi1_info *abi1;
+    size_t abi1_size;
+    if (elf_load_note(bin_fd, binary, ABI1_NOTE_TYPE, ABI1_NOTE_ALIGN,
+		ABI1_NOTE_MAX_SIZE, (void **)&abi1, &abi1_size) == -1)
+        errx(1, "%s: No Solo5 ABI information found in executable",
+                binary);
+    printf("ABI target: %s\n", abi_target_to_string(abi1->abi_target));
+    printf("ABI version: %u\n", abi1->abi_version);
```
hannesm: if this is to-be-used by another program, it would imho make sense to output structured data.

elftool/elftool.c
mato: Sorry about that. Will fix in the next iteration of changes.

elftool/elftool.c
```diff
@@ -210,18 +213,66 @@ static int mfttool_dump(const char *binary)
...
+	printf("        { \"name\": \"%s\", \"type\": \"%s\" }%s\n",
+		mft->e[i].name, mft_type_to_string(mft->e[i].type),
+		(i == (mft->entries -1) ? "" : ","));
     }
 
+    printf("    ]\n"
+	   "}\n");
+
     free(mft);
     close(bin_fd);
```
cfcs: I guess we can `close(bin_fd);` immediately after `elf_load_note()`, then we don't have to keep it around for the rest of the function (and the close on line `211` becomes redundant).

elftool/elftool.c
```diff
@@ -210,18 +213,66 @@ static int mfttool_dump(const char *binary)
...
+    size_t abi1_size;
+    if (elf_load_note(bin_fd, binary, ABI1_NOTE_TYPE, ABI1_NOTE_ALIGN,
+		ABI1_NOTE_MAX_SIZE, (void **)&abi1, &abi1_size) == -1)
+        errx(1, "%s: No Solo5 ABI information found in executable",
+                binary);
+    printf("ABI target: %s\n", abi_target_to_string(abi1->abi_target));
+    printf("ABI version: %u\n", abi1->abi_version);
+    
+    free(abi1);
+    close(bin_fd);
```
cfcs: Same here, we can `close(bin_fd)` immediately after `elf_load_note()` here as well (which doesn't close it before `errx()` here, but that's just stylishness I guess)

PR #402, opened 2019-09-16
Implement Multidevice support for Muen

Kensan:
This updates the Muen bindings wrt. the recent API changes and implements multi-device support for network devices (see issue #372). Block devices are still not supported. Only the necessary API changes have been done to the dummy blockdev implementation.

### DIFF
--- a/bindings/muen/muen-block.c
+++ b/bindings/muen/muen-block.c
@@ -18,23 +18,31 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-#include "../bindings.h"
+#include "../hvt/bindings.h"
 
-solo5_result_t solo5_block_write(solo5_off_t offset __attribute__((unused)),
-                         const uint8_t *buf __attribute__((unused)),
-                         size_t size __attribute__((unused)))
+solo5_result_t solo5_block_acquire(const char *name __attribute__((unused)),
+        solo5_handle_t *handle __attribute__((unused)),
+        struct solo5_block_info *info __attribute__((unused)))
 {
     return SOLO5_R_EUNSPEC;
 }
 
-solo5_result_t solo5_block_read(solo5_off_t offset __attribute__((unused)),
-                        uint8_t *buf __attribute__((unused)),
-                        size_t size __attribute__((unused)))
+solo5_result_t solo5_block_write(solo5_handle_t handle __attribute__((unused)),
+        solo5_off_t offset __attribute__((unused)),
+        const uint8_t *buf __attribute__((unused)),
+        size_t size __attribute__((unused)))
 {
     return SOLO5_R_EUNSPEC;
 }
 
-void solo5_block_info(struct solo5_block_info *info __attribute__((unused)))
+solo5_result_t solo5_block_read(solo5_handle_t handle __attribute__((unused)),
+        solo5_off_t offset __attribute__((unused)),
+        uint8_t *buf __attribute__((unused)),
+        size_t size __attribute__((unused)))
 {
-    assert(0);
+    return SOLO5_R_EUNSPEC;
+}
+
+void block_init(struct hvt_boot_info *bi __attribute__((unused)))
+{
 }
--- a/bindings/muen/muen-block.c
+++ b/bindings/muen/muen-block.c
@@ -18,7 +18,7 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-#include "../hvt/bindings.h"
+#include "bindings.h"
 
 solo5_result_t solo5_block_acquire(const char *name __attribute__((unused)),
         solo5_handle_t *handle __attribute__((unused)),
--- a/bindings/muen/muen-net.c
+++ b/bindings/muen/muen-net.c
@@ -126,7 +126,7 @@ bool muen_net_pending_data(solo5_handle_t handle)
                                          &net_devices[handle].net_rdr);
 }
 
-void generate_mac_addr(uint8_t *addr)
+static void generate_mac_addr(uint8_t *addr)
 {
     const char *subject_name = muen_get_subject_name();
     uint64_t data;
--- a/bindings/muen/muen-net.c
+++ b/bindings/muen/muen-net.c
@@ -18,6 +18,7 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include "mft_abi.h"
 #include "../hvt/bindings.h"
 #include "sinfo.h"
 #include "reader.h"
@@ -31,15 +32,31 @@ struct net_msg {
     uint16_t length;
 } __attribute__((packed));
 
-static struct muchannel *net_in;
-static struct muchannel *net_out;
-static struct muchannel_reader net_rdr;
+struct muen_net_device {
+    bool acquired;
+    uint8_t mac_addr[SOLO5_NET_ALEN];
+    struct muchannel *net_in;
+    struct muchannel *net_out;
+    struct muchannel_reader net_rdr;
+};
 
-static uint8_t mac_addr[6];
+extern struct mft *muen_manifest;
 
-solo5_result_t solo5_net_write(const uint8_t *buf, size_t size)
+static struct muen_net_device net_devices[MFT_MAX_ENTRIES];
+
+/**
+ * Initialize Muen network device with given name.
+ */
+static bool muen_net_dev_init(const char *name, struct muen_net_device *device,
+        struct solo5_net_info *info);
+
+solo5_result_t solo5_net_write(solo5_handle_t handle,
+        const uint8_t *buf, size_t size)
 {
     struct net_msg pkt;
+
+    if (handle >= MFT_MAX_ENTRIES || !net_devices[handle].acquired)
+        return SOLO5_R_EINVAL;
 
     if (size > PACKET_SIZE)
         return SOLO5_R_EINVAL;
@@ -48,20 +65,25 @@ solo5_result_t solo5_net_write(const uint8_t *buf, size_t size)
     cc_barrier();
     pkt.length = size;
     memcpy(&pkt.data, buf, size);
-    muen_channel_write(net_out, &pkt);
+    muen_channel_write(net_devices[handle].net_out, &pkt);
 
     return SOLO5_R_OK;
 }
 
-solo5_result_t solo5_net_read(uint8_t *buf, size_t size, size_t *read_size)
+solo5_result_t solo5_net_read(solo5_handle_t handle,
+        uint8_t *buf, size_t size, size_t *read_size)
 {
     enum muchannel_reader_result result;
     struct net_msg pkt;
 
+    if (handle >= MFT_MAX_ENTRIES || !net_devices[handle].acquired)
+        return SOLO5_R_EINVAL;
+
     if (size < PACKET_SIZE)
         return SOLO5_R_EINVAL;
 
-    result = muen_channel_read(net_in, &net_rdr, &pkt);
+    result = muen_channel_read(net_devices[handle].net_in,
+                               &net_devices[handle].net_rdr, &pkt);
     if (result == MUCHANNEL_SUCCESS) {
         memcpy(buf, &pkt.data, pkt.length);
         *read_size = pkt.length;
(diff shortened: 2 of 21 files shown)


bindings/muen/muen-block.c
```diff
@@ -18,23 +18,31 @@
...
+        struct solo5_block_info *info __attribute__((unused)))
 {
     return SOLO5_R_EUNSPEC;
 }
 
-solo5_result_t solo5_block_read(solo5_off_t offset __attribute__((unused)),
-                        uint8_t *buf __attribute__((unused)),
-                        size_t size __attribute__((unused)))
+solo5_result_t solo5_block_write(solo5_handle_t handle __attribute__((unused)),
+		solo5_off_t offset __attribute__((unused)),
```
mato: Style: Indentation is messed up here.

bindings/muen/muen-net.c
```diff
@@ -18,6 +18,7 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include "../../include/solo5/mft_abi.h"
```
mato: It should be sufficient to just `#include "mft_abi.h"` here?

bindings/muen/muen-net.c
```diff
@@ -71,16 +92,38 @@ solo5_result_t solo5_net_read(uint8_t *buf, size_t size, size_t *read_size)
...
+            MFT_DEV_NET_BASIC, &mft_index);
+    if (mft_e == NULL)
+        return SOLO5_R_EINVAL;
+
+    handle = (solo5_handle_t)mft_index;
+
+    if (devices[handle].acquired) {
+        log(WARN, "Solo5: Net: Device '%s' already acquired\n", name);
+        return SOLO5_R_EINVAL;
+    }
```
mato: I see what you're doing here -- `solo5_net_acquire()` does the actual initialisation by calling `muen_net_dev_init()`, and errors if it's called more than once. This is slightly out of spec, in that I had intended the `_acquire_` APIs to allow for multiple calls.

Having said that, I can see how `_acquire_` implies a transition of some kind, so perhaps it'd be better to just change the docs (solo5.h) to say that you're only allowed to call it once per handle?

bindings/muen/muen-net.c
```diff
@@ -103,32 +146,36 @@ void generate_mac_addr(uint8_t *addr)
...
 {
-    char mac_str[18];
     const uint64_t epoch = muen_get_sched_start();
+    char mac_str[18];
+    char buffer[64];
+    snprintf(buffer, 64, "%s|out", name);
     const struct muen_resource_type *const
-        chan_out = muen_get_resource("net_out", MUEN_RES_MEMORY);
+        chan_out = muen_get_resource(buffer, MUEN_RES_MEMORY);
+	snprintf(buffer, 64, "%s|in", name);
```
mato: Style: Please use `sizeof buffer` instead of copying around `64`.

bindings/muen/muen-net.c
```diff
@@ -138,27 +185,39 @@ void net_init(void)
 
     if (!chan_in) {
         log(WARN, "Solo5: Net: No input channel\n");
```
mato: I suggest modifying this and other messages in this function to include the manifest device name, otherwise it'll be impossible to diagnose in the presence of multiple devices.

bindings/muen/muen-net.c
```diff
@@ -31,37 +32,57 @@ struct net_msg {
...
     pkt.length = size;
     memcpy(&pkt.data, buf, size);
-    muen_channel_write(net_out, &pkt);
+    muen_channel_write(devices[handle].net_out, &pkt);
 
     return SOLO5_R_OK;
 }
 
-solo5_result_t solo5_net_read(uint8_t *buf, size_t size, size_t *read_size)
+solo5_result_t solo5_net_read(solo5_handle_t handle __attribute__((unused)),
```
hannesm: the `handle` is not `unused` anymore AFAICT

Kensan:
Thanks for the review. The update should address all comments.

bindings/muen/muen-net.c
Kensan: I would have expected a compiler warning here... Thanks for pointing it out.

mato:
Working in parallel with @Kensan, added a few more cleanups in 7666df7 and a change to the bindings link order + linker scripts in e138267 to always locate `_start` at `TEXT_START` (0x100000), which suits Muen better.

bindings/muen/muen-net.c
```diff
@@ -103,62 +145,80 @@ void generate_mac_addr(uint8_t *addr)
     addr[0] |= 0x02;
 }
 
-void net_init(void)
+static bool muen_net_dev_init(const char *name, struct muen_net_device *device,
+        struct solo5_net_info *info)
 {
-    char mac_str[18];
     const uint64_t epoch = muen_get_sched_start();
```
cfcs: I guess we could check `device->acquired` here as well?

bindings/muen/muen-net.c
```diff
@@ -103,62 +145,80 @@ void generate_mac_addr(uint8_t *addr)
...
+	snprintf(buffer, sizeof buffer, "%s|in", name);
     const struct muen_resource_type *const
-        chan_in = muen_get_resource("net_in", MUEN_RES_MEMORY);
+        chan_in = muen_get_resource(buffer, MUEN_RES_MEMORY);
 
     if (!chan_out) {
-        log(WARN, "Solo5: Net: No output channel\n");
-        return;
+        log(WARN, "Solo5: Net: '%s': No output channel\n", name);
+        return false;
```
cfcs: Is it possible to end up in a state where `chan_out` is `NULL` but `chan_in` is allocated/leaked here?

bindings/muen/muen-net.c
```diff
@@ -103,62 +145,80 @@ void generate_mac_addr(uint8_t *addr)
...
-    log(INFO, "Solo5: Net: Input  channel @ 0x%llx, size 0x%llx\n",
+    device->net_in = (struct muchannel *)(chan_in->data.mem.address);
+    muen_channel_init_reader(&device->net_rdr, MUENNET_PROTO);
+    log(INFO, "Solo5: Net: '%s': Input  channel @ 0x%llx, size 0x%llx\n",
+        name,
         (unsigned long long)chan_in->data.mem.address,
         (unsigned long long)chan_in->data.mem.size);
 
-    generate_mac_addr(mac_addr);
+    generate_mac_addr(device->mac_addr);
```
cfcs: I would have expected `&device->mac_addr` here since `device { uint8_t mac_addr[6]; }`, and `generate_mac_addr(uint8_t *)`, but maybe that's just me not understanding C :)

bindings/muen/muen-net.c
```diff
@@ -103,62 +145,80 @@ void generate_mac_addr(uint8_t *addr)
...
     snprintf(mac_str, sizeof(mac_str), "%02x:%02x:%02x:%02x:%02x:%02x",
-             mac_addr[0], mac_addr[1], mac_addr[2],
-             mac_addr[3], mac_addr[4], mac_addr[5]);
-    log(INFO, "Solo5: Net: Using MAC address %s\n", mac_str);
+             device->mac_addr[0], device->mac_addr[1], device->mac_addr[2],
+             device->mac_addr[3], device->mac_addr[4], device->mac_addr[5]);
+    memcpy(info->mac_address, device->mac_addr, sizeof info->mac_address);
```
cfcs: `solo5_net_info` has `uint8_t mac_address[SOLO5_NET_ALEN]`, `device` has `uint8_t mac_addr[6]`.
Would it be reasonable to either use the `SOLO5_NET_ALEN` to define the size of the `mac_addr` field of the `muen_net_device` struct, or have a `_Static_assert()` verifying that they are equal?

bindings/muen/muen-platform_lifecycle.c
```diff
@@ -18,18 +18,37 @@
...
+
+    /*
+     * Get the built-in manifest "out of" the ELF NOTE and validate it. Note
+     * that the size must be adjusted from n_descsz to remove any internal
+     * alignment. Once validated, it is available for access globally by the
+     * bindings.
+     */
+    struct mft *mft = &__solo5_mft1_note.m;
+    size_t mft_size = __solo5_mft1_note.h.n_descsz -
+        (offsetof(struct mft1_note, m) - sizeof (struct mft1_nhdr));
```
cfcs: @mato do all backends have to have code like this, or is there a macro or something that could be used instead?

bindings/muen/muen-yield.c
```diff
@@ -18,23 +18,25 @@
...
-    bool rc = false;
+    solo5_handle_set_t tmp_ready_set = 0;
+
     do {
-        if (muen_net_pending_data()) {
-            rc = true;
-            break;
+        for (solo5_handle_t i = 0U; i < MFT_MAX_ENTRIES; ++i) {
+            if (muen_net_pending_data(i))
+                tmp_ready_set |= 1UL << i;
```
cfcs: While this is a neat trick, this only works when `MFT_MAX_ENTRIES <= sizeof(tmp_ready_set)`.
I am a little bit skeptical about hardcoding that invariant in too many places without either statically checking the `MFT_MAX_ENTRIES` constant or the ABI version, WDYT @mato?

bindings/muen/muen-net.c
```diff
@@ -71,16 +91,38 @@ solo5_result_t solo5_net_read(uint8_t *buf, size_t size, size_t *read_size)
...
 {
-    return muen_channel_has_pending_data(net_in, &net_rdr);
+    solo5_handle_t handle = 0;
+    unsigned mft_index;
+    struct mft_entry *mft_e = mft_get_by_name(muen_manifest, name,
+            MFT_DEV_NET_BASIC, &mft_index);
+    if (mft_e == NULL)
+        return SOLO5_R_EINVAL;
+
+    handle = (solo5_handle_t)mft_index;
```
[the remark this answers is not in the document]
mato: `solo5_handle_t` is an `uint64_t`, so this cast is not necessary.

bindings/muen/muen-net.c
mato: Ok, I'll formalise that in a separate PR.

bindings/muen/muen-net.c
mato: @cfcs In C a `char []` is the same as a `char *`, so the code is correct.

bindings/muen/muen-net.c
mato: I'd just re-use SOLO5_NET_ALEN, it's in `solo5.h` so won't go away.

bindings/muen/muen-platform_lifecycle.c
mato: @cfcs Only those that don't have tenders, the others get the mft via different means (`xxx_boot_info`). I'll see what I can do about adding a suitable macro to `mft_abi.h`.

bindings/muen/muen-net.c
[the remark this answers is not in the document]
Kensan: I actually took this code from Virtio.

bindings/muen/muen-net.c
Kensan: We could but it felt a bit redundant since the function is only called internally.

bindings/muen/muen-yield.c
mato: @cfcs We could do something like (not tested)

```
_Static_assert(MFT_MAX_ENTRIES <= (sizeof(solo5_ready_set_t) * 8), "MFT_MAX_ENTRIES too large")
```

...but I don't know where to put that, as it depends on _both_ `mft_abi.h` and `solo5.h`, the two of which are supposed to be independent.

bindings/muen/muen-net.c
Kensan: Yes, that is possible however I did not think it is too problematic since `acquired` will only be set if everything succeeds.
I can move all of the assignments to the end of the function after all checks for both channels have passed.

bindings/muen/muen-net.c
Kensan: Will use `SOLO5_NET_ALEN`.

bindings/muen/muen-net.c
```diff
@@ -31,37 +31,57 @@ struct net_msg {
...
+    struct muchannel *net_in;
+    struct muchannel *net_out;
+    struct muchannel_reader net_rdr;
+};
 
-static uint8_t mac_addr[6];
+extern struct mft *muen_manifest;
 
-solo5_result_t solo5_net_write(const uint8_t *buf, size_t size)
+static struct muen_net_device devices[MFT_MAX_ENTRIES];
```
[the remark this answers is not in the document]
Kensan: I had used `netdevs` but opted for `devices` since its in the `muen-net.c` file. Will update accordingly.

bindings/muen/muen-net.c
Kensan: Actually, disregard that last comment. At this point `acquired` has not been set yet. That only happens after everything succeeds at the end of `solo5_net_acquire`.

mato:
@cfcs I think @Kensan has addressed all the comments not addressed to me in the latest update, OK to merge it? Debian CI failure is the usual `slept too little`, unrelated. Travis CI stuck in orange is something to do with PPC, we can ignore that.

bindings/muen/muen-net.c
cfcs: So does that mean that we *should* check `acquired` here?

bindings/muen/muen-net.c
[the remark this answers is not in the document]
cfcs: I had missed that it was static in `muen-net.c`, I think it's fine as is.

cfcs:
Yes, LGTM :tada:

bindings/muen/muen-net.c
Kensan: Sorry for the confusing comments: no, I don't think it is necessary since the function is only used internally and `acquired` has not yet been set at this point. We could check for `!acquired` but not sure its worth it?

mato:
Thanks for all the review! All done here.

PR #404, opened 2019-09-18
manifest: Add extra assertions, enforce const-ness

mato:
As suggested by @cfcs in https://github.com/Solo5/solo5/pull/402#discussion_r325192686, and during earlier reviews.

### DIFF
--- a/tenders/common/mft.c
+++ b/tenders/common/mft.c
@@ -29,7 +29,7 @@
 
 #include "mft.h"
 
-int mft_validate(struct mft *mft, size_t mft_size)
+int mft_validate(const struct mft *mft, size_t mft_size)
 {
     /*
      * Note that this function is called early in the loading process by the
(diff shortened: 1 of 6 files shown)


tenders/common/mft.c
```diff
@@ -65,28 +65,37 @@ int mft_validate(struct mft *mft, size_t mft_size)
...
          * in the array itself.
          */
-        mft->e[i].name[MFT_NAME_MAX] = 0;
+        if (mft->e[i].name[MFT_NAME_MAX] != 0)
+            return -1;
         /*
          * Sanitize private fields (to be used by the tender/bindings):
          *
-         * 1. b.hostfd must be set to an invalid value, attached to false.
+         * 1. .attached must be false (which implies that .u is uninitialised)
```
cfcs: I guess it also implies that `mft->e[i].b.hostfd` is uninitialized?

tenders/common/mft.h
```diff
@@ -46,23 +46,28 @@ int mft_validate(struct mft *mft, size_t mft_size);
...
-void mft_get_builtin_mft1(struct mft1_note *note, struct mft **out_mft,
-        size_t *out_mft_size);
+void mft_get_builtin_mft1_unconst(const struct mft1_note *note,
+        struct mft **out_mft, size_t *out_mft_size);
+void mft_get_builtin_mft1(const struct mft1_note *note,
+        const struct mft **out_mft, size_t *out_mft_size);
 
 /*
  * Return the manifest entry matching (name), of type (type), or NULL if none
  * found.
```
cfcs: Not related to this PR, but I think it would make sense if the comment documenting this function mentioned that if `index != NULL`, and if an entry is found, then the index will be stored to `*index` (so you can cache it and use `mft_get_by_index` in future calls I guess?)

tenders/common/mft.c
mato: Yeah, all of `.b`, will add to comment. I had to remove setting `.b.hostfd = -1` to preserve constness.

tenders/common/mft.h
mato: I'll update the comment.

PR #409, opened 2019-09-19
Enable stack protector for tenders, elftool. Run tests with 2MB memory.

mato:


mato:
Travis is failing due to `ppc64le` build apt issues, merging anyway.

PR #413, opened 2019-09-20
don't warn when libseccomp is new enough

mjl-:
i've got 2.4.1, which should pass >=2.3.3.


mato:
Thanks, I obviously hadn't had enough coffee when I wrote that :-)

PR #420, opened 2019-09-26
test compilation failure on FreeBSD 12+

hannesm:
on FreeBSD-12 with clang 8.0.1 (and their ld), I get:
```
ld -nostdlib -z max-page-size=0x1000 -static  -T /usr/home/hannes/.opam/4.09.0/.opam-switch/build/solo5-bindings-hvt.0.6.2/bindings/hvt/solo5_hvt.lds /usr/home/hannes/.opam/4.09.0/.opam-switch/build/solo5-bindings-hvt.0.6.2/bindings/hvt/solo5_hvt.o test_tls.o manifest.o -o test_tls.hvt
ld: error: test_tls.o has an STT_TLS symbol but doesn't have an SHF_TLS section
```
when attempting to link test_tls

and it looks the "opam install" always triggers a testbuild...

this is a workaround which disables `test_tls` on FreeBSD

### DIFF
--- a/tests/test_tls/test_tls.c
+++ b/tests/test_tls/test_tls.c
@@ -49,7 +49,7 @@ static void puts(const char *s)
     solo5_console_write(s, strlen(s));
 }
 
-#if defined(__OpenBSD__)
+#if defined(__OpenBSD__) || defined(__FreeBSD__)
 /* __thread is not supported in OpenBSD (this test fails on it). */
 volatile uint64_t _data;
 #else


mato:
I'll look into it. Must be something that changed in between clang 8.0.0 -> 8.0.1 or 11.3-RELEASE -> 12 as I just updated the FreeBSD CI templates to 11.3-RELEASE and clang 8.0.0, and don't see this problem there.

mato:
@hannesm:

What FreeBSD version are you referring to here, exactly? A freshly upgraded 11.3-RELEASE has clang 8.0.0, and a freshly upgraded 12.0-RELEASE has clang 6.0.0. If you are referring to something that's in current, then I can't test that as it's a moving target.

hannesm:
I use these days the `stable/12` branch of FreeBSD (FreeBSD clang version 8.0.1 (tags/RELEASE_801/final 366581) (based on LLVM 8.0.1)), the same (8.0.1) was merged into `releng/12.1`, so it'll be included in the 12.1 release (according to the [roadmap](https://www.freebsd.org/releases/12.1R/schedule.html) rc1 was just built -- and it is [available on ftp.freebsd.org](http://ftp.freebsd.org/pub/FreeBSD/releases/ISO-IMAGES/12.1/) (below the beta3 images)).

I don't understand enough bats magic, but would propose to just disable the TLS test on FreeBSD (since there aren't any users thereof afaik).

mato:
I don't want to disable tests without having at least some idea about what is actually causing the problem. Will wait for 12.1-RELEASE to come out and then come back to this.

hannesm:
as an update here, [FreeBSD 12.1 is now released](https://www.freebsd.org/releases/12.1R/announce.html)

mato:
Superseded by #436.

PR #422, opened 2019-10-08
build: Use pkg-config to find libseccomp on Linux

### LINKED ISSUE #418: build: Fail meaningully if no targets can be built
OPAM installations use `make CONFIG_<TARGET>= ...` to de-select all but the target being packaged. This can result in cases where an external dependency such as libseccomp is not available, which causes a warning, de-selects the sole target remaining and the resulting `make opam-install-%` fails with an obscure error.

### LINKED ISSUE #419: Support Linux distributions which install seccomp.h in places other than /usr/include
OpenSUSE `libseccomp-devel` installs `seccomp.h` as `/usr/include/libseccomp/seccomp.h`, unlike every other distribution we've tested so far.

In order to support this, a bunch of surgery needs to be done on the build system to figure out the `libseccomp` location at `configure.sh` time (possibly via pkg-config, but maybe not) and we then need have to pass `-I` down to Makefiles via `HOSTCPPFLAGS`.

See also #418.

mato:
Some Linux distributions install libseccomp in non-standard directories.
Use pkg-config to look for it instead of guessing. Note that this
makes pkg-config a hard build-time dependency.

While we're at it, we can use pkg-config to do version tests, so make
libseccomp >= 2.3.3 a hard requirement and warn but proceed on
libseccomp < 2.4.1.

Fixes #419, fixes #418 (by eliminating the no-targets case).

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -62,7 +62,7 @@ EOM
 
 gcc_check_header()
 {
-    ${CC} -x c -o /dev/null - <<EOM >/dev/null 2>&1
+    ${CC} ${PKG_CFLAGS} -x c -o /dev/null - <<EOM >/dev/null 2>&1
 #include <$@>
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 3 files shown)


mato:
(CI is actually passing, with a dummy commit that I have now removed)

PR #433, opened 2019-11-12
Add support for "tools only" build

mato:
For use with Muen where we need a `solo5-elftool` around, add support for a "tools only" build and install via `./configure.sh --only-tools` and `make install-tools`.

/cc @Kensan


Kensan:
LGTM, works as advertised :)

PR #443, opened 2020-03-04
Stop host kernels from executing Solo5 binaries

### LINKED ISSUE #442: Solo5 executables should not be executable by a host kernel
If a user attempts to run a Solo5 executable directly (regardless of target), the host kernel will in most cases attempt to load and run the executable.

This is undesirable for several reasons:

1. Bad UX for a "naive or beginning user": Running such an executable directly produces `Segmentation fault` on Linux and `Bus error (core dumped)` on FreeBSD.
2. Accidentally attempting to directly run a unikernel from an untrusted source is a potential security risk, since the host kernel's ELF loader runs in kernel space.

I considered several solutions:

1. Follow the ELF spec: set `e_ident[EI_OSABI]` to `ELFOSABI_STANDALONE`. Unfortunately the Linux (but not FreeBSD) ELF loader ignores these values entirely and will attempt to execute any ELF file with a matching processor architecture. Even if that were not the case, without providing our own Solo5-specific compiler toolchain we have no easy way to control these values in the resulting binary.

2. Provide a "stub": similarly to how PE executables have a DOS stub, that would be executed by the host kernel instead of the Solo5 start-up code. This is very hard, if at all possible, since tenders would have no way of distinguishing the "stub" `PT_LOAD` segment from the "real" `PT_LOAD` segments that doesn't involve inventing out-of-spec schemes to distinguish the two.

3. Provide a `PT_INTERP` pointing to a non-existent path: on Linux, this produces `No such file or directory` when attempting to directly run such a binary. On FreeBSD, this produces `ELF interpreter /nonexistent/solo5 not found, error 2`.

I'm going to go with option (3) as this is the least complex and best option overall. While the Linux response of `No such file or directory` is not ideal, it's still better than the current `Segmentation fault`.

mato:
This is implemented for hvt, spt, muen and virtio bindings by:

- Adding a PT_INTERP and .interp in binding-specific linker scripts.
- Populating .interp via DECLARE_ELF_INTERP in binding-specific start.c.
- The content of PT_INTERP is "/nonexistent/solo5/".

The Genode bindings have not been changed, as they seem to do their own
thing regarding PT_INTERP.

With this change if a Solo5 unikernel binary is run directly (e.g.
./test_hello.hvt) on a Linux/FreeBSD/OpenBSD host, the host kernel will
not attempt to execute the binary and will fail with a more or less
sensible error message rather than produce a segmentation fault or
similar.

This provides both for a nicer "naive user" experience, and is a
security improvement against accidentally exposing the host kernel to
unikernel binaries from untrusted sources.

**TODO** before merging: Manually test targets that are not tested in CI:

1. [x] muen: Ensure that the Muen toolchain does not get confused by the presence of a `PT_INTERP`.
2. [x] virtio with byhve on FreeBSD.
3. [ ] virtio using Google Compute Engine.

Fixes #442.

### DIFF
--- a/bindings/hvt/solo5_hvt.lds
+++ b/bindings/hvt/solo5_hvt.lds
@@ -34,6 +34,7 @@ ENTRY(_start)
  * into a separate PT_NOTE header, we need to lay these out explicitly.
  */
 PHDRS {
+    interp PT_INTERP;
     text PT_LOAD FLAGS(5); /* No FILEHDR or PHDRS, force R/E only.
                               FLAGS values come from PF_x in elf.h */
     data PT_LOAD;
@@ -51,6 +52,10 @@ SECTIONS {
      * :text: The following input sections are placed in the R/E :text segment.
      */
     _stext = .;
+
+    .interp : {
+        *(.interp)
+    } :interp :text
 
     .text :
     {
--- a/bindings/hvt/start.c
+++ b/bindings/hvt/start.c
@@ -50,10 +50,14 @@ void _start(const void *arg)
 }
 
 /*
+ * Place the .interp section in this module, as it comes first in the link
+ * order.
+ */
+DECLARE_ELF_INTERP
+
+/*
  * The "ABI1" Solo5 ELF note is declared in this module.
  *
- * TODO: We will want a separate start.c and muen_abi.h for Muen bindings, but
- * this will work for now.
  */
 ABI1_NOTE_DECLARE_BEGIN
 {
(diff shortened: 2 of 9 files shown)


mato:
Updated to use `/nonexistent/solo5/` (trailing slash added) as discussed in the linked issue.

mato:
@yomimono I believe you are a user of Solo5/MirageOS on Google Compute Engine. As I don't currently have a straightforward way to test there, would you mind building any random MirageOS unikernel using Solo5/virtio off this branch and testing that it boots correctly? Given that both `qemu` and `bhyve` have no issue with the changes here I'd expect it to be 99% yes, but it won't hurt to confirm.

mato:
Commit amended to note that OpenBSD has a different behaviour, which we can do nothing about.

yomimono:
@mato - sorry, just saw this notification today.  I'm not using GCE for running unikernels anymore, but I can help someone through the process if anyone's interested in trying it at the hack retreat perhaps?

PR #447, opened 2020-04-08
add openbsd mprotect ept support

adamsteen:
This change add support for mprotect_ept ioctl in openbsd, but will only be enabled on an OpenBSD 6.7 system (just gone into beta)

the ifdefs check for VMM_IOC_MPROTECT_EPT, and call if available

Would like to get reviews, so we can add this when 6.7 is released.

### DIFF
--- a/tenders/hvt/hvt_openbsd.c
+++ b/tenders/hvt/hvt_openbsd.c
@@ -170,6 +170,7 @@ int hvt_guest_mprotect(void *t_arg, uint64_t addr_start, uint64_t addr_end,
         int prot)
 {
     struct hvt *hvt = t_arg;
+    int ret;
 
     assert(addr_start <= hvt->mem_size);
     assert(addr_end <= hvt->mem_size);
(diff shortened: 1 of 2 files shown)


mato:
@adamsteen Thanks. I presume that it's straightforward to install OpenBSD 6.7 beta from scratch in a VM? Or, if I want to `sysupgrade` an existing VM, is there anything special I need to do to get the 6.7 beta release?

adamsteen:
Yeah very straight forward! It’s a lot of hitting enter!

sysupgrade is only available from 6.6 onwards, so would need to upgrade from 6.4 to 6.5 and then to 6.6 from there, but once you got to “current” you shouldn’t/can’t go back to “release“/“stable“ 

Any snapshot from today onwards is a 6.7 beta image!

Sorry on my phone

tenders/hvt/hvt_openbsd.c
```diff
@@ -184,12 +185,35 @@ int hvt_guest_mprotect(void *t_arg, uint64_t addr_start, uint64_t addr_end,
...
+    vmep = calloc(1, sizeof (struct vm_mprotect_ept_params));
+    if (vmep == NULL) {
+        warn("calloc");
+        return -1;
+    }
+
+    vmep->vmep_vm_id = hvb->vcp_id;
+    vmep->vmep_vcpu_id = hvb->vcpu_id;
+
+    // sgpa = vmr->vmr_gpa(0x0) + (addr - vmr->vmr_va)
```
mato, line 209: Nit: This looks like a left-over comment from development?

tenders/hvt/hvt_openbsd_x86_64.c
```diff
@@ -190,6 +190,14 @@ int hvt_vcpu_loop(struct hvt *hvt)
                     hvt_gpa_t gpa = vei->vei.vei_data;
                     fn(hvt, gpa);
                     break;
+#if defined(VMM_IOC_MPROTECT_EPT)
```
mato, line 193: Minor: Is `VMX_EXIT_EPT_VIOLATION` dependent on the presence of `VMM_IOC_MPROTECT_EPT`? In other words, would this case be useful, and can it compile on, OpenBSD < 6.7? If yes then there's no need for the `#ifdef` here.

adamsteen: @mato kinda yes, they were both added in the same commit, if won't compile on OpenBSD < 6.7

tenders/hvt/hvt_openbsd.c
adamsteen: @mato this has been removed.

adamsteen:
@mato Would you like me to squash the commits?

mato:
> 
> 
> @mato Would you like me to squash the commits?

@adamsteen Yes, please. 

Also, can you add a note in the commit message that this is gated on VMM_IOC_MPROTECT_EPT, which was added in OpenBSD 6.7? That way, when someone comes to look at the code years later they know why the #ifdefs were done this way.

adamsteen:
@mato I think we are now all good to go!

mato:
@adamsteen Awesome, thanks!

PR #451, opened 2020-04-24
tests.bats: Various improvements

mato:
PR #447 introduces full W^X support for hvt on OpenBSD 6.7. This followup enables the corresponding tests on OpenBSD 6.7.

As part of this, add some abstractions for declarative skipping of tests and update the call sites to use them where appropriate.

/cc @adamsteen

PR #443 (issue #442) was missing any kind of tests, so add a basic test for this.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 #
-# THIS IS NOT THE BUILD SCRIPT FOR SOLO5.
+# THIS IS NOT THE BUILD SCRIPT FOR SOLO5. DO NOT RUN THIS SCRIPT MANUALLY
+# UNLESS YOU KNOW WHAT YOU ARE DOING, AND WHAT THE SCRIPT DOES.
 #
 # This script is used for automated builds with surf-build and Travis CI.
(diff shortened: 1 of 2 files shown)


mato:
I've added an OpenBSD 6.7 beta CI node to the mix, and it passes `wnox` and `xnow` as expected.

I've also made various improvements to tests.bats, mainly to make it easier to eyeball test results / CI runs -- among other things tests that are skipped due to the underlying target not being built in the fist place are now skipped without printing a reason. What I really want in this case is "pretend this test does not exist" but bats doesn't have that, so this will do for now.

mato:
Ok, so 3d460df was "fun". Whoever wrote the FreeBSD ELF loader must have thought "ELF interpreter ... not found" is a fatal error indeed, to [bother](http://fxr.watson.org/fxr/source/kern/imgact_elf.c?v=FREEBSD-12-STABLE#L1037) sending it to the ctty with `uprintf()` .

/cc @hannesm @sg2342

OpenBSD67 failure is transient, and can't be bothered waiting for Travis ppc64le which will just work, merging this as "good enough for now".

PR #453, opened 2020-04-27
FreeBSD/OpenBSD do not copy endian headers into crt

hannesm:
this reduces the header files from the operating system copied around, and synchronizes them (more or less).

it still feels a rather ad-hoc which system headers are copied around, the main rationale is "whatever is needed to compile the ocaml runtime system and openlibm". endian.* were around to compile nocrypto^Wmirage-crypto, but it turns out this is a bad idea.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -166,12 +166,11 @@ config_host_freebsd()
     [ "${CONFIG_ARCH}" = "x86_64" ] ||
         die "Only 'x86_64' is supported on FreeBSD"
     INCDIR=/usr/include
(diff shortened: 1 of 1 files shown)


mato:
Yes, this is a bug. The *BSD file-header-copying code has been around since the initial FreeBSD port in 2016 and is a bit fragile (tries to copy some files that don't exist on certain OS versions). I reviewed the git blame and commit history, and there's no clear reason for endian.h to be provided at this level.

Merging this, and I have a follow-up that cleans up the process in the pipeline.

PR #454, opened 2020-04-28
muen: Drop explicit FPU initialization

Kensan:
The MXCSR register is initialized to the default value [1] so there is
no need to explicitly set it.

NOTE: The ABI version is unchanged since there has not been a Solo5 (or
      Muen) release since it was raised to 2.

[1] - Muen commit f10bd6b


mato:
Thanks, LGTM. E2E error is random CI flakiness.

PR #455, opened 2020-04-29
configure.sh: Improve CRT header copying on BSDs

mato:
The *BSD C runtime header copying code is somewhat fragile and requires
manually keeping track of dependencies of each header that gets copied.

This change uses cc -M to gather a list of dependencies of all the
requested header files, and then cpio to copy the result including
intermediate directories to the target directory.

While I'm at it, remove osreldate.h from SRCS= on FreeBSD, since
as far as I can tell there's nothing that depends on that.

/cc @hannesm @adamsteen

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -87,6 +87,24 @@ ld_is_lld()
     ${LD} --version 2>&1 | grep -q '^LLD'
 }
 
+# Arguments: PATH, FILES...
+# For the header FILES..., all of which must be relative to PATH, resolve their
+# dependencies using the C preprocessor and output a list of FILES... plus all
+# their unique dependencies, also relative to PATH.
+get_header_deps()
+{
+    local path="$1"
+    shift
+    (
+        set -o pipefail &&
+        cd ${path} &&
+        ${CC} -M "$@" | \
+            sed -e 's!.*\.o:!!g' -e "s!${path}/!!g" | \
+            tr ' \\' '\n' | \
+            sort | uniq
+    )
+}
+
 config_host_linux()
 {
     # On Linux/gcc we use -nostdinc and copy all the gcc-provided headers.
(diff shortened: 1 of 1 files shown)


samoht:
AFAIK mirage-crypto also somehow needs `x86intrin.h`, `clzerointrin.h` and `cpuid.h`. Are you installing these too?

mato:
@samoht It's complicated:

1. On GCC-based systems: Solo5 uses `-nostdinc` which tells the compiler to not use _any_ standard include paths (intrinsics, system, or anything else). configure.sh copies the compiler headers _as a whole_ from the directory pointed to by `cc -print-file-name=include`, e.g. on my Debian system this is `/usr/lib/gcc/x86_64-linux-gnu/8/include`. This then contains `std*.h`, `float.h` and a bunch of the compiler intrinsics-specific headers `cpuid.h`, `*intrin.h`, etc.

2. On clang-based systems: Solo5 uses `-nostdlibinc` which tells the compiler to not use any system include paths, but still use its internal headers for intrinsics. We then copy only the `std*.h` and `float.h` from `/usr/include`, as unlike on GCC-based systems, on the BSDs these headers are _not_ shipped alongside the intrinsics headers.

It was years since I wrote this code, but I _believe_ the underlying reason for not copying the clang intrinsics headers as a whole (like we do for GCC) was that some clang/*BSD combination did not (and possibly still do not) support `cc -print-file-name=include`, thus we had no straightforward way of figuring out where to copy the headers _from_.

So, to answer your question, the intrinsics headers are installed only on GCC-based systems, on clang-based systems they are implicitly used by the compiler.

mato:
Anyway, I think I'm all done here. Better not fix anything else that's not broken :-) Merging.

hannesm:
Certainly the approach is legitimate to provide the minimal set of headers required for using solo5.h (btw, that could get rid of `#include <stdbool.h>` as far as I can tell), but from a more pragmatic point of view, the headers copied were "everything needed to compile an OCaml runtime and openlibm" (plus the extra endian for nocrypto - which as I mentioned in #453 was wrong), does this sound appropriate (at least from a historical point of view)?

I have not tested this PR, and neither looked deeper into it, but I recently tried to minimize that set of headers copied around, and stumbled upon `osreldate.h` being used by the OCaml runtime (this is no longer the case with the current trunk (OCaml release 4.12)), see https://github.com/ocaml/ocaml/pull/9506 for more details if interested. Of course `osreldate.h` can be patched or provided by ocaml-freestanding.

I'd suggest to define the `SRCS="float.h stddef.h stdint.h stdbool.h stdarg.h"` only once in configure.sh (to reuse for all clang platforms), and am still struggling with the GCC way of doing things (but maybe this is how to do it with gcc) - it would IMHO be cleaner to use the same approach for both compilers.

hannesm:
NB: I checked, and this PR works as advertised on FreeBSD :D

mato:
> Of course osreldate.h can be patched or provided by ocaml-freestanding.

Ugh, so that (ocaml/ocaml#9506) is why it's in there. I'll re-instate it, since w/o it ocaml-freestanding will not build on FreeBSD.

We _could_ clean it up, but right now I don't feel like spending effort on a boring API-breaking-upper-bounds-dance between solo5 and ocaml-freestanding for the sake of what is a very minor issue. I will, however, add an issue so that we don't forget about it, and if noting else, can remove it some years down the line once everyone is using OCaml >= 4.12.

> I'd suggest to define the SRCS="float.h stddef.h stdint.h stdbool.h stdarg.h" only once

Doesn't make sense now due to the osreldate.h thing. Also, there may be more subtle variations between the *BSDs in future, so we'd just be back to defining multiple versions.

> there does not seem to be a unified approach between compilers, so let's just keep it as is

Indeed. I'd also like a unified approach, but with the current state of compiler toolchains that's not possible, or at least not worth the effort.

Issue #458, opened 2020-05-04
can't install solo5-bindings-hvt/spt on fedora


anuragsoni:
Hello,

I'm trying build a mirageOS unikernel targetting the hvt (i also tried this with spt) backend, and i'm getting a compilation error when opam tries to install solo5-bindings-hvt. My OS is Fedora 32

```
[ERROR] The compilation of solo5-bindings-hvt failed at "/home/asoni/.opam/opam-init/hooks/sandbox.sh build ./configure.sh".

#=== ERROR while compiling solo5-bindings-hvt.0.6.4 ===========================#
# context     2.0.6 | linux/x86_64 | ocaml-variants.4.10.0+flambda | https://opam.ocaml.org#88a179f5
# path        ~/.opam/4.10.0+flambda/.opam-switch/build/solo5-bindings-hvt.0.6.4
# command     ~/.opam/opam-init/hooks/sandbox.sh build ./configure.sh
# exit-code   1
# env-file    ~/.opam/log/solo5-bindings-hvt-164580-d2c37b.env
# output-file ~/.opam/log/solo5-bindings-hvt-164580-d2c37b.out
### output ###
# configure.sh: ERROR: Only 'gcc' 4.x+ is supported on Linux
```

I have gcc installed, and i've installed all the dependencies that depext could find. (kernel-headers is installed). 

OS: Fedora 32
gcc: 10.0.1
mirage: v3.7.6

anuragsoni:
https://github.com/Solo5/solo5/blob/26bc7f2eae311e18f3f1eae5caf5c8f322f7ca8c/configure.sh#L35
This is probably the culprit for the build failure

mato:
Fixed by #459.

PR #459, opened 2020-05-05
get configure script to work for gcc > 9

anuragsoni:
This fixes the build for me and i'm able to install solo5-bindings-hvt after this, and build my mirageOS application. Apologies if the upper limit for gcc was set to 9 on purpose, and 10 is not supposed to be supported just yet.

reference: #458

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -32,7 +32,7 @@ warn()
 
 cc_maybe_gcc()
 {
-    ${CC} -dM -E - </dev/null | grep -Eq '^#define __GNUC__ [4-9]$'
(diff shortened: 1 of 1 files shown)


mato:
> This fixes the build for me and i'm able to install solo5-bindings-hvt after this, and build my mirageOS application.

Thanks!

> Apologies if the upper limit for gcc was set to 9 on purpose, and 10 is not supposed to be supported just yet.

No reason, presumably 10.x seemed a long way away. I see you've set it to accept anything >= 4, so we're all good.

anuragsoni:
Thanks!

PR #461, opened 2020-05-12
configure.sh: Don't install osreldate.h on FreeBSD

### LINKED ISSUE #456: configure.sh: FreeBSD still needs osreldate.h
> Of course osreldate.h can be patched or provided by ocaml-freestanding.

Ugh, so that (ocaml/ocaml#9506) is why it's in there. I'll re-instate it, since w/o it ocaml-freestanding will not build on FreeBSD.

We _could_ clean it up, but right now I don't feel like spending effort on a boring API-breaking-upper-bounds-dance between solo5 and ocaml-freestanding for the sake of what is a very minor issue. I will, however, add an issue so that we don't forget about it, and if noting else, can remove it some years down the line once everyone is using OCaml >= 4.12.

> I'd suggest to define the SRCS="float.h stddef.h stdint.h stdbool.h stdarg.h" only once

Doesn't make sense now due to the osreldate.h thing. Also, there may be more subtle variations between the *BSDs in future, so we'd just be back to defining multiple versions.

> there does not seem to be a unified approach between compilers, so let's just keep it as is

Indeed. I'd also like a unified approach, but with the current state of compiler toolchains that's not possible, or at least not worth the effort.

_Originally posted by @mato in https://github.com/Solo5/solo5/pull/455#issuecomment-621714700_

mato:
With the release of ocaml-freestanding 0.6.0, we can drop osreldate.h
from the list of system/CRT headers installed on FreeBSD.

Fixes #456.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -188,7 +188,7 @@ config_host_freebsd()
     # clang-provided headers for compiler instrinsics. We copy the rest
     # (std*.h, float.h and their dependencies) from the host.
     INCDIR=/usr/include
-    SRCS="float.h osreldate.h stddef.h stdint.h stdbool.h stdarg.h"
+    SRCS="float.h stddef.h stdint.h stdbool.h stdarg.h"
     DEPS="$(mktemp)"
     get_header_deps ${INCDIR} ${SRCS} >${DEPS} || \
         die "Failure getting dependencies of host headers"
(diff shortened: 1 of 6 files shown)


mato:
`20-e2e-x86_64-Debian10` timed out at the last minute, but was actually successful. Merging.

PR #466, opened 2020-05-15
Genode: do not yield for expired deadlines

ehmry:


mato:
Thanks. I have no way of testing this, but I trust that all is fine.

PR #468, opened 2020-06-04
Xen bindings

mato:
This is part of mirage/mirage#1159. The interesting commits are from f4b47d11983bf8f32aaf4ec496a045c35e4e582f onwards.

These are minimal bindings to Xen, for x86_64 only. The purpose of these bindings is primarily to serve as a replacement for Mini-OS in the MirageOS Xen platform stack. As such, no network or block device support is implemented and the related Solo5 APIs are stubbed out.

Only the Xen PVH v2 ABI is supported. This is available from Xen 4.10 onwards.

Being based off the existing virtio bindings, page table setup is entirely static, but is extended to a 4GB identity map in order to be able to access Xen PV device rings which the hypervisor places at the top of memory below the 4GB mark.

Like virtio, all pages are marked RWX for now. This will be addressed in later changes.

In order for MirageOS to be able to interface directly with Xen while still using Solo5 for early boot, a number of private ABIs are introduced:

1. The location of the hypercall page is exported using the `solo5__xen_HYPERCALL_PAGE` symbol.
2. The location of the shared_info page is exported using the `solo5__xen_HYPERVISOR_SHARED_INFO` symbol.
3. The event channel upcall from Xen may be overriden by providing a strong symbol for `solo5__xen_evtchn_vector_handler`.

These private ABIs are exclusively for the use of MirageOS, and may change or go away entirely at any time.


mato:
All done here from the Solo5 point of view, merging.

PR #472, opened 2020-08-26
Add Dune runes

samoht:
This PR adds dune rules to build the opam package (and thus remove the existing Makefile rules for opam, so that we don't have 2 build systems).

The main change that this PR introduces is a new `solo5-bindings` package which contains:
- the solo5 headers (installed in `<prefix>/include/solo5`)
- the CRT headers (installed in `<prefix>/include/crt`)
- solo5-elftool (installed in `<prefix>/bin`)
- solo5-bindings.pc (installed in `<prefix>/lib/pkgconfig`)

Note that `<prefix>` can either be the usual opam root but also (and this is what this PR is about) the local dune workspace prefix (e.g. `_build/install/<context>`).

The new solo5-bindings are useful to build libraries to build against the solo5 API (whatever the bindings is).

This is using the OCaml syntax for `dune` file, as we need to generate a lot of boiler-plate code. Hopefully it won't be too hard to maintain ; we can work with the `dune` maintainers to upstream what is needed to make this simpler in the future.


samoht:
while I'm sure the CI is currently broken for this PR, I'm not sure the current shown CI errors are related:

```
not ok 58 mft_maxdevices spt
# (from function `expect_success' in file tests/tests.bats, line 163,
#  in test file tests/tests.bats, line 577)
#   `expect_success' failed
# timeout: the monitored command dumped core
FAILURE: 'sudo tests/run-tests.sh' failed: Status: 1
```

mato:
I'm assuming that this is superseded by #486, so closing.

PR #476, opened 2020-10-06
xen: Fix assembler warnings in hypercall_page.S

mato:
Recent versions of gas emit a warning when the type of a symbol is
changed between incompatible types. DECLARE_HYPERCALL() does this
intentionally, so silence the warning using the method suggested in the
manual [1].

[1] https://sourceware.org/binutils/docs/as/Type.html#Type


mato:
Travis failures here are due to #477, merging this.

PR #479, opened 2020-10-06
spt: Load seccomp filter manually

### LINKED ISSUE #477: spt: Crash in seccomp_load() on startup due to indirect call to brk()
Recently, Travis jobs have started failing with crashes in the _spt_ version of `test_mft_maxdevices` (example: https://travis-ci.org/github/Solo5/solo5/jobs/733246645#L791).

Here's an example `strace` of the problematic case, using an Ubuntu `xenial` toolchain in a container:

```
brk(0x421a7000)                         = 0x421a7000
brk(0x421c8000)                         = 0x421c8000
brk(0x421e9000)                         = 0x421e9000
prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0)  = 0
seccomp(SECCOMP_SET_MODE_STRICT, 1, NULL) = -1 EINVAL (Invalid argument)
seccomp(SECCOMP_SET_MODE_FILTER, 0, {len = 734, filter = 0x421d8440}) = 0
brk(0x421c1000 <unfinished ...>
+++ killed by SIGSYS (core dumped) +++
Bad system call
```

Note the attempt to lower the process memory allocation by `brk()` after the seccomp filter has been loaded, which does not allow access to `brk()`. It turns out that libseccomp calls `free()` in `gen_bpf_release()`, which is called after the seccomp filter has been installed. If the memory allocated during BPF filter generation is large enough, then some versions of libc will attempt to give back memory to the kernel, thus calling `brk()` and leading to the observed crash.

Possible ways forward:

1. Add `brk()` with suitable restrictions to the default seccomp filter. I'm including this for completeness but it's not a good solution, as the guest has no business invoking this system call while running in the first place.
2. Export the generated BPF filter using `seccomp_export_bpf()` and loading it into the kernel ourselves. The API for this requires a file descriptor, so we'll need to play tricks with `memfd_create()` to get us an in-memory anonymous file to store the BPF in.
3. Wish that Linux seccomp had a way of gradually _dropping_ privileges, similarly to how OpenBSD pledge does it.

In the mean time, I'm going to disable `test_mft_maxdevices` to get the Travis builds to go green until this is resolved.

mato:
libseccomp's seccomp_load() calls free(), which, depending on the libc
malloc() behaviour and generated BPF filter size may call brk(), which
we do not have in our seccomp filter. This leads to a crash during
solo5-spt start-up on some systems.

Work around this by exporting the generated BPF and loading it into the
kernel manually with seccomp().

Fixes #477.

### DIFF
--- a/tenders/spt/spt_core.c
+++ b/tenders/spt/spt_core.c
@@ -38,6 +38,15 @@
 #include <sys/personality.h>
 #include <sys/epoll.h>
 #include <sys/timerfd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+#include <limits.h>
+
+#include <linux/filter.h>
+#include <linux/seccomp.h>
 
 #if defined(__x86_64__)
 #include <asm/prctl.h>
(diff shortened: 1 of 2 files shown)


mato:
Surf-build errors here are due to the CI middleware getting confused; the builds are actually green. Looks good, seems to work, so merging as this fixes an issue that will hit >1 users of spt.

PR #482, opened 2020-11-04
Xen PVH booting via multiboot protocol (from pvgrub2-pvh)

marmarek:
See https://github.com/QubesOS/qubes-issues/issues/6162 for context.

TODO items:
 - [ ] reorganize headers: avoid cross-platform includes;
 - [ ] deduplicate multiboot parsing (move to bindings/multiboot.c?); alternative: don't parse multiboot info at all - will loose cmdline, but memmap can be retrieved via XENMEM_memory_map hypercall
 - [ ] consider mapping low memory too (multiboot boot information is placed there) - will avoid the need to parse hvm_start_info and multiboot in different places

### DIFF
--- a/bindings/xen/boot.S
+++ b/bindings/xen/boot.S
@@ -20,9 +20,28 @@
 
 #include "../cpu_x86_64.h"
 #include "xen/elfnote.h"
+#include "../virtio/multiboot.h"
 
 #define ENTRY(x) .text; .globl x; .type x,%function; x:
 #define END(x)   .size x, . - x
+
+#define XEN_HVM_START_MAGIC_VALUE 0x336ec578
+
+#define MYMULTIBOOT_FLAGS \
+    (MULTIBOOT_PAGE_ALIGN | MULTIBOOT_MEMORY_INFO | MULTIBOOT_AOUT_KLUDGE)
+
+.section .data.multiboot
+
+    .align 4
+    _multiboot_header:
+    .long MULTIBOOT_HEADER_MAGIC
+    .long MYMULTIBOOT_FLAGS
+    .long -(MULTIBOOT_HEADER_MAGIC+MYMULTIBOOT_FLAGS)
+    .long _multiboot_header
+    .long 0x100000
+    .long _edata
+    .long _ebss
+    .long _start32
 
 /*
  * Tell Xen that we are a PVH-capable kernel.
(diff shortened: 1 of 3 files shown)


bindings/xen/platform.c
```diff
@@ -295,6 +341,30 @@ void platform_init(const void *arg __attribute__((unused)))
...
+     */
+    if (multiboot)
+        parse_multiboot(arg);
+
+    pagetable_init();
+    hypercall_init();
+    console_init();
+
+    /*
+     * Parse Xen HVM/PVH start_info structure after setting hypercall page - it
```
mato, line 362: "If booted directly, parse the Xen hvm_start_info structure after setting up hypercalls"

bindings/xen/solo5_xen.lds
```diff
@@ -55,16 +55,17 @@ SECTIONS {
...
     {
+        *(.data.multiboot)
         *(.text)
         *(.text.*)
     } :text
 
+    .interp : {
+        *(.interp)
+    } :interp :text
+
```
mato, line 68: Why did this need moving, exactly? It doesn't really matter for Xen, but I'd still like to understand what's going on.

mato (commented):
Looks good, thanks! Various comments in-line.

bindings/xen/boot.S
```diff
@@ -55,6 +74,13 @@ ENTRY(_start32)
     pushl $0
     pushl %ebx
```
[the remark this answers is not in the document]
marmarek: Yes, exactly.

bindings/xen/solo5_xen.lds
marmarek: The multiboot header needs to be in the first 8k of the file. Without this change, first 4k were program headers, then the second 4k were just this section, pushing multiboot header beyond first 8k.

bindings/xen/boot.S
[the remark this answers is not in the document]
mato: Okay, can you please update both the header comment for this entry point and the comment immediately above to make sense in the new arrangement?

bindings/xen/solo5_xen.lds
mato: You should be able to accomplish the same thing without re-ordering by placing `*(.data.multiboot)` into the `.text` output section as is done in `solo5_virtio.lds`?

marmarek: It's already here and it wasn't enough. `.text` was linked at 8k offset in the file - just beyond what we need. I'm not sure what's the difference with virtio, perhaps more notes here?

mato: Ok, don't worry about it, leave it as-is (and in a separate commit). Thanks.

mato: (... looking at the linker scripts and readelf output, I actually think `.interp` doesn't belong in `:text` at all but I'm not going to muck with that now...)

mato:
> reorganize headers: avoid cross-platform includes;
> deduplicate multiboot parsing (move to bindings/multiboot.c?)

As I mentioned, there is more to sync/de-duplicate between virtio and xen than this, I'm happy to do it at some stage, no need to do it here.

> consider mapping low memory too (multiboot boot information is placed there) - will avoid the need to parse hvm_start_info and multiboot in different places

I'd prefer to leave the low memory unmapped as it stops the unikernel from trivially mucking with its pagetables, intentionally or not.

marmarek:
> I'd prefer to leave the low memory unmapped as it stops the unikernel from trivially mucking with its pagetables, intentionally or not.

I was considering mapping read-only.

mato:
> I was considering mapping read-only.

I like the fact that you can't even introspect your pagetables (or anything else Xen might have "helpfully" put in low memory) without reloading `%cr3`.

marmarek:
Sure, no problem - as you can see it isn't a blocker.

xaki23:
tested with a qubes-builder build of mirage-firewall.
build works.
kernel binary starts in both "direct xen" and "via grub" modes.
basic firewall functionality works in both modes. 

minor confusion about memory-limits (maxmem setting actualy "works" now), but thats just "different" (compared to minios-pv), not "wrong".

mato (approved):
LGTM.

PR #486, opened 2020-12-03
Put Solo5 headers and tools in a standalone package

TheLortex:
This PR separates headers/tools from bindings, allowing to set-up ocaml-freestanding as a cross-compiler without having any binding installed, see https://github.com/mirage/ocaml-freestanding/pull/90. 

This also implies that bindings are now co-installable, as they are chosen link-time by the mirage tool.  

I've messed up a bit with opam files, I'm not sure what version bump you want to make (-> `0.7.0` or `1.0.0`) so I assumed `0.7.0`.

### DIFF
--- a/opam/solo5-bindings-hvt.opam
+++ b/opam/solo5-bindings-hvt.opam
@@ -17,7 +17,7 @@ install: [make "V=1" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENOD
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
-  "solo5-headers"
+  "solo5-headers" {= version}
 ]
 depexts: [
   ["linux-headers"] {os-distribution = "alpine"}
--- a/opam/solo5-bindings-hvt.opam
+++ b/opam/solo5-bindings-hvt.opam
@@ -17,6 +17,7 @@ install: [make "V=1" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENOD
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
+  "solo5-headers"
 ]
 depexts: [
   ["linux-headers"] {os-distribution = "alpine"}
@@ -25,12 +26,7 @@ depexts: [
   ["linux-libc-dev"] {os-family = "debian"}
 ]
 conflicts: [
-  "ocaml-freestanding" {< "0.6.0"}
-  "solo5-bindings-spt"
-  "solo5-bindings-virtio"
-  "solo5-bindings-muen"
-  "solo5-bindings-genode"
-  "solo5-bindings-xen"
+  "ocaml-freestanding" {< "0.7.0"}
 ]
 available: [
   (arch = "x86_64" | arch = "arm64") &
--- a/opam/solo5-bindings-muen.opam
+++ b/opam/solo5-bindings-muen.opam
@@ -17,7 +17,7 @@ install: [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_GENODE
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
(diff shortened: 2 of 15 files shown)


mato:
@TheLortex Thanks. This is awesome! Very happy to see that you've thought about the co-installability as well. 

I started reviewing it, then realised that I wasn't entirely clear on what goes where in terms of packages and filesystem paths, so I wrote up a high-level discussion in #487. Could you please comment on that? Once we have agreement there then we can revise this along those lines.

Minor points which I can answer now:
- Yes, version 0.7.0 please.
- I think that `solo5-bindings-XXX` still need to conflict with the _old_ `solo5-bindings-XXX { <"0.7.0"}` otherwise upgrades of existing switches will break.
- `-headers` and `-bindings` need to be the **exact** same version, can this be expressed in OPAM?

I've not looked yet in enough detail at the downstream mirage/ocaml-freestanding#90 so I may be wrong here:
- In your current setup, the headers `.pc` file contains `!PC_CFLAGS`; I think those should be only in `solo5-bindings-XXX`, so that the headers package doesn't actually need to do the host-toolchain-inspection bits of `configure.sh`?
- Note that `configure.sh` has an `--only-tools`, it should gain an `--only-headers` and make use of that so you don't need the mucking with `CONFIG_XXX` in `opam` for non-bindings packages. Though as I write immediately above ideally headers don't need configure at all.

opam/solo5-bindings-hvt.opam
```diff
@@ -17,6 +17,7 @@ install: [make "V=1" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENOD
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
+  "solo5-headers"
```
reynir, line 20:
```suggestion
  "solo5-headers" {= version}
```

opam/solo5-bindings-muen.opam
```diff
@@ -17,14 +17,10 @@ install: [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_GENODE
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
+  "solo5-headers"
```
reynir, line 20:
```suggestion
  "solo5-headers" {= version}
```

opam/solo5-bindings-spt.opam
```diff
@@ -17,6 +17,7 @@ install: [make "V=1" "CONFIG_HVT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENOD
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {os = "linux"}
+  "solo5-headers"
```
reynir, line 20:
```suggestion
  "solo5-headers" {= version}
```

opam/solo5-bindings-virtio.opam
```diff
@@ -17,14 +17,10 @@ install: [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_MUEN=" "CONFIG_GENODE="
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
+  "solo5-headers"
```
reynir, line 20:
```suggestion
  "solo5-headers" {= version}
```

opam/solo5-bindings-xen.opam
```diff
@@ -17,14 +17,10 @@ install: [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN="
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
+  "solo5-headers"
```
reynir, line 20:
```suggestion
  "solo5-headers" {= version}
```

hannesm:
@mato
> headers and -bindings need to be the *exact same version*, can this be expressed in OPAM?

yes, this can be expressed via (in `depends:`):
`solo5-headers {=version}`

...as @reynir also indicated in his commit suggestions.

mato:
Superseded by #494.

Issue #487, opened 2020-12-07
Split Solo5 into multiple packages


mato:
This is a general issue to work out two things:

1. Split out tools (`solo5-elftool`, `solo5-virtio-mkimage`, `solo5-virtio-run`) and headers (`include/*`) into separate packages.
2. Do this in a way that will pave the way for OS/distribution-friendly packages of Solo5.

The motivation for 1 are the Mirage 4 cross-compilation improvements (PoC by @TheLortex #486, mirage/ocaml-freestanding#90), the motivation for 2 is to be able to distribute binary unikernels (/cc @hannesm).

mato:
Currently we have multiple `solo5-bindings-*` packages, each conflicting with one another in OPAM, which install various combinations of the following, depending on the actual target XXX:

1. Header files:
    1. Public and private Solo5 headers: `include/solo5/*` **AS** `$PREFIX/include/solo5-bindings-XXX/solo5/*`.
    2. Base C compiler headers appropriated from the host C compiler toolchain at configure time: `include/crt/*` **AS** `$PREFIX/include/solo5-bindings-XXX/crt/*`.
2. Tools:
    1. ELF manifest/ABI tool: `solo5-elftool` **AS** `$PREFIX/bin/solo5-elftool`
    2. Various scripts: `solo5-virtio-mkimage.sh`, `solo5-virtio-run.sh` **AS** `$PREFIX/bin/solo5-YYY`.
3. Bindings:
    1. Compiled bindings as an object file for linking: `solo5_XXX.o` **AS** `$PREFIX/lib/solo5-bindings-XXX/solo5_XXX.o`.
    2. Linker script: `solo5_XXX.lds` **AS** `$PREFIX/lib/solo5-bindings-XXX/solo5_XXX.lds`.
    3. Pkg-config information `solo5-bindings-XXX.pc` **AS** `$PREFIX/lib/pkgconfig/solo5-bindings-XXX.pc`.
4. Tender (hvt and spt only):
    1. Compiled tender binary: `solo5-XXX` **AS** `$PREFIX/bin/solo5-XXX`.

#486 gets us part of the way there, but since we're splitting out packages and thus also incompatibly breaking the build system downstream, we should take this opportunity to go as far as we can given what we know about current and future requirements.

What I _think_ we want to get to is something like:

[...]

PR #488, opened 2020-12-09
WIP: Split up into multiple packages

mato:
Work in progress, to supersede #486. Discussion in #487.

## Split up into multiple packages

This change splits up Solo5 into multiple OPAM packages, and provides
infrastructure enabling downstream maintainers to easily package Solo5
components for their operating system.

Makefiles now support conventional use of DESTDIR= and PREFIX= to
control installation for packaging scripts.

configure.sh now implements an --enable-targets=target,... option. The
default is "all targets available on this build OS/CPU architecture".
"none" may be specified if you are interested in building and installing
common tools or headers only.

The following components can now be selectively installed through the
use of different Makefile targets after running configure.sh and make:

install-tools: Installs common tools (elftool, virtio scripts).
install-headers: Installs common headers (crt headers from toolchain,
public Solo5 API headers).
install-bindings: Installs bindings for enabled targets.
install-tenders: Installs tenders for enabled targets.
install: All of the above.

Other notable changes:
- Makefiles: Old OPAM-specific "install" and "uninstall" targets
  removed.
- Makefiles: Reorganise in-tree include/ to follow the installed
  directory structure.
- hvt: "solo5-hvt-configure", deprecated for some time, has been
  removed.
- OPAM: New package: solo5, containing common tools and header files.
- xen: Install Xen headers as part of the Xen bindings.

## Replace pkg-config with solo5-config

This replaces the pkg-config based way of passing Solo5 compile-time
parameters to downstream consumers with a new tool, "solo5-config".

Briefly, use

    solo5-config --target=TARGET ...

to access compiler and linker parameters for the selected target.

solo5-config also understands the concept of "toolchains", where
different Solo5 targets may share a single "toolchain", thus enabling
multiple Solo5 bindings sharing a common toolchain to coexist in OPAM /
OCaml installations.

There is currently a single toolchain configuration, "static",
represented by the OPAM package solo5-config-ocaml-static.

The unit tests in tests/ use an in-tree copy of solo5-config during
build, this ensures that the system is tested during normal development
work (i.e. without requiring "make install").

Other changes:
- configure.sh now has a --prefix= parameter. Packagers should use this
rather than PREFIX= during make, since the generated solo5-config script
needs to embed the installation prefix.
- Use ginstall on OpenBSD, because their install _has_ to be different,
of course.

## Remove Genode bindings

The Genode bindings have been unsupported for some time, and require
extra work to be forward-ported to the new solo5-config scheme. Remove
them for now, they can be reinstated later if there is interest and a
solution for the toolchain issues can be found.


configure.sh
```diff
@@ -163,20 +188,17 @@ config_host_linux()
...
+        MAKECONF_LIBSECCOMP_LDLIBS="$(pkg-config --libs libseccomp)"
     fi
-    if ! PKG_CFLAGS=${MAKECONF_SPT_CFLAGS} gcc_check_header seccomp.h; then
+    if ! PKG_CFLAGS="${MAKECONF_LIBSECCOMP_CFLAGS}" gcc_check_header seccomp.h; then
         die "Could not compile with seccomp.h"
     fi
-    if ! PKG_LIBS=${MAKECONF_SPT_LIBS} gcc_check_lib -lseccomp; then
-        die "Could not link with -lseccomp"
+    if [ -n "${MAKECONF_LIBSECCOMP_LDLIBS}" ]; then
+        if ! gcc_check_lib "${MAKECONF_LIBSECCOMP_LDLIBS}"; then
```
TheLortex, line 198: This check doesn't work on my computer / in my CI testing. 
Removing the quotes seems to do the trick but I'm not sure of the implications.
```suggestion
        if ! gcc_check_lib ${MAKECONF_LIBSECCOMP_LDLIBS}; then
```

mato: Thanks, yes, the quoting is a bit subtle there, I think `$MAKECONF_LIBSECCOMP_LDLIBS` should get passed as multiple arguments. I really should also run all the shell code through shellcheck.

I've made that change, rebased on master and force-pushed.

For reference, what is the value of that variable / the output of `pkg-config --libs libseccomp` on your machine? And what Linux distro are you on?

TheLortex: Okay thank you, this works for me. 

On my machine the output is `-lseccomp`, using Archlinux.

opam/solo5-bindings-spt.opam
```diff
@@ -10,13 +10,14 @@ bug-reports: "https://github.com/solo5/solo5/issues"
...
-  ["./configure.sh"]
-  [make "V=1" "CONFIG_HVT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENODE=" "CONFIG_XEN="]
+  ["./configure.sh" "--prefix=%{prefix}%" "--enable-targets=spt"]
+  [make "V=1"]
 ]
-install: [make "V=1" "CONFIG_HVT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENODE=" "CONFIG_XEN=" "install-opam-spt" "PREFIX=%{prefix}%"]
+install: [make "V=1" "install-bindings" "install-tenders"]
 depends: [
-  "conf-pkg-config"
-  "conf-libseccomp" {os = "linux"}
```
TheLortex, line 19: I think `conf-libseccomp` is still necessary for the spt bindings as it's needed to install the `libseccomp` external dependency.

TheLortex:
On FreeBSD 12.2 the configure step fails with the following error
```
# cc: error: no such file or directory: 'float.h stddef.h stdint.h stdbool.h stdarg.h'
# cc: error: no input files
# configure.sh: ERROR: Failure getting dependencies of host headers
```
@hannesm could you try and confirm the issue, and maybe come up with a solution ?

hannesm:
@TheLortex would you mind to expand what is actually failing?

For me, `./configure.sh` on FreeBSD 12.1 works nicely (reporting `Enabled targets: hvt,virtio,muen,xen.`). A `./configure.sh --enable-targets=none` fails with the error @TheLortex reported above. A `./configure.sh --enable-targets=hvt` works fine as well.

mato:
@hannesm We discussed this with@lortex yesterday, it appears that something has changed in the `clang` and/or `/usr/include` layout in 12.2 which causes `configure.sh` to fail.

TheLortex:
I found out that the `get_header_deps` function is failing, because it receives for argument `"float.h stddef.h stdint.h stdbool.h stdarg.h"` which is directly passed through in `cc -M "float.h stddef.h stdint.h stdbool.h stdarg.h"` which fails.. the correct quoting would be `cc -M "float.h" "stddef.h" "stdint.h" "stdbool.h" "stdarg.h"`

hannesm:
ok, FWIW the difference between `--enable-targets=none` and no argument passed to `configure.sh` is:

no arguments:
```
cd /usr/include
cc -M float.h stddef.h stdint.h stdbool.h stdarg.h
```

`--enable-targets=none`
```
cd /usr/include
cc -M 'float.h stddef.h stdint.h stdbool.h stdarg.h'
```

and with the second line (header files surrounded by `'`), `cc` invocation fails. is this the same on 12.2?

hannesm:
and the issue is (for `--enable-targets=none`) that `enable_targets()` does not reset the `IFS` to `OLDIFS` if the for loop is not entered.

hannesm:
Possible fix:
```bash
# Arguments: Comma-separated list of TARGETS to enable.  Sets CONFIG_TARGET=1
# for each TARGET that is enabled, dies with an error if any TARGET is not
# supported on the current CONFIG_HOST/CONFIG_ARCH.
enable_targets()
{
    OLDIFS=$IFS
    IFS=,
    for target in $@; do
        if echo ",${AVAILABLE_TARGETS}," | grep -q ",${target},"; then
            tmp="$(echo ${target} | tr '[a-z]' '[A-Z]')"
            eval "CONFIG_${tmp}=1"
        else
            die "--enable-targets: Target '${target}' is not available on ${CONFIG_HOST}/${CONFIG_ARCH}"
        fi
    done
    IFS=$OLDIFS
}
```

(moving the `IFS=$OLDIFS` line after the loop)

opam/solo5-bindings-spt.opam
mato: Correct. Fixed.

mato:
@hannesm Thanks for tracking down the `IFS` issue. I hadn't caught it, since had not tested `--enable-targets=none` on FreeBSD. I've pushed a fix.

mato:
Rebased, re-ordered commits, wrote up proper commit messages and added coreutils build-time depext for OpenBSD.

Hopefully I've not broken anything. @TheLortex / @hannesm Please test when you get a chance.

hannesm:
Thanks for your updates. I looked through this, and tested it on FreeBSD and OpenBSD. I have remote access to a OpenBSD box without superuser privileges -- since `ginstall` is not installed there (and I wanted to test `install` as well), I opened https://github.com/mato/solo5/pull/6 which removes the `ginstall` dependency. If you like the changes, feel free to integrate that commit into your clean series of commits here.

mato:
@hannesm Thanks, applied into 6d7b898, rebased and force-pushed.

hannesm:
thanks @mato -- depending on how you want to merge (squash or keep individual commits), I'd suggest to merge it into 09d044a, since now:
- 09d044a uses `${INSTALL}` in the Makefile
- 6d7b898 replaces that with `install -p`

Alternatively, it may be worth to keep `${INSTALL}` and manually remove the `ginstall` and `#ifdef` from the old second commit. I'm not sure which style is better.

mato:
@hannesm I do want to keep the individual commits. I'll re-work 6d7b898 to use `$(INSTALL)`.

hannesm:
I tested this from a packaging point of view on FreeBSD, and it works fine. (running configure where muen,xen,virtio,hvt is enabled, followed by make and make install leads to a staging area where bin has the tools + tenders, lib/solo5 the bindings and linker scripts and include/solor5 the crt, xen, and solo5 header files.

mato:
Superseded by #494.

PR #493, opened 2020-12-22
CI: Disable Genode

mato:
Genode bindings are failing to build on Debian testing / GCC 10.x. Given
that these are going to be removed soon, remove them from CI now so that
we can get it to go green.


mato:
OpenBSD67 failure is the usual nodejs/surf-build `EBADF` nonsense, i.e. an actual internal error, but the Debian 11 build succeeds now.

PR #494, opened 2021-03-09
New packaging, toolchain, cross-compilation support

mato:
This is the latest iteration of the packaging and toolchain changes discussed in #487. Many thanks to @TheLortex for working with me over the past couple of weeks to finally get us to something that works, both for Solo5 and for MirageOS / ocaml-freestanding downstream.

The first commit removes the Genode bindings, as these have been unsupported for some time. 

The bulk of the changes are in the 2nd commit, see the commit message for the full description. I'm reproducing the main goals and changes here:

1. **Make Solo5 installable system-wide**: `./configure.sh --prefix=PREFIX` and `[ DESTDIR=DESTDIR ] make install` are now supported and work in the conventional manner. This also enables packaging of Solo5 for the BSDs and Linux distributions. Packagers wishing to split "host" and "development" packages will find the `--disable-toolchain` option to `configure.sh` and/or the `install-*` sub-targets to `make` useful for this purpose.

2. **Replace Solo5-internal use of "pkg-config"** by generated toolchain wrappers installed as `PREFIX/bin/ARCH-solo5-none-static-{cc,ld,objcopy}`. These are intended to be used by downstream consumers to compile and link Solo5 unikernels. The Solo5 bindings to use when linking are specified by calling the `cc` or `ld` wrappers with the option `-z solo5-abi=ABI`.

3. **All Solo5 bindings (ABIs) are now co-installable** under the same PREFIX.

4. **OPAM packaging is greatly simplified**; there is now just a single `solo5` package, installable on the union of all OSes and ARCHs that can build some Solo5 bindings. This package installs all bindings and tenders (if any) supported/buildable on the host system.

5. **Experimental: Support for cross-compiling Solo5 bindings** and installing a Solo5 cross-toolchain targetting a different architecture. See `./configure.sh --help` for details, and the `solo5-cross-aarch64` package for a possible way to package this in OPAM -- this package is NOT yet intended for publishing to the public OPAM repository.

Before this is merged, I'd appreciate as much testing as possible, especially on Linuxen that are not Debian. It'd also be good to document how the new toolchain works in more detail.


Ericson2314:
@mato Excited about these changes, especially splitting the host and development stuff---we are always trying to do that in Nixpkgs. Two things though:

 -  I think I'd miss `pkg-config` support. You mention it was mainly for internal use, but I would like to use it externally: i dream of making all libraries even the "runtime" ones work with pkg-config and getting rid of all the `-I` `-L` hacks in Nixpkgs.
 
 - > All Solo5 bindings (ABIs) are now co-installable under the same PREFIX.
   
    What does the layout look like then? In particular in Nixpkgs because we aren't constrained by a single FHS prefix, we always just used the standard `$prefix/lib` `$prefix/include` etc. per each package. Schemes that do use e.g. `$prefix/lib/${target}/` to disambiguate we want to opt-out of because our separate prefixes for everything is sufficient. Is that possible?

sternenseemann:
> All Solo5 bindings (ABIs) are now co-installable under the same PREFIX.

I don't really understand this. Currently you can do that, i. e. we install all solo5 bindings for ocaml into the same directory already now in our `solo5` package (as opposed to splitting them up into multiple packages).

> I think I'd miss pkg-config support. You mention it was mainly for internal use, but I would like to use it externally: i dream of making all libraries even the "runtime" ones work with pkg-config and getting rid of all the -I -L hacks in nixpkgs.

Agreed, `pkg-config` makes `solo5` and things depending on `solo5` like `ocaml-freestanding` work with nixpkgs's unconventional 
layout with almost zero messing about.

mato:
@Ericson2314 @sternenseemann Thank you for bringing this up. It's precisely why we wanted some "external" review on this and the companion mirage/ocaml-freestanding#92 PR, so that we can be sure this will work well in diverse settings going forward.

I'm going to be offline for a couple of weeks, but I'll try and address your initial questions now. @TheLortex and @dinosaure also understand the motivation behind these changes, so you can discuss with them while I'm away if anything is unclear. See also #487 for lots of discussion leading up to this.

Please note that I don't have any experience with NixOS, so I may not be understanding all constraints from your side.

> I think I'd miss pkg-config support. You mention it was mainly for internal use, but I would like to use it externally:

We really want to get rid of all use of pkg-config in Solo5 and for any (OPAM) packages in Mirage 4. There are several reasons for that:

- We want to ensure that nothing that the _host_ exposes via pkg-config leaks into the Solo5 _target toolchain_, even accidentally.
- mirage/ocaml-freestanding#92 builds a self-contained OCaml "cross" _toolchain_ matching the Solo5 _toolchain_. It does this in a way that (through support in dune) no longer has any _need_ to pass flags around using pkg-config.
- we want to support as many _host_ systems as is practical for building Solo5 unikernels
- and, at the same time, have our scheme be future-proof for allowing _multiple_ Solo5 toolchains to co-exist and be discoverable on the _host_

Considering the above, pkg-config is both brittle and insufficient. It's always been a "hack" to get the current scheme to work, but nothing that we wanted to keep. The alternatives were (discussed in #487):

1. Some kind of "better pkg-config", i.e. a solo5-config script that lets you interrogate toolchains, retrieve the relevant compiler flags, and so on.
2. Toolchain wrappers that make each Solo5 toolchain look as much as possible as a standard C _cross toolchain_.

We initially went down the rabbit hole of option 1 (#488), but that started looking more and more like writing "a database in shell scripts", and, intuitively, I've always wanted option 2. So this is how we got to #494, which we believe is the right way to go.

> All Solo5 bindings (ABIs) are now co-installable under the same PREFIX.

This is not clear, we're talking about OPAM: What we mean here is that all Solo5 bindings available on a particular _host_ and targetted to its native arch are now part of a single `solo5` OPAM package. Previously, they were split up into individual `solo5-bindings-xxx` packages. So, if anything, it seems to be more like the setup you want for NixOS?

> What does the layout look like then?

You remind me that this should be properly documented before this PR is merged. Using the default prefix of `/usr/local`, here is what Solo5's `make install` installs, on `x86_64`:

`/usr/local/lib/x86_64-solo5-none-static/`: Toolchain-specific libraries: All Solo5 bindings available for the `x86_64-solo5-none-static` toolchain, i.e. one `.o` and one `.lds` for each. These were previously in `PREFIX/lib/solo5` (with the constraint that they weren't co-installable in OPAM, and the flaw that this layout did not allow for co-installing multiple Solo5 toolchains).
 `/usr/local/include/x86_64-solo5-none-static/`: Toolchain-specific headers: This is where we install "internal" headers _appropriated_ from the _host_ system's toolchain which are required to be able to compile C code. These were previously in `PREFIX/include/solo5/crt`, i.e. same flaw as above, no provision for co-installing multiple toolchains.
`/usr/local/include/solo5/`: Public Solo5 headers. These are shared by all Solo5 toolchains, so there's the constraint that you can have only one version of Solo5 development packages installed under /usr/local, but that is consistent with what I'd expect of other development packages.
`/usr/local/include/solo5/xen/`: On `x86_64`, this is a copy of the Xen public headers that match the Solo5 `xen` bindings, since `mirage-xen` needs to use these downstream.
`/usr/local/bin/x86_64-solo5-none-static-TOOL`: These are the actual wrappers implementing the Solo5 toolchain. Currently we only have cc, ld and objcopy, but there's no reason why we could not add other relevant tools (readelf, strip, ar, ...) in the future.
`/usr/local/bin/`: Common binaries shared by Solo5 on the _host_ system. i.e. all tenders available for _host_, `solo5-elftool` and some support scripts.

I hope this helps to understand how things are laid out. I'm not 100% convinced that it's the _right_ layout (e.g., we could have experimented with more of a "sysroot"-like layout for the toolchain's include and lib directories), but it seems to be the best we can do with what we know today.

I also hope I've explained our motivation for doing this a  bit better. It's certainly helped me to write it up here, I can use this text to add to the Solo5 documentation as part of the PR when I get back.

dinosaure:
> We have mechanisms to make sure our PKG_CONFIG_PATHs for platforms are never mixed, so that isn't a problem.

I think you know that it's not true for others distributions 🙂 where the environment is _more mutable_. Currently, we mostly try to pass C flags given by our own OCaml compiler (`ocaml-freestanding`) to C stubs (in our OCaml libraries) without any strange or external mechanisms (such as `pkg-config`) in the point-of-view of the OCaml eco-system. So we choose to use `dune` as the tool to get C flags from the current OCaml compiler used and pass them then to compile C stubs.

Ericson2314:
> I think you know that it's not true for others distributions slightly_smiling_face where the environment is more mutable.

Well I would say other distributions (baring a few like https://exherbo.org/) don't try to do cross at all, so it's better to just ignore them completely :). Yes, our central conundrum for cross in Nixpkgs has been undoing the workarounds people are forced to do to get around their package manager not understanding cross compilation and bootstrapping. It's understandable but tragic, because to keep things as composable as they are in the native case, those issues must be handled by the package manager and not the packages themselves.

> Currently, we mostly try to pass C flags given by our own OCaml compiler (`ocaml-freestanding`) to C stubs (in our OCaml libraries) without any strange or external mechanisms (such as `pkg-config`) in the point-of-view of the OCaml eco-system. So we choose to use `dune` as the tool to get C flags from the current OCaml compiler used and pass them then to compile C stubs.

That actually sounds find to me. That C code is *within* the OCaml package as you say, and `pkg-config` is for getting information between packages. Using `dune` for that sounds perfectly fine to me.

Is there any pkg-config that was for inter-package not intra-package communication, or was that all handled by the tool-chain wrappers?

dinosaure:
> Is there any pkg-config that was for inter-package not intra-package communication, or was that all handled by the tool-chain wrappers?

We handle that with a tool-chain wrapper. OPAM packages and any libraries should not care about which C flags we need to cross-compile to a specific target when:
1) these libraries will be downloaded locally with [`opam monorepo`](https://github.com/ocamllabs/opam-monorepo)
2) `dune` will expand C flags (from `ocaml-freestanding`) locally to these libraries when it will compile the final unikernel

At the end, [cross-compiled artifacts](https://github.com/mirage/mirage-crypto/tree/main/src-freestanding) such as C stubs needed by some of our OCaml libraries will be deleted in the mid-term and, then, any libraries with or without C stubs will be compatible (and can be cross-compiled) with MirageOS _à priori_. So, I think (but I'm not an expert of NixOS), with such update along our stack, it will be easier to distribute our OCaml libraries into NixOS - where libraries/packages will not have any mention of cross-compiled stuff (such as [special `META` file](https://github.com/mirage/mirage-crypto/blob/main/META.mirage-crypto.template), cross-compiled `*.a` artifacts into the distribution & [optional dependencies to `ocaml-freestanding`](https://github.com/mirage/mirage-crypto/blob/3013c5286f563743e2f5c13da068a6bdf90cd805/mirage-crypto.opam#L24-L26)).

Ericson2314:
OK sounds good. And it looks like in https://github.com/mirage/ocaml-freestanding/pull/92 ocaml-freestanding is in fact keeping it's pc file? That would also be nice since it looks like it's staying a regular C (no dune, no mandatory opam) library.

Ericson2314:
It sounds like we're converging here :), but one last idea to keep in mind: As Nixpkgs likes to assembly it's own toolchains, I can imagine we could make a solo5-based C one that could work for MirageOS, IncludueOS, and other solo5 projects alike? While the MirageOS OCaml libraries of course have C stubs and those are built together as discussed, it can be nice to package the all-C-no-Ocaml bits in a non-MirageOS-specific way to "witness" the separation of concerns.

(Maybe the way includeOS and MirageOS deal with e.g. libc libm etc. is different, but the large composability point stands that it's nice to always package the same upstream thing once and compose it with different downstream consumers, and generally be very granular so downstream concerns leak upstream as little as possible.)

Anyway, excited to see MirageOS 4 come along, and package things the new way!

Ericson2314:
Thanks it's coming together. To be clear, we can function without pkg-config because tons of things don't install it. But it we do like it. I guess the pc files in this repo were pretty opam-specific, so dropping isn't so bad

> Yes it does.....You don't have to follow the "everything in one package" model if you don't want to. See the various different `install-*` subtargets in the Makefile. However, if you do come up with your own split, I'd appreciate an overview of what that is along with justification for it, so that we don't accidentally de-support it later.

Hmm this is going to be a bit harder. For example, because of the single package, the elftool Makefile has `HOST*` everything. But if we are building `elftool` on it's own, we want to use regular `CC` and friends: it's the one thing we are installing, and the thing are installing is the regular platform *by definition*.

If you could you do something like `@echo "MAKE $@" CC=HOSTCC` for `elftool` from the main Makefile, that would be much appreciated!

> Please do not rely on that, that is there to allow for some backward compatibility for future LTS releases of Mirage 3 (not Mirage 4) to continue to work with the new Solo5 and ocaml-freestanding.

> Solo5 deliberately does not provide a libc or a libm. For MirageOS, this is provided by ocaml-freestanding, and exposed via dune to C as well as OCaml packages. Now, this may well be the weak point in the whole setup, since in a "conventional" C cross compilation scenario there would instead be a sysroot, and ocaml-freestanding (or IncludeOS, or whoever) would install libc, libm or whatever other C libraries are needed into that sysroot.

Yeah this fine. We can do toolchains without libcs. That could be the toolchain we'd use to build ocaml-freestanding (or IncludeOS, or whoever itself, for example. We don't do sysroots eithers, for us, we try to keep it very uniform as "just" a compiler and some run-time libraries. The only reason those libraries are special at all is because downstream packages don't mention them by name and we want flags like `-nostdinc` etc. to work.

adamsteen:
I would like to test this on OpenBSD can anyone tell me the best way?

hannesm:
I can confirm that this PR works fine on FreeBSD-13. I suggest to incorporate https://github.com/hannesm/solo5/commit/5022b503bd1acd7cee0b4eb8aaf1f6c9a266779a (to enable virtio testing on FreeBSD) and https://github.com/hannesm/solo5/commit/d9e890f3b3b3f458d0b80ed28ff8c4b74f63233e (to deal with a race condition in the solo5-virtio-run script) -- both are in hannesm/solo5.git on the freebsd-test branch.

dinosaure:
Close by #506

PR #496, opened 2021-05-21
configure: make cc headers writeable after copying

Mic92:
This is necessary on both [NixOS](nixos.org) and [GuixSD](http://guix.gnu.org/) because their headers
are on a read-only filesystem.


hannesm:
Hi, thanks for the report. I don't quite understand what the issue is here -- what is actually failing since the headers are read-only?

Mic92:
I don't remember, but I also no longer need this fix.

PR #500, opened 2021-11-17
[WIP] virtio: detect PIIX4 and attempt to power off via PIIX4 (broken)

### LINKED ISSUE #499: virtio: implement ACPI power-off
Currently, when unikernels terminate they just kind of hang forever on GCE instances instead of powering off.

Additionally, when a GCE instance receives a stop command, it sends an ACPI power-off signal to the instance. On Solo5 it appears this is ignored, so the GCE hypervisor waits up to 90 seconds before actually shutting down the instance.

Not sure if this requires a full ACPI implementation that matches the underlying hardware running the server, which would be a huge ordeal, or if GCE provides a more virtualized form that can be harmlessly targeted on non-GCE hosts.

mbacarella:
This is attempt 1 of 2 to solve #499 (attempt 2 is at #501)

Summary:

This is my failed attempt to resolve #499 . It will detect the PIIX4 on GCE and attempt to power off if you exit the mirage unikernel (e.g. raise an uncaught exception right after you start), but it hangs instead of powering off because it's not a correct implementation.

Discussion:

Detecting the PIIX4 on the PCI bus works; it requires a simple change to the `pci_enumerate` function to list "functions" of PCI devices. The part where it tries to actually instruct the PIIX4 to power off is completely broken though.

I'm at the limits of my understanding of low-level x86 architecture, but will post this anyway in the event that someone with more expertise can determine if this is 90% of the way there and make use of it.

The specific problem is the base address that PCI enumeration probes for the PIIX4 device is 0, which obviously can't be used to inw/outw the device. I tried to learn the from [this Linux kernel code](https://github.com/torvalds/linux/blob/master/drivers/power/reset/piix4-poweroff.c) but it became too challlenging for me to reason about.

### DIFF
--- a/bindings/virtio/bindings.h
+++ b/bindings/virtio/bindings.h
@@ -48,16 +48,21 @@ int tscclock_init(void);
 uint64_t tscclock_monotonic(void);
 uint64_t tscclock_epochoffset(void);
 void cpu_block(uint64_t until);
+void cpu_wasteful_milli_sleep(uint64_t millis);
 
-/* pci.c: only enumerate for now */
+/* pci.c: only enumerate for now (except with a hack for poweroff). */
 struct pci_config_info {
     uint8_t bus;
     uint8_t dev;
+    uint8_t fun;
     uint16_t vendor_id;
+    uint16_t device_id;
     uint16_t subsys_id;
     uint16_t base;
     uint8_t irq;
 };
+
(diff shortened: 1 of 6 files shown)


dinosaure:
Replaced by #647, thanks for your initial work, it helps me to go further.

PR #501, opened 2021-11-17
[WIP] virtio: detect PIIX4 and attempt to power off via ACPI hack (broken)

### LINKED ISSUE #499: virtio: implement ACPI power-off

mbacarella:
This is attempt 2 of 2 to solve #499 (attempt 1 was #500)

Summary:

This is my failed attempt to resolve #499 . It will detect the PIIX4 on GCE and attempt to power off if you exit the mirage unikernel (e.g. raise an uncaught exception right after you start), but it hangs instead of powering off because it's not a correct implementation.

Discussion:

Detecting the PIIX4 on the PCI bus works; it requires a simple change to the pci_enumerate function to list "functions" of PCI devices. Then it attempts to parse just enough of ACPI to learn parameters needed to power off. This approach is probably not generally feasible, but may work on GCE.

This adapts the method from C code here https://forum.osdev.org/viewtopic.php?t=16990 though it fails.

More information at #499.

### DIFF
--- a/bindings/GNUmakefile
+++ b/bindings/GNUmakefile
@@ -50,7 +50,7 @@ virtio_SRCS := virtio/boot.S virtio/start.c $(common_SRCS) \
     virtio/platform.c virtio/platform_intr.c \
     virtio/pci.c virtio/serial.c virtio/time.c virtio/virtio_ring.c \
     virtio/virtio_net.c virtio/virtio_blk.c virtio/tscclock.c \
-    virtio/clock_subr.c virtio/pvclock.c 
+    virtio/clock_subr.c virtio/pvclock.c virtio/acpi.c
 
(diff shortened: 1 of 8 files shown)


dinosaure:
Replaced by #647, thanks for your initial work, it helps me to go further.

PR #495, opened 2021-04-05
Clean Up OpenBSD a bit

adamsteen:
6.7 no longer supported
all support releases use lld

PR #502, opened 2021-12-09
Fix LD scripts syntax

renatoalencar:
Although this syntax is supported by newer versions of GNU ld, it looks it's not supported some older releases. This seems to fix the problem and works for both older and newer versions.

Here's my version of ld, which is still the newer version for Fedora 35 https://src.fedoraproject.org/rpms/binutils
```
GNU ld version 2.37-10.fc35
Copyright (C) 2021 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) a later version.
This program has absolutely no warranty.
```

I have tested on the current trunk for GNU binutils and it seems to work tough, which is the same minor version.
```
GNU ld (GNU Binutils) 2.37.50.20211207
Copyright (C) 2021 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) a later version.
This program has absolutely no warranty.
```

### DIFF
--- a/bindings/hvt/solo5_hvt.lds
+++ b/bindings/hvt/solo5_hvt.lds
@@ -38,7 +38,7 @@ PHDRS {
     text PT_LOAD FLAGS(5); /* No FILEHDR or PHDRS, force R/E only.
                               FLAGS values come from PF_x in elf.h */
     data PT_LOAD;
-    note.not-openbsd PT_NOTE; /* Must come first. */
+    note.not_openbsd PT_NOTE; /* Must come first. */
     note.abi PT_NOTE;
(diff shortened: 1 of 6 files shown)


hannesm:
Thanks for your investigation and fix. @dinosaure @TheLortex could we get a 0.6.9 release with this (and #495) before the big merge of #494? I stumbled upon this issue in https://github.com/mirage/mirage-net-solo5/pull/37 where OCaml-CI uses alpine-3.15.

dinosaure:
Thanks, let's do a minor release for OpenBSD.

PR #503, opened 2021-12-13
Update CHANGES for release v0.6.9

dinosaure:
/cc @hannesm, everything is ok?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,13 @@
+## 0.6.9 (2021-12-13)
+
+* virtio: add missing reset for net and blk devices (#491)
+* CI: disable genode (#493)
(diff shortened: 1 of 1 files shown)


hannesm:
looks good to me, a minor nit is that "fix syntax error in ld script" is not very descriptive - -this only affects some bintuils/ld versions, maybe: `avoid "-" in section names (in the ld scripts) since some GNU ld versions reject that (alpine 3.15, Fedora 35)`?

dinosaure:
I force-pushed the new version :+1:.

TheLortex (approved):

dinosaure:
Solo5 is released on `opam`, we can merge this PR.

PR #506, opened 2021-12-15
New toolchain (2)

dinosaure:
This PR is #494 rebased according to `master` which includes #502 & #495. The only conflict is about the `configure.sh` and I just took the Martin's version. I would like to say that we mostly agreed (already) to merge this PR and cut a release. Our plan was postponed by the fact that if we want to continue to support MirageOS 3 or not.

Indeed, on top of this PR, @TheLortex did few patches to include the `pkg-config` path to pass C flags across packages. This PR **does not** include these patches and we agreed (the MirageOS core team) that we don't want to support MirageOS 3 and `pkg-config` stuffs at this layer (and on the `ocaml-freestanding` layer).

That mostly means that this PR can be merge as is and we can cut a release (`0.7.0`) then. However, I would like a triple check from the CI, @TheLortex and someone else from @mirage/core before merging this PR.

PR #509, opened 2022-01-25
Fix syntax error in stubs ld file

TheLortex:
When switching to Fedora 35 and updating to solo5 0.7.0, I've encountered again the error that was fixed by https://github.com/Solo5/solo5/pull/502.
What happened is that in the new stubs ld file introduced in https://github.com/Solo5/solo5/pull/506, the breaking syntax was still used.


dinosaure:
Arf, it's a shame, I will cut a Solo5.0.7.1 as soon as I can. Thanks!

dinosaure:
I understand why we did not integrate that on `0.7.0`, #502 updated different files according to the old layout and we did not upstream this change correctly in #506 indeed. Thanks!

PR #510, opened 2022-01-30
elftool: generate manifests from stdin or to stdout

ehmry:


hannesm:
LGTM, another way would be to use /dev/stdin and /dev/stdout explicitly (but `-` is very convenient indeed)

ehmry:
Ah, I didn't know about that trick.

dinosaure:
Thanks!

PR #514, opened 2022-03-17
Fix the compilation of test_tls on ppc64 architecture

dinosaure:
/cc @hannesm, /cc @TheLortex it's a possible fix for `ppc64` architecture. The fix cast the pointer to `tcb{1,2}._data` to an `uintptr_t` to be able to apply arithmetic operations (`+`) and add the `PPC64_TLS_OFFSET`.


hannesm:
Thanks for taking care of this. I won't have time and hardware to review it properly. Since it is a test, from my perspective merge if it compiles and works. :)

dinosaure:
Thansk!

PR #517, opened 2022-05-26
xen: remove the "skip the first token in the cmdline" from multiboot code path

hannesm:
On recent QubesOS (4.1), this "opaque 'name'" is not present in the multiboot
cmdline.

//cc @xaki23 @palinp @marmarek


palainp:
This code was originally in the virtio parse_multiboot, perhaps it is specific to virtio? From my observations, it seems safe to remove it here.

PR #520, opened 2022-07-11
tenders, FreeBSD, tap: Ignore "up" variable to avoid compiler warning

### LINKED ISSUE #519: Build error for FreeBSD
Build process fails for FreeBSD with set but unused variable warning treated as error.

cc -MT common/tap_attach.o -MMD -MP -MF common/tap_attach.Td -fstack-protector-strong -Wall -Werror -std=c11 -O2 -g -I/usr/home/nix/.opam/4.12.0/.opam-switch/build/solo5.0.7.2/include -c common/tap_attach.c -o common/tap_attach.o
common/tap_attach.c:99:9: error: variable 'up' set but not used [-Werror,-Wunused-but-set-variable]
    int up = 0;
        ^
1 error generated.
gmake[1]: *** [GNUmakefile:45: common/tap_attach.o] Error 1

hannesm:
The unused-but-set-variable warning was emitted for the variable "up", which is
not used on FreeBSD. To keep the code simple (not introducing further ifdefs),
just ignore it via an empty conditional. Fixes #519.

### DIFF
--- a/tenders/common/tap_attach.c
+++ b/tenders/common/tap_attach.c
@@ -162,6 +162,13 @@ int tap_attach(const char *ifname)
 
 #elif defined(__FreeBSD__)
 
+    /*
+     * Avoid unused-but-set-variable warning on FreeBSD, where the tap device
+     * is only up once open() was called by the process.
+     */
+    if (!up)
+      ;
+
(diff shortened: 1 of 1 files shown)


dinosaure:
Thanks!

PR #522, opened 2022-07-18
Be more informative when tap_attach fails

reynir:
As it is now the error message when attaching to a tap interface fails does not give any explanation as to why it failed.
```
solo5-spt: Could not attach interface: temp0
```
I reviewed the code for `tap_attach`, and in all paths where `-1` is returned either `errno` is set directly or a previous call will have set it. With this patch `strerror(errno)` is added at the end of the warning:
```
solo5-spt: Could not attach interface: temp0: Network is down
```
Now I know to run `ip link set temp0 up` :-)


reynir:
Odd. It's unfortunate that there isn't more information on the failed builds. I can't imagine how this change would fail on either platform.

PR #527, opened 2022-09-08
block_attach: check file is aligned

reynir:
Otherwise if the file is not a multiple of block_size (512) we have to make up bytes for the missing part.


reynir:
After testing it turns out that unlike mirage-block-unix solo5 doesn't conjure up magic zero bytes but instead "truncates" the block device. That is, a 1000 byte file will be seen as a block device consisting of a single 512 byte sector(!)

dinosaure:
Thanks!

PR #528, opened 2022-09-13
Custom block size

reynir:
This adds an option to specify a block size different from 512. The block size must be a power of 2 greater than or equal 512.

Marked as a draft as
- it's not well tested,
- I'm not sure about a few details and decisions.

Some things to investigate (CC @Julow):
- [ ] Does the restriction on block size make sense?
- [ ] What other components require block sizes to be >= 512 ?
- [ ] Is here anything that might misbehave if it no longer exactly 512 ?
- [ ] The mirage library has "preferred block size" to give a chance to the lower levels to enforce the block size. Should there be something similar here (get feedback from the lower levels instead of writing the rule in the parser) ?

### DIFF
--- a/tenders/common/block_attach.c
+++ b/tenders/common/block_attach.c
@@ -29,12 +29,13 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <stdint.h>
 
 /*
  * Attach to the block device specified by (path), returning its capacity in
  * bytes in (*capacity).
  */
-int block_attach(const char *path, off_t *capacity_)
+int block_attach(const char *path, uint16_t block_size, off_t *capacity_)
 {
     int fd = open(path, O_RDWR);
     if (fd == -1)
@@ -42,11 +43,12 @@ int block_attach(const char *path, off_t *capacity_)
     off_t capacity = lseek(fd, 0, SEEK_END);
     if (capacity == -1)
         err(1, "%s: Could not determine capacity", path);
-    if (capacity < 512)
-        errx(1, "%s: Backing storage must be at least 1 block (512 bytes) "
-                "in size", path);
-    if (capacity & (512 - 1))
-        errx(1, "%s: Backing storage size must be block aligned (512 bytes)", path);
+    if (capacity < block_size)
+        errx(1, "%s: Backing storage must be at least 1 block (%hu bytes) "
+                "in size", path, block_size);
+    if ((capacity & (block_size - 1)) != 0)
+        errx(1, "%s: Backing storage size must be block aligned (%hu bytes)",
+                path, block_size);
 
     *capacity_ = capacity;
     return fd;
--- a/tenders/common/block_attach.c
+++ b/tenders/common/block_attach.c
@@ -30,13 +30,15 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <stdint.h>
+#include <assert.h>
 
 /*
  * Attach to the block device specified by (path), returning its capacity in
- * bytes in (*capacity).
+ * bytes in (*capacity). The block size (block_size) must be a power of two.
  */
 int block_attach(const char *path, uint16_t block_size, off_t *capacity_)
 {
+    assert((block_size & (block_size - 1)) == 0);
     int fd = open(path, O_RDWR);
     if (fd == -1)
         err(1, "Could not open block device: %s", path);
@@ -46,6 +48,7 @@ int block_attach(const char *path, uint16_t block_size, off_t *capacity_)
     if (capacity < block_size)
         errx(1, "%s: Backing storage must be at least 1 block (%hu bytes) "
                 "in size", path, block_size);
+    /* this assumes block_size is a power of 2 */
     if ((capacity & (block_size - 1)) != 0)
         errx(1, "%s: Backing storage size must be block aligned (%hu bytes)",
                 path, block_size);
--- a/tenders/common/block_attach.c
+++ b/tenders/common/block_attach.c
@@ -45,6 +45,8 @@ int block_attach(const char *path, off_t *capacity_)
     if (capacity < 512)
         errx(1, "%s: Backing storage must be at least 1 block (512 bytes) "
                 "in size", path);
+    if (capacity & (512 - 1))
+        err(1, "%s: Backing storage size must be block aligned (512 bytes)", path);
 
     *capacity_ = capacity;
     return fd;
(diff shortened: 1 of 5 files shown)


tenders/hvt/hvt_module_blk.c
```diff
@@ -158,7 +197,8 @@ static int setup(struct hvt *hvt, struct mft *mft)
 
 static char *usage(void)
 {
-    return "--block:NAME=PATH (attach block device/file at PATH as block storage NAME)";
+    return "--block:NAME=PATH (attach block device/file at PATH as block storage NAME)\n"
+	"  [ --block-size:NAME=BLOCKSIZE ] (set block size for block device NAME; must be a power of two greater than or equal 512)";
```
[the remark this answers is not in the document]
reynir: Maybe the option should be renamed to `--block-sector-size` so as to not be confused with the total size of the block device.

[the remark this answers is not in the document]
Julow: Good idea! No more incorrect values

tenders/hvt/hvt_module_blk.c
```diff
@@ -108,30 +108,62 @@ static void hypercall_block_read(struct hvt *hvt, hvt_gpa_t gpa)
 
 static int handle_cmdarg(char *cmdarg, struct mft *mft)
 {
-    if (strncmp("--block:", cmdarg, 8) != 0)
+    enum {
+        opt_block,
+        opt_block_size,
+    } which;
+    if (strncmp("--block:", cmdarg, 8) == 0)
```
Julow, line 115: Size of constant strings can be computed at compile time:
```c
#define STARTS_WITH(prefix, str) (strncmp(prefix, cmdarg, sizeof(prefix) - 1) == 0)
```

```c
    if (STARTS_WITH("--block:", cmdarg))
```

Perhaps a cast can be used to annotate the type of `prefix` in the macro ? eg. `sizeof((char[])prefix)`, not sure if that would give the right result.

reynir: Hmm, yes, that's a good point. It's basically the same code as in `*_module_net`. I don't think we can use a cast to assert the type. Since we use the same pattern in (at least) 4 different files I wonder where it would make sense to define the macro.

Julow: `tenders/common/cc.h` might have room for such macro.

Julow: This macro doesn't misbehave but is weirder:
```c
#define STARTS_WITH(prefix, str) (strncmp(#prefix, str, sizeof(#prefix) - 1) == 0)

STARTS_WITH(--block:, cmdarg)
```
But this might be better defined as a inlineable C function. `strlen` might be evaluated at compile time but that doesn't matter anyway.

tenders/hvt/hvt_module_blk.c
```diff
@@ -106,32 +106,68 @@ static void hypercall_block_read(struct hvt *hvt, hvt_gpa_t gpa)
...
-        return -1;
-    struct mft_entry *e = mft_get_by_name(mft, name, MFT_DEV_BLOCK_BASIC, NULL);
-    if (e == NULL) {
-        warnx("Resource not declared in manifest: '%s'", name);
-        return -1;
+    uint16_t block_size;
+    if (which == opt_block) {
+        block_size = 512;
+        int rc = sscanf(cmdarg,
+                "--block:%" XSTR(MFT_NAME_MAX) "[A-Za-z0-9]="
```
Julow, line 132:
```suggestion
                BLOCK_PREFIX "%" XSTR(MFT_NAME_MAX) "[A-Za-z0-9]="
```

tenders/hvt/hvt_module_blk.c
```diff
@@ -106,32 +106,68 @@ static void hypercall_block_read(struct hvt *hvt, hvt_gpa_t gpa)
...
+        /* e->u.block_basic.block_size is set either by option or generated
+         * later by setup().
+         */
+        e->u.block_basic.capacity = capacity;
+        e->b.hostfd = fd;
+        e->attached = true;
+        module_in_use = true;
+    } else if (which == opt_block_size) {
+        int rc = sscanf(cmdarg,
+                "--block-sector-size:%" XSTR(MFT_NAME_MAX) "[A-Za-z0-9]="
```
Julow, line 153:
```suggestion
                BLOCK_SECTOR_SIZE_PREFIX "%" XSTR(MFT_NAME_MAX) "[A-Za-z0-9]="
```

reynir:
I think this is ready once we add to the usage message a warning that changing this could break atomicity.

tenders/hvt/hvt_module_blk.c
```diff
@@ -106,32 +106,68 @@ static void hypercall_block_read(struct hvt *hvt, hvt_gpa_t gpa)
...
+        if (rc != 2)
+            return -1;
+
+        struct mft_entry *e = mft_get_by_name(mft, name, MFT_DEV_BLOCK_BASIC, NULL);
+        if (e == NULL) {
+            warnx("Resource not declared in manifest: '%s'", name);
+            return -1;
+        }
+        off_t capacity;
+        int fd = block_attach(path, block_size, &capacity);
```
reynir, line 143: This is wrong. We don't know block_size yet. We should call `block_attach()` in `setup()`.

reynir: Fixed in latest commit.

tenders/common/block_attach.c
```diff
@@ -42,9 +44,6 @@ int block_attach(const char *path, off_t *capacity_)
     off_t capacity = lseek(fd, 0, SEEK_END);
     if (capacity == -1)
         err(1, "%s: Could not determine capacity", path);
-    if (capacity < 512)
-        errx(1, "%s: Backing storage must be at least 1 block (512 bytes) "
-                "in size", path);
```
dinosaure, line 47: Why this line was deleted? According to what you say, we expect a block size equal or greater than 512.

dinosaure: Ok, I see. Thanks!

reynir: We could keep the check in `block_attach()` and report `%s: Backing storage must be at least 1 block (>= 512 bytes) in size`, but then we'd have to check twice; once again when we know the block size.

The problem is if we set block size to say 1024 bytes and the file is 512 bytes - then we can't read or write to the file, but it passes this check. So we have to check (again) once we know the block size.

dinosaure:
Thanks all of you for this great work. We must notice that such work is only integrated for `hvt` and `spt`. The work seems fine as far as I can say and thanks to add more tests about that 👍 .

Issue #529, opened 2022-10-21
Support aarch64 machines with non-4k pagesizes


kit-ty-kate:
Running an hvt image on an apple M1 laptop running linux natively, I get:
```
solo5-hvt: dist/noop.hvt: Bad address
```
Looking into the matter it seems like the pagesize is hardcoded to 4k:

https://github.com/Solo5/solo5/blob/0eb8cb8f57943e7872a94e19695911a1e0b8aef6/tenders/hvt/hvt_cpu_aarch64.h#L91-L92

However M1 machines have a 16k pagesize.

PR #530, opened 2022-10-25
[PoC] Change the default pagesize on aarch64 from 4k to 16k to support apple hardware

kit-ty-kate:
This is an attempt at hotfixing https://github.com/Solo5/solo5/issues/529 for 16k aarch64 hardware as I need to have a working solo5 to test https://github.com/mirage/ocaml-solo5/pull/122 on my work laptop (apple M1 hardware).

I don't think this should be merged as-is as I think `sysconf(_SC_PAGESIZE)` should be used instead but the goal is to show which parts of the code needs to change for solo5 to work on 16k pagesize hardware.

As of afa7fdf2d09b55acdb955b5783e41f92585e2ee0: only `solo5-spt` works. `solo5-hvt` infinit loops for some reason, so something else might be needed.

### DIFF
--- a/bindings/cpu_aarch64.h
+++ b/bindings/cpu_aarch64.h
@@ -21,8 +21,8 @@
 #define __CPU_AARCH64_H__
 
 /* memory defines */
-#define PAGE_SIZE   4096
-#define PAGE_SHIFT  12
+#define PAGE_SHIFT  14
+#define PAGE_SIZE   (1 << 14)
 #define PAGE_MASK   ~(0xfff)
 
 #ifndef _BITUL
--- a/bindings/cpu_aarch64.h
+++ b/bindings/cpu_aarch64.h
@@ -22,7 +22,7 @@
 
 /* memory defines */
 #define PAGE_SHIFT  14
-#define PAGE_SIZE   (1 << 14)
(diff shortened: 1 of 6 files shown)


TImada:
For `solo5-hvt`, modification of `aarch64_setup_memory_mapping` [1] and `TCR_TG_FLAGS` [2] will be needed at least.

The current implementation creates a page table having mixed 1GB/2MB/4KB blocks. However, a page table having mixed 32MB/16KB blocks needs to be created to support Apple hardware.
(Check [3] for details)

In addition, `TG0` and `TG1` bits of `TCR_EL1` needs to be configured to use the 16KB granule size.
(Check [4] for details)

[1] https://github.com/Solo5/solo5/blob/master/tenders/hvt/hvt_cpu_aarch64.c#L38
[2] https://github.com/Solo5/solo5/blob/master/tenders/hvt/hvt_kvm_aarch64.c#L131
[3] https://developer.arm.com/documentation/den0024/a/The-Memory-Management-Unit/Translation-tables-in-ARMv8-A/Effect-of-granule-sizes-on-translation-tables
[4] https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/TCR-EL1--Translation-Control-Register--EL1-

16KB page size support requires slightly complicated work. 😢

PR #531, opened 2022-10-26
Fix build when using git worktree

reynir:
When using `git worktree` the file `.git` is a file and not a directory. The file contains a line `gitdir: .../.git/worktrees/...`.


dinosaure:
Thanks, seems corret (even if I don't have such usage).

PR #532, opened 2022-10-27
Allow more log levels

reynir:
While reading the source code I found that you can pass `--solo5:quiet` to the **guest** to set the log level to **ERROR**. This change adds `--solo5:error` (alias of `--solo5:quiet`), `--solo5:warn` and `--solo5:info`.

Calling `log_set_level(-1)` should silence all log levels *if* it compiles. I can change the semantics of `--solo5:quiet` to that if anyone can confirm whether it is safe to pass `-1`.

### DIFF
--- a/bindings/cmdline.c
+++ b/bindings/cmdline.c
@@ -22,6 +22,9 @@
 char *cmdline_parse(const char *cmdline)
 {
     const char opt_quiet[] = "--solo5:quiet";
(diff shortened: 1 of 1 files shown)


dinosaure:
Let's merge this PR (#528 already adds a new argument, so we can take this opportunity to improve the CLI). Thanks!

PR #535, opened 2022-11-04
changes for 0.7.4

hannesm:


dinosaure:
I finally made the OpenBSD 7.2 CI and the last commit fix an issue on our tests: Since https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9, `git rev-parse --show-toplevel` can not be executed by root which does not own the cloned repository... I don't have full details between `surf`, OpenBSD, `doas` and `git` but the last patch adds our cloned Git repository as a "safe" repository for the root user and let it to run `git rev-parse`.

PR #538, opened 2022-12-05
xen console: update the "to be written" count

palainp:
Previously we can get an infinite loop as `written` is the number of bytes written to the console returned by `console_write_some` (from 0 to its `len` argument) (l111), then it's compared against len for teminating the loop (l126).

The issue is that if `len` is greater than the Xen buffer size (2048B), we have to loop over the buffer and use `console_write_some` more than one time (and it will return something in the interval `[0,len-written[` which will never equals `len` (l126).

The proposed patch lower the `len` which will stand for "how many bytes I still have to write", and stops the loop when it comes to 0.


bindings/xen/console.c
```diff
@@ -102,13 +102,15 @@ void console_write(const char *buf, size_t len)
...
         return;
 
-    size_t written = 0;
+    size_t written = 0, count = 0;
     XENCONS_RING_IDX cons = LOAD_ACQUIRE(&console_ring->out_cons);
 
     do
     {
-        /* Try and put some data into the ring. */
-        written = console_write_some(&buf[written], len - written);
```
talex5, line 111: Perhaps this was supposed to be:
```c
written += console_write_some(&buf[written], len - written);
```

palainp: Of course! It's really simpler, thanks :)

avsm (approved):
Fix looks right to me. If anyone has some time, a 'console stress test' would be a useful regression test to have for all the backends.

dinosaure:
Thanks, the test is done by @hannesm, I will cut a release.

Issue #537, opened 2022-12-04
Xen console overload


palainp:
Hi, this issue is related to https://github.com/mirage/qubes-mirage-firewall/issues/166.
With the following unikernel (`config.ml` is the same as `mirage-skeleton/tutorial/hello/config.ml`):
```ocaml
open Lwt.Infix

module Hello (Time : Mirage_time.S) = struct
  let start _time =
    let cs = Cstruct.create 441 in
    
    Logs.info (fun f -> f "log: %a" Cstruct.hexdump_pp cs);
    Time.sleep_ns (Duration.of_sec 2) >>= fun () ->
    Lwt.return_unit
end
```
The unikernel leads to an infinite loop with Qubes 4.1 up to date. If I change the Cstruct size to 440, all is fine. With spt it works fine as well so this is probably related to the Xen part of solo5.

palainp:
Reading `console.c` I wonder if the "wait for xen to consume data" loop should update the `cons` position here https://github.com/Solo5/solo5/blob/17312abe47a18d145594b1bce454f7d707de8991/bindings/xen/console.c#L122 (like at l106)?

palainp:
In fact it wasn't. The infinite loop issue was caused by a loop condition that never comes to false. The PR https://github.com/Solo5/solo5/pull/538 fixes the test unikernel but it still remains an issue with log indentation:
```
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.7.1-62-g71139c2378
Solo5: Memory map: 32 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x1fffff)
Solo5:     rodata @ (0x200000 - 0x23bfff)
Solo5:       data @ (0x23c000 - 0x30cfff)
Solo5:       heap >= 0x30d000 < stack < 0x2000000
2022-12-05 09:08:55 -00:00: INF [application] log: 00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
                       00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
                       00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
                       00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
[...]

PR #539, opened 2022-12-06
Test over 4k bytes of output (to reproduce #537 #538

hannesm:
This is a test case for #537 / #538 for all targets, so we do not end up in any regression.

### DIFF
--- a/tests/test_output/GNUmakefile
+++ b/tests/test_output/GNUmakefile
@@ -1,0 +1,23 @@
+# Copyright (c) 2015-2019 Contributors as noted in the AUTHORS file
+#
+# This file is part of Solo5, a sandboxed execution environment.
+#
+# Permission to use, copy, modify, and/or distribute this software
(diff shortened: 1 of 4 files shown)


hannesm:
turns out the xen unikernels aren't tested by the debian 10 system (`ok 12 output xen # skip xl not available`), I'm wondering what the CI setup is, is this done on a physical machine with xen?

dinosaure:
> turns out the xen unikernels aren't tested by the debian 10 system (ok 12 output xen # skip xl not available), I'm wondering what the CI setup is, is this done on a physical machine with xen?

We probably missed an image of Debian with Xen indeed but I'm not a user of Xen at all 😕. I will make an issue to keep track about that. If someone has the time to explain how to install/use Xen on Debian 11, it will be super useful for me to create a new image (or update the current one).

dinosaure:
I will merge the test as is and I made an issue for the CI. Thanks!

PR #541, opened 2022-12-07
Update CHANGES for release v0.7.5

dinosaure:


CHANGES.md
```diff
@@ -1,3 +1,17 @@
+## v0.7.5 (2022-12-07)
+
+* Since MirageOS moved from PV mode to PVH, and thus replacing Mini-OS with
```
reynir, line 3: I think we can make it clearer that it only affects Xen (as far as I know).
```suggestion
* Since MirageOS moved from PV mode to PVH on Xen, and thus replacing Mini-OS with
```

palainp: I agree only target Xen (and Qubes, same as Xen as a target) is concerned. Virtio, Muen use for loops that print one char at a time, spt uses `sys_write`, and hvt uses `hvt_do_hypercall`.

Issue #543, opened 2023-02-01
-execute-only does not support intermingling data and code


adamsteen:
OpenBSD Current has added `-execute-only` flag to ld.lld by default, which means the current build of solo5 fails.

```
ld.lld: error: cannot place /home/adams/devl/solo5/toolchain/bin/../lib/x86_64-solo5-none-static/solo5_virtio.o:(.data.multiboot) into .text: -execute-only does not support intermingling data and code
cc: error: linker command failed with exit code 1 (use -v to see invocation)
```

see PR https://github.com/Solo5/solo5/pull/544

PR #544, opened 2023-02-01
On OpenBSD only build for HVT

adamsteen:
fixes [Issue 543](https://github.com/Solo5/solo5/issues/543)


hannesm:
I've read about these changes, but am wondering how/whether we could make the virtio target working on OpenBSD (maybe by passing some CFLAGS/LDFLAGS), since I believe this should still be usable...

adamsteen:
@hannesm I am unsure if I will get enough time to dig into something like this, I expect its just a linker script change, will see what I can do

dinosaure:
Thanks, I will keep your issue about `-execute-only` opens and see if people are interested to compile other targets than hvt on OpenBSD.

PR #546, opened 2023-02-07
Tls export symbols

palainp:
This PR is an follow-up of the previous TLS PR. I'm currently not sure if the exported API is clear enough or makes sense at all, I'd feel better if someone could comment that.

Now we can export TDATA as the address of the the thread variables' initialization values, LTDATA the length of the tdata section, and LTBSS the length of the tbss section.
This way, we can allocate, for each thread, an area of memory (the size is calculated with `solo5_tls_size()`), set the tp pointer to its value with `solo5_tls_tp_offset(.)`, and copy the initial value of tdata into our new memory `solo5_tls_data_offset(.)`. As an example, we can do:
```C
void* tcb1;                                      // defines a TLS block
tcb1 = calloc(solo5_tls_size(), sizeof(char));   // gets memory for it, set tbss to 0
memcpy(solo5_tls_data_offset(tcb1), TDATA, LTDATA);       // copy tdata
solo5_set_tls_base(solo5_tls_tp_offset((uintptr_t)tcb1);  // sets tp ptr
```

I've come across 2 _bugs_, the first is that if tdata is not a `PT_LOAD` section, it is not loaded by the tender, so the initial values of tdata are not copied into memory, the second is that `memcpy` can be inlined and cause errors in our TLS use case (I haven't investigated this much so far, I'm going to try it out with ocaml-solo5 to see if this could be a bigger issue).

### DIFF
--- a/include/solo5.h
+++ b/include/solo5.h
@@ -126,6 +126,64 @@ void solo5_exit(int status) __attribute__((noreturn));
  */
 void solo5_abort(void) __attribute__((noreturn));
 
+
+/*
+ * XXX(palainp): This is an incomplete implementation of TLS.
+ * It's limited to the local-exec pattern (offset directly from the TLS_BASE
+ * register).
+ * The usage is as the following for each thread:
+ *   void* tcb1;                                      // defines a TLS block
+ *   tcb1 = calloc(solo5_tls_size(), sizeof(char));   // gets memory for it
+ *   memcpy(solo5_tls_data_offset(tcb1), TDATA, LTDATA);       // copy tdata
+ *   solo5_set_tls_base(solo5_tls_tp_offset((uintptr_t)tcb1);  // sets tp ptr
+ */
+
+/* _stdata is the address begining of the "not bss" thread variables block */
+extern char _stdata[];
+#define TDATA ((void*)_stdata)
+
+/* _ltdata and _ltbss are the lengths of the tdata and tbss blocks */
+extern char _ltdata[], _ltbss[];
+#define LTDATA ((size_t)_ltdata)
+#define LTBSS ((size_t)_ltbss)
+
+/* The following define the overhead to the tdata and tbss blocks
+ * It's processor related
+ */
+#if defined(__x86_64__) || defined(__powerpc64__)
+    /* Variant II */
+    struct tcb {
+        void* tp;
+    };
+    
+    /* solo5_tls_tp_offset tells where we need to set the tp pointer next to
+     * the data block. It depends on the Variant, for Variant II it's at the end.
+     */
+    #define solo5_tls_tp_offset(x) (x + LTDATA + LTBSS)
+    #define solo5_tls_data_offset(x) (x)
+    
+    #define PPC64_TLS_OFFSET 0x7000
+
+#elif defined(__aarch64__)
+    /* Variant I */
+    struct tcb {
+        void* tp;
+        void* pad;
+    };
+    
+    /* for Vairant I it's at the begining */
+    #define tp_offset(x) (x)
+    #define solo5_tls_data_offset(x) (x + sizeof(struct tcb))
+
+#else
+#error Unsupported architecture
+#endif
+
+/*
+ * Returns the size needed for the thread local storage.
+ */
+#define solo5_tls_size() (LTDATA + LTBSS + sizeof(struct tcb))
+
 /*
  * Set the architecture-specific TLS base register to (base).
  *
--- a/bindings/lib.c
+++ b/bindings/lib.c
@@ -67,7 +67,7 @@ void *memset(void *dest, int c, size_t n)
     return dest;
 }
 
-void *memcpy(void *restrict dest, const void *restrict src, size_t n)
+void *__attribute__ ((noinline)) memcpy(void *restrict dest, const void *restrict src, size_t n)
 {
     unsigned char *d = dest;
     const unsigned char *s = src;
--- a/bindings/solo5_hvt.lds
+++ b/bindings/solo5_hvt.lds
@@ -40,7 +40,7 @@ PHDRS {
     rodata PT_LOAD FLAGS(4);
 
     data PT_LOAD;
-    tdata PT_TLS;
+    tdata PT_LOAD;
     tbss PT_TLS;
     note.not_openbsd PT_NOTE; /* Must come first. */
     note.abi PT_NOTE;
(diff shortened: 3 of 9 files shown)


palainp:
Oups it seems that something is wrong between my laptop and CI, I'll check that and update. Sorry :/

palainp:
The virtio failure (probably the same for xen) will be a bit harder to fix.

The multiboot headers, in `bindings/{virtio,xen}/boot.S`, defined as https://www.gnu.org/software/grub/manual/multiboot/multiboot.html#Header-address-fields, should explicit the end of the tdata segment (tdata have to be copied too).
I tried to fix that but I don't understand yet how is calculated the offset from %fs for both virtio and xen (it's very much not the same as for the other targets, this may be an error in the linker script again).

palainp:
Update: I'm starting to understand how ld and lld work with TLS. So far it seems that ld uses the address in the elf binary, and lld uses the memsize addresses, so with a linker script like :
```
...
    .tdata { *(.tdata) } :tdata
    . = ALIGN(CONSTANT(MAXPAGESIZE));
    _edata = .;
    .tbss { *(.tbss) } :tbss
...
```
We have with `readelf -lW`:
```
Elf file type is EXEC (Executable file)
Entry point 0x100020
There are 7 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
...
  TLS            0x005018 0x0000000000104018 0x0000000000104018 0x000008 0x011068 R   0x20
...
```
`FileSiz` is the actual size of the tdata section in the elf (1 8B value in tdata as tbss should be set to 0 at runtime), `MemSiz` is the size of the sections in memory (not sure yet how this is calculated).
We have an offset :
- about 1 page from %fs for tdata (because we have an ALIGN directive in the linker script, it was mandatory to set the _edata symbol), and enough bytes for the %fs tbss thread variables for the tbss section with ld
- a MemSiz offset for tdata, and a (MemSiz - pagesize, because ALIGN) for the tbss variables with lld

I'll try to figure out if I can give an option to ld or lld to force similar behaviour regarding the offsets.

palainp:
To my understanding, the big memsize in the readelf output was due to bss being merged by lld with the previous sections. So I added a bss section in the elf header which solve that part.
Now I have something more comprehensible:
```
...
  LOAD           0x001000 0x0000000000100000 0x0000000000100000 0x003f44 0x003f44 R E 0x1000 # <- this is text
  LOAD           0x005000 0x0000000000104000 0x0000000000104000 0x000018 0x000018 RW  0x1000 # <- this is data
  TLS            0x006000 0x0000000000105000 0x0000000000105000 0x001000 0x001000 R   0x8 # <- this is tdata
  TLS            0x005018 0x0000000000106000 0x0000000000106000 0x000000 0x000008     0x8 # <- this is tbss
  NULL           0x000000 0x0000000000106000 0x0000000000106000 0x000000 0x010080     0x20 # <- this is bss
...
```

It works on my linux laptop with ld.lld but still fails with {Free,Open}BSD CI. I'll try to re-install a BSD on a spare laptop to understand what is the failure with CI.

palainp:
Before the last commits, I didn't set any value for what was (before this PR) the `tp` field in `struct tcb` (e.g. nothing like `tcb1.tp = &tcb1.tp`). I just set the TCB address to a correct value with `solo5_set_tls_base`, and it seems that linux is ok with that (having an address for its %fs register), but BSD does an indirection on that address and so has %fs set to 0 (which explains the page fault).

That different behavior seems peculiar to me, I'll search the internet for something about that.

Now I also use some function to have compiler type checking and hiding the TCB/TLS details outside solo5, but it force me to have some casts in `tls.c` between `void*` and `uintptr_t`.

bindings/tls.c
```diff
@@ -92,7 +92,7 @@ solo5_result_t solo5_tls_init(uintptr_t tls)
 
     /* set tp at its proper place in the TLS block */
     uintptr_t *tmp = (uintptr_t*)solo5_tls_tp_offset(tls);
-    *tmp = *tmp;
```
palainp: Oops, no dereferencing here, it should have been `*tmp = (uintptr_t)tmp;` or something like that if we want to avoid a second call for `solo5_tls_tp_offset(.)`.

palainp: CI seems to be fine with that update.

palainp:
The last thing I'm worried about is the need for ` . = ALIGN(CONSTANT(MAXPAGESIZE));` in the .tdata section. I haven't found how to push the alignment out of the section yet, this is due to some tests in https://github.com/Solo5/solo5/blob/a63a755d710a1286a0d0eea1253762ba25e866b7/tenders/common/elf.c#L283 and later (I haven't kept notes on which tests failed exactly :/ ).

bindings/tls.c
```diff
@@ -20,6 +20,86 @@
...
+#else
+#error Unsupported architecture
+#endif
+
+    return data;
+}
+
+solo5_result_t solo5_tls_init(uintptr_t tls)
+{
+	if ((void*)tls == NULL) return SOLO5_R_EINVAL;
```
hannesm, line 91: the indentation looks wrong on the line above

include/solo5.h
```diff
@@ -126,6 +126,39 @@ void solo5_exit(int status) __attribute__((noreturn));
  */
 void solo5_abort(void) __attribute__((noreturn));
 
+
+/*
+ * XXX(palainp): This implementation of TLS is limited to the local-exec model
```
hannesm, line 131: I'd remove the `XXX(palainp)`

palainp:
Thank you very much for your comment. I've updated the indentation and the comment in `solo5.h`.

Right now, the API update shouldn't break anything, but you're right, there is now a piece of code that tries to write somewhere without checking that the destination size is sufficient. To me the only way around to solve this potential issue seems to bring back memory allocation into solo5 to keep control over the `malloc(solo5_tls_size())` call.

palainp:
I was also thinking of another possibility which might be to export the address and length of `.tdata` plus an offset into the thread memory and delegate the `memcpy` to the caller. As solo5 does not use the `__thread` variable, this should be safe.

dinosaure:
I am indeed inclined to merge this PR and do a release. The comment from @hannesm confirms a double code review and I'm a bit more confident about the next step. Regarding the security aspect, the idea of wrapping the area with `READ_ONLY` segments is not possible?

PR #547, opened 2023-02-11
Fix noexec test

greydot:
Since some point recently, this test fails (as it is expected to) but with a different error message. This PR adds a case for this.

Alternatively, I propose removing the error message check altogether, since it is too fragile and might be affected - among other things - by locale settings.

### DIFF
--- a/tests/tests.bats
+++ b/tests/tests.bats
@@ -196,7 +196,7 @@ xen_expect_abort() {
   run test_hello/test_hello.hvt
   case "${CONFIG_HOST}" in
   Linux)
-    [ "$status" -eq 127 ] && [[ "$output" == *"No such file or directory"* ]]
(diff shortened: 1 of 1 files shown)


ehmry:
This test might be fine for the BSDs but I don't think it can be reliable for Linux.  The error comes from userspace and Linux doesn't have one.

greydot:
>The error comes from userspace and Linux doesn't have one.

What do you mean? This error comes from glibc.

ehmry:
Right, so the test must check which version of Glibc is present rather than if the host is Linux.

greydot:
I don't think this is feasible. We could either include all variants -- so far there seem to be only two -- or just rely on the exit code. I'd vote for the latter.

dinosaure:
Thanks!

Issue #549, opened 2023-02-22
Clock drift


reynir:
Many of the backends including xen, hvt and virtio get the wall clock time only at boot and then compute an offset used for computing the wall clock time in `solo5_clock_wall`. This minimizes the number of hypercalls, but means the guest never synchronizes with the host's wall clock. It is most noticeable on a host such as a laptop where the computer is suspended, but can also happen due to natural clock drift.

https://github.com/Solo5/solo5/blob/a63a755d710a1286a0d0eea1253762ba25e866b7/bindings/hvt/tscclock.c#L110-L120

https://github.com/Solo5/solo5/blob/a63a755d710a1286a0d0eea1253762ba25e866b7/bindings/hvt/time.c#L33-L37

There are a number of strategies imaginable to mitigate clock skew (e.g. recompute the wall clock offset every set interval according to the monotonic clock), but it is perhaps best decided in client code how to approach that.

A way forward could be to expose the hypercall and leave it to the client code (mirage-clock) to keep an offset in order to not make hypercalls as often.

dinosaure:
After a discussion with @reynir reynir, we agreed on what the Solo5 clock is. This clock is a monotonic clock. That is to say that it is mainly used to have the CPU time.

It is therefore normal, when suspending a microkernel, that this clock is no longer aligned with what we expect from a wall clock.

In other words, if you expect a clock that corresponds to the "real elapsed time", you should not use the clock proposed by Solo5.

However, there are several solutions:
1) The first solution would be for Solo5 to manage a wall-clock (in addition to the monotonic clock) by regularly making a `HYPERCALL_WALLTIME` hypercall.
2) The second solution is to expose this hypercall and let the user manage his/her own wall-clock
3) The third solution, which is more about the interest of having a POSIX clock, would be to make a regular NTP request and synchronize an internal POSIX clock.

The design of Solo5 is made to be as simple as possible and integrating a new clock goes against this philosophy. Especially since the interest of a wall-clock for unikernels is not systematic - some of my unikernels can work "in the past" without it disturbing their functioning.

However, the interest of having a wall-clock may be necessary in other situations. In this respect, the second solution can be interesting to let the user manage his/he own wall-clock.

Finally, the last solution, much more complex since it requires a network stack, is still interesting but this complexity should relate, in our case, to MirageOS through the "device" mechanisms.

[...]

Issue #540, opened 2022-12-07
The CI matrix is incomplete


dinosaure:
Currently, our CI does not allow tests with Xen because `xl` (the hypervizor) is not available on our images. We probably should update, at least, the Debian 11 image with the Xen support.

PR #550, opened 2023-03-22
Use the hypercall instead of the TSC-based clock to implement the wall clock

dinosaure:
/cc @reynir. See #549 for more information.

### DIFF
--- a/bindings/virtio/time.c
+++ b/bindings/virtio/time.c
@@ -42,7 +42,7 @@ solo5_time_t solo5_clock_wall(void)
     if (use_pvclock)
         return pvclock_monotonic() + pvclock_epochoffset();
     else
-        return tscclock_monotonic() + tscclock_epochoffset();
+        return (rtc_gettimeofday());
 }
 
 static int timer_handler(void *arg __attribute__((unused)))
--- a/bindings/hvt/time.c
+++ b/bindings/hvt/time.c
@@ -33,5 +33,7 @@ uint64_t solo5_clock_monotonic(void)
 /* return wall time in nsecs */
 uint64_t solo5_clock_wall(void)
 {
-    return tscclock_monotonic() + tscclock_epochoffset();
+    struct hvt_hc_walltime t;
+    hvt_do_hypercall(HVT_HYPERCALL_WALLTIME, &t);
+    return t.nsecs;
 }
--- a/bindings/hvt/tscclock.c
+++ b/bindings/hvt/tscclock.c
@@ -20,9 +20,6 @@
 
 #include "bindings.h"
 
-/* Wall clock offset at monotonic time base. */
-static uint64_t wc_epochoffset;
-
 /* Base time values at the last call to tscclock_monotonic(). */
 static uint64_t time_base;
 static uint64_t tsc_base;
@@ -45,7 +42,7 @@ static uint32_t tsc_mult;
 #endif
 
 /*
- * Beturn monotonic time using TSC clock.
+ * Return monotonic time using TSC clock.
  */
 uint64_t tscclock_monotonic(void)
 {
@@ -107,25 +104,5 @@ int tscclock_init(uint64_t tsc_freq)
     tsc_base = READ_CPU_TICKS();
     time_base = mul64_32(tsc_base, tsc_mult, tsc_shift);
 
-    /*
-     * Compute wall clock epoch offset by subtracting monotonic time_base from
-     * wall time at boot.
-     *
-     * TODO: This arrangement minimises the use of hypercalls, but is subject
-     * to clock skew over time and cannot get corrections from the host (via
-     * e.g. NTP). 
-     */
-    struct hvt_hc_walltime t;
-    hvt_do_hypercall(HVT_HYPERCALL_WALLTIME, &t);
-    wc_epochoffset = t.nsecs - time_base;
-
     return 0;
-}
-
-/*
- * Return epoch offset (wall time offset to monotonic clock start).
- */
-uint64_t tscclock_epochoffset(void)
-{
-	return wc_epochoffset;
 }
--- a/bindings/virtio/bindings.h
+++ b/bindings/virtio/bindings.h
@@ -46,7 +46,7 @@ uint64_t pvclock_epochoffset(void);
 /* tscclock.c: TSC/PIT-based clock and sleep */
 int tscclock_init(void);
 uint64_t tscclock_monotonic(void);
-uint64_t tscclock_epochoffset(void);
+uint64_t rtc_gettimeofday(void);
 void cpu_block(uint64_t until);
 
 /* pci.c: only enumerate for now */
--- a/bindings/virtio/tscclock.c
+++ b/bindings/virtio/tscclock.c
@@ -70,9 +70,6 @@
 #define	RTC_STATUS_A	0x0a
 #define	RTC_UIP		(1<<7)
 
-/* RTC wall time offset at monotonic time base. */
-static uint64_t rtc_epochoffset;
-
 /*
  * TSC clock specific.
  */
@@ -144,7 +141,7 @@ static inline uint8_t rtc_read(uint8_t reg) {
  * Return current RTC time. Note that due to waiting for the update cycle to
  * complete, this call may take some time.
  */
-static uint64_t rtc_gettimeofday(void) {
+uint64_t rtc_gettimeofday(void) {
     struct bmk_clock_ymdhms dt;
 
     cpu_intr_disable();
@@ -189,18 +186,12 @@ uint64_t tscclock_monotonic(void) {
  * Calibrate TSC and initialise TSC clock.
  */
 int tscclock_init(void) {
-    uint64_t tsc_freq, rtc_boot;
+    uint64_t tsc_freq;
 
     /* Initialise i8254 timer channel 0 to mode 2 at 100 Hz */
     outb(TIMER_MODE, TIMER_SEL0 | TIMER_RATEGEN | TIMER_16BIT);
     outb(TIMER_CNTR, (TIMER_HZ / 100) & 0xff);
     outb(TIMER_CNTR, (TIMER_HZ / 100) >> 8);
-
-    /*
-     * Read RTC "time at boot". This must be done just before tsc_base is
-     * initialised in order to get a correct offset below.
-     */
-    rtc_boot = rtc_gettimeofday();
 
     /*
      * Calculate TSC frequency by calibrating against an 0.1s delay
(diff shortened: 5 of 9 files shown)


bindings/muen/time.c
```diff
@@ -0,0 +1,37 @@
...
+uint64_t solo5_clock_monotonic(void)
+{
+    return tscclock_monotonic();
+}
+
+/* return wall time in nsecs */
+uint64_t solo5_clock_wall(void)
+{
+    return tscclock_monotonic() + tscclock_epochoffset();
+}
```
reynir, line 37: looks like we need to make a similar change for muen

bindings/virtio/time.c
```diff
@@ -42,7 +42,7 @@ solo5_time_t solo5_clock_wall(void)
     if (use_pvclock)
         return pvclock_monotonic() + pvclock_epochoffset();
```
reynir, line 43:
```suggestion
        return pvclock_read_wall_clock();
```

bindings/muen/time.c
dinosaure: Yes, `muen` is the only one when I did not really change the function. It seems a bit more complex to communicate with the host and ask the wall clock - I don't want to introduce a privilege escalation when `muen` should assert some accesses.

bindings/virtio/time.c
dinosaure: Yes, I will delete your commit. With PVH, a memory area is assigned to contains the real wall clock of the host system and we are allowed to read into it.

Kensan:
> This PR is ready to merge however, the test does bot say if the muen change is good or not. I'm not an expert of Muen. @Kensan it will be nice if you can take a look and be sure that everything is good on your side 👍.

I will try to take some time to look over the changes wrt. Muen later today.

bindings/muen/time.c
```diff
@@ -33,5 +35,25 @@ uint64_t solo5_clock_monotonic(void)
 /* return wall time in nsecs */
 uint64_t solo5_clock_wall(void)
 {
-    return tscclock_monotonic() + tscclock_epochoffset();
+  const struct muen_resource_type *const
```
Kensan, line 38: I think indentation is not correct. Should be 4 spaces AFAIK.

bindings/muen/time.c
```diff
@@ -33,5 +35,25 @@ uint64_t solo5_clock_monotonic(void)
...
+  uint64_t epoch = 0;
+  cc_barrier();
+
+  do
+  {
+    epoch = time_info->tsc_time_base * 1000;
+    cc_barrier();
+  } while (epoch == 0);
+
+  return epoch;
```
Kensan, line 58: Since "epoch" is always returned unaltered, without adding the current tscclock offset as before, it will return the same value for the lifetime of the Solo5 unikernel, i.e. time stands still. While the old implementation is a kludge and incorrect itself, it is still preferable, since at least `solo5_clock_wall` returns an increasing value.

Kensan:
I have added some comments regarding the Muen-related changes. The "proper" way would be to fix and simplify the time/clock implementation for Muen, which I can take care of. However, I think I will only have time to finish the changes this coming weekend. If you would like to get this merged ASAP, I would recommend to drop a9e551325084. I can then rework the time handling for Muen on top and then submit it as a separate merge request. Maybe that makes the most sense either way, what do you think?

bindings/muen/time.c
dinosaure: Yes, that's what I thought. Again, I don't know about Muen but is it exist a way to get the host's wall clock system via Muen (something else than the `tsc_time_base`)? We can wait more to have something which works perfectly :+1:. The idea is to delete the TSC-based clock and replace it by the right "hypercall".

To explain, we noticed a clock dritf on MirageOS and, for certain applications, it can be a real problem.

bindings/muen/time.c
dinosaure: Yes, I will do a lint at the end of this PR.

Kensan:
I have reworked/simplified the clock implementation in the Muen binding, see https://github.com/codelabs-ch/solo5/tree/wall-clock-muen . The changes were tested on a Muen system running in QEMU and also on hardware (with a few tweaks, I will be submitting some fixes separately).

Feel free to include the two Muen commits in this PR.

Kensan:
> Again, the purpose of this PR is to (on `hvt`/`xen`/`virtio`) do the hypercall whenever the user requests the wall clock (and get the wall clock from the host instead of keeping an epoch and just getting the monotonic clock). We decided to do this because we noticed a drift in the wall clock when the unikernel is resumed.

What is the exact cause of the clock drift that is at the root of this PR? I suspect it is because the TSC stops counting when a system is put in a sleep state, is that correct?

I ask because a difference to other bindings is, that Muen requires hardware with an invariant TSC and more importantly: suspend/resume is currently not supported.

> So the question is: do we have a way to ask the hypervisor via Muen for the host's wall clock at any time? I'm not against keeping the same behaviour as before (including clock drift) for Muen (but this will require a special note to say that we haven't fundamentally changed Muen support if we compare it to other targets).

There is no such hypercall or service. What is can though is, that the time information may be updated asynchronously. That is why the time_info struct is now volatile. I think under the specific conditions of Muen Systems, the wallclock is and should remain accurate (We did some improvements a while ago to reduce potential for clock drift).

Kensan:
Also, since the Muen changes I prepared do not reflect the subject of this PR, I can submit them separately after this PR (without [a9e5513](https://github.com/Solo5/solo5/pull/550/commits/a9e551325084462c2674760760ed68a753c421c7)) .

dinosaure:
> What is the exact cause of the clock drift that is at the root of this PR? I suspect it is because the TSC stops counting when a system is put in a sleep state, is that correct?
> 
> I ask because a difference to other bindings is, that Muen requires hardware with an invariant TSC and more importantly: suspend/resume is currently not supported.

> There is no such hypercall or service. What is can though is, that the time information may be updated asynchronously. That is why the time_info struct is now volatile. I think under the specific conditions of Muen Systems, the wallclock is and should remain accurate (We did some improvements a while ago to reduce potential for clock drift).

Thank you for this clarification :+1:. So let's leave the same behaviour for Muen then plus your latest patches. Sorry for the questions and my non-expertise with Muen but we finally found a solution that works for everyone. I will add a note in the documentation to point out to the user that as soon as Muen is involved, the wall-clock is TSC-based.

> Also, since the Muen changes I prepared do not reflect the subject of this PR, I can submit them separately after this PR (without [a9e5513](https://github.com/Solo5/solo5/pull/550/commits/a9e551325084462c2674760760ed68a753c421c7)) .

Yes, I will delete the last commit and accept your last two patches regarding the use of `volatile`.

dinosaure:
Ok, now the last question is about Xen (the CI does not test the Xen bindings - see #540). Someone (/cc @palainp?) can try this patch with Xen and check if the wall-clock works correctly (mirage/mirage-clock#57 can help to test this version of Solo5 and MirageOS)?

palainp:
Hi @dinosaure, I use Xen in a Qubes environnment (so it might onboard some side effect?), I hope to have done the right thing (the unikernel is the one proposed by reynir in mirage-clock#57). Unfortunately it doesn't seem to work out of the box (I'm with Ocaml5 and that can also have a side effect?), note that 04/17 is my laptop starting boot day (yesterday):
```
[2023-04-18 15:19:30] Logfile Opened
[2023-04-18 15:19:31] Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
[2023-04-18 15:19:31]             |      ___|
[2023-04-18 15:19:31]   __|  _ \  |  _ \ __ \
[2023-04-18 15:19:31] \__ \ (   | | (   |  ) |
[2023-04-18 15:19:31] ____/\___/ _|\___/____/
[2023-04-18 15:19:31] Solo5: Bindings version v0.7.5-26-g2961161069
[2023-04-18 15:19:31] Solo5: Memory map: 32 MB addressable:
[2023-04-18 15:19:31] Solo5:   reserved @ (0x0 - 0xfffff)
[2023-04-18 15:19:31] Solo5:       text @ (0x100000 - 0x236fff)
[2023-04-18 15:19:31] Solo5:     rodata @ (0x237000 - 0x254fff)
[2023-04-18 15:19:31] Solo5:       data @ (0x255000 - 0x32afff)
[2023-04-18 15:19:31] Solo5:       heap >= 0x32b000 < stack < 0x2000000
[2023-04-18 15:19:31] 2023-04-17 06:05:05 -00:00: INF [application] 2023-04-17 06:05:05 -00:00
[2023-04-18 15:19:32] 2023-04-17 06:05:05 -00:00: INF [application] 2023-04-17 06:05:05 -00:00
[2023-04-18 15:19:33] 2023-04-17 06:05:05 -00:00: INF [application] 2023-04-17 06:05:05 -00:00
....
```
The Qubes logs (first column) are increased correctly, but the mirage timestamps are freezed.
In any case, if that can be helpful:
```
$ cat unikernel.ml
open Lwt.Infix

module Hello (Time : Mirage_time.S) (Pclock : sig include Mirage_clock.PCLOCK val refresh : unit -> unit end) = struct
  let start _time _pclock =
    let rec loop () =
      Pclock.refresh ();
      Logs.info (fun f -> f "%a" (Ptime.pp_human ()) (Ptime.v (Pclock.now_d_ps ())));
      Time.sleep_ns (Duration.of_sec 1) >>= fun () -> loop ()
    in
    loop ()
end
$ cat config.ml 
open Mirage

let main = main ~packages:[ package "duration" ] "Unikernel.Hello" (time @-> pclock @-> job)
let () = register "hello" [ main $ default_time $ default_posix_clock ]
$ opam pin
mirage-clock.4.2.0                       git  git+https://github.com/reynir/mirage-clock.git#clock-experiment
mirage-clock-solo5.4.2.0                 git  git+https://github.com/reynir/mirage-clock.git#clock-experiment
mirage-clock-unix.4.2.0                  git  git+https://github.com/reynir/mirage-clock.git#clock-experiment
ocaml-solo5.0.8.1                        git  git+https://github.com/mirage/ocaml-solo5.git#500-cleaned
solo5.0.7.5                              git  git+https://github.com/solo5/solo5.git#wall-clock
```

dinosaure:
I looked into https://xenbits.xen.org/docs/4.8-testing/misc/pvh.html and it seems that we must keep a TSC-based wall-clock. I will revert my patch on Xen and merge this PR. Finally, the Muen and the Xen bindings don't change.

Kensan:
> Thank you for this clarification 👍. So let's leave the same behaviour for Muen then plus your latest patches. Sorry for the questions and my non-expertise with Muen but we finally found a solution that works for everyone. I will add a note in the documentation to point out to the user that as soon as Muen is involved, the wall-clock is TSC-based.

Thanks and certainly no need to say/feel sorry. IMHO it's good to discuss things and have an explanation for the why etc.

PR #553, opened 2023-04-18
Improve `muen` clock

dinosaure:
/cc @Kensan


Kensan:
A bit late but just for the record: I tested it again on Muen+Hardware and it works as expected. Thanks for merging the changes!

PR #555, opened 2023-04-25
Fix the compilation of solo5-aarch64 and the TLS support

dinosaure:
/cc @palainp can you confirm that it's the right fix for `aarch64`?

### DIFF
--- a/bindings/tls.c
+++ b/bindings/tls.c
@@ -78,6 +78,7 @@ uintptr_t _solo5_tls_data_offset(uintptr_t tls)
 #if defined(__x86_64__) || defined(__powerpc64__)
     data = tls;
 #elif defined(__aarch64__)
+    data = tls;
     data = data + sizeof(struct tcb);
 #else
 #error Unsupported architecture


palainp:
Oh sorry I missed that thank you! I unfortunately can't test it due to lack of hardware :(

It looks good to me, perhaps you'd merge both liges into single one ?

dinosaure:
> It looks good to me, perhaps you'd merge both liges into single one ?

I think it's fine. About testing, we still are under the experimental state about the `aarch64` support :+1:. Thanks for the quick review.

bindings/tls.c
```diff
@@ -78,6 +78,7 @@ uintptr_t _solo5_tls_data_offset(uintptr_t tls)
 #if defined(__x86_64__) || defined(__powerpc64__)
     data = tls;
 #elif defined(__aarch64__)
+    data = tls;
     data = data + sizeof(struct tcb);
```
hannesm, line 82: shouldn't these two lines be `data = tls + sizeof(struct tcb);`? The `data = data..` looks fishy.

dinosaure: I will push on `master` directly with your change indeed :+1:.

PR #557, opened 2023-04-27
Support for Solo5 unikernel lifecycle management on Muen & ABI increase

Kensan:
The first two commits make it possible to (re-)use the existing Subject Lifecycle Management of Muen and also apply it to Solo5 unikernels. A possible use case is to configure automatic restarting of unikernels that invoke `solo5_exit()`.

Increasing the ABI version is necessary due to the TLS changes. @dinosaure, maybe you prefer to squash this change into https://github.com/Solo5/solo5/pull/556/commits/806d40ab98595a3f7e6738d8bc29dc036fd111b7, which would be fine by me.

Finally, the adjustment to `test_fpu` has become necessary to successfully run the test on Muen, otherwise a #GP exception is raised. I guess we were lucky with alignment on other x86 bindings. Note that I have only verified that the test works on Muen but CI should show us if it unexpectedly broke `hvt` or any of the others.

Addendum: I manually ran the tests on Muen and the test that are expected to pass ran successfully.

### DIFF
--- a/bindings/muen/muen-platform_lifecycle.c
+++ b/bindings/muen/muen-platform_lifecycle.c
@@ -19,10 +19,23 @@
  */
 
 #include "bindings.h"
+#include "sinfo.h"
 
 const struct mft *muen_manifest = NULL;
 
 extern const struct mft1_note __solo5_mft1_note;
+
+void trigger_exit_event()
+{
+    const struct muen_resource_type *const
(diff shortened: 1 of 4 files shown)


dinosaure:
Thanks for your fix, I will redo the `0.8.0` release which is not yet published in OPAM. I will write down an email to say that we change the _tarball_ to include your changes :+1:.

PR #561, opened 2023-12-08
Be able to use `-dumpmachine` on our toolchain

dinosaure:
This little patch allows us to call `cc` from our toolchain with the `-dumpmachine` argument. By this way, we are able to recognize, at least, the assembler produced by our toolchain. It's related to mirage/ocaml-solo5#128 and mirage/ocaml-gmp#22.

### DIFF
--- a/toolchain/cc.in
+++ b/toolchain/cc.in
@@ -62,6 +62,9 @@ for arg do
     fi
 
     case "$arg" in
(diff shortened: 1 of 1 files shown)


dinosaure:
Not needed for `ocaml-solo5`/`ocaml-gmp`.

PR #563, opened 2024-01-06
Fix prototype for solo5_tls_size.

greydot:
This prevents a warning when built with -Wstrict-prototypes.


palainp (approved):
Thanks I appreciate!

hannesm (approved):

dinosaure (approved):
Thanks, I will integrate that into the next release.

PR #565, opened 2024-01-08
opam-release: prevent release if there's "-Werror" in Makefile.common

hannesm:
//cc @palainp as discussed in #564

### DIFF
--- a/scripts/opam-release.sh
+++ b/scripts/opam-release.sh
@@ -3,6 +3,9 @@
 # This script is used to generate OPAM metadata for publishing a Solo5 release
 # to ocaml/opam-repository. It should be run from the root of a clean, tagged
 # Solo5 git repository.
+# From Makefile.commmon, the '-Werror' should be removed for the tag of the
+# release, to avoid failures with future C compilers (which may add new
+# warnings).
 
 if [ ! -d "./opam" ]; then
     echo "ERROR: missing ./opam. Run this from the root of a Solo5 repository."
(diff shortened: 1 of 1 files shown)


scripts/opam-release.sh
```diff
@@ -18,6 +21,13 @@ if ! echo "$GIT_VERSION" | egrep -q '^v[0-9]+.[0-9]+.[0-9]+$'; then
     DEV=~dev
 fi
 
+WERROR=$(git grep -c ' -Werror' Makefile.common | cut -d ':' -f 2)
+if [ "$WERROR" -gt 0 ]; then
+    echo "WARNING: There are occurences of '-Werror' in 'Makefile.common'."
+    echo "WARNING: This is almost certainly not what you want."
+    exit 1
```
palainp, line 28: LGTM. To be nitpicky, as the script exits with an error code, maybe echo _ERROR:_ instead of _WARNING:_?

palainp (approved):

scripts/opam-release.sh
hannesm: thanks, addressed in [f861e0a](https://github.com/Solo5/solo5/pull/565/commits/f861e0a5b2e037adbc8d6ee2ffc06fc1a38101e8)

palainp (approved):

dinosaure (approved):
I'm fine with that, let's do the next release with it.

Issue #548, opened 2023-02-18
bug with hvt debug


a74731248:
When I try to use live debugging of hvt unikernels, I meet some trouble. When I use  command `gdb --ex="target remote localhost:1234" test_hello.hvt`  connect to gdb, the another terminal exit with info as fellows:
```
root@9fbe6a174ca1:~/solo5/tests/test_hello# ../../tenders/hvt/solo5-hvt-debug --gdb test_hello.hvt
solo5-hvt-debug: Waiting for a debugger. Connect to it like this:
solo5-hvt-debug:        gdb --ex="target remote localhost:1234" KERNEL
solo5-hvt-debug: Connection from debugger at 127.0.0.1
solo5-hvt-debug: GDB: Exiting as we could not receive the next command from the debugger.
```

greydot:
Can confirm. GDB 12.1.

reynir:
With gdb 10.1 it works for me FWIW
```
$ solo5-hvt-debug --gdb hello.hvt
solo5-hvt-debug: Waiting for a debugger. Connect to it like this:
solo5-hvt-debug: 	gdb --ex="target remote localhost:1234" KERNEL
solo5-hvt-debug: Connection from debugger at 127.0.0.1
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.7.3
Solo5: Memory map: 512 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x1dffff)
Solo5:     rodata @ (0x1e0000 - 0x215fff)
Solo5:       data @ (0x216000 - 0x2c0fff)
Solo5:       heap >= 0x2c1000 < stack < 0x20000000
2023-02-28 06:58:18 -00:00: INF [application] hello
2023-02-28 06:58:19 -00:00: INF [application] hello
2023-02-28 06:58:20 -00:00: INF [application] hello
2023-02-28 06:58:21 -00:00: INF [application] hello
Solo5: solo5_exit(0) called
```
What does GDB say?

a74731248:
> With gdb 10.1 it works for me FWIW
> 
> ```
> $ solo5-hvt-debug --gdb hello.hvt
> solo5-hvt-debug: Waiting for a debugger. Connect to it like this:
> solo5-hvt-debug: 	gdb --ex="target remote localhost:1234" KERNEL
[...]

PR #567, opened 2024-02-25
Fix that GDB HVT bug

greydot:
https://github.com/Solo5/solo5/issues/548

Took me a while to get to this, mostly because GDB wire format is weird.

### DIFF
--- a/tenders/hvt/hvt_gdb_x86_64.h
+++ b/tenders/hvt/hvt_gdb_x86_64.h
@@ -26,18 +26,9 @@
 #define HVT_GDB_X86_64_H
 
 /*
- * XXX: Can't find any gdb include file with the list of registers per
- * architecture (something like ia64_regs.h).  The closest I can get is a
- * list of the registers from gdb (debugging an ordinary x86_64 binary):
- *
- * (gdb) info registers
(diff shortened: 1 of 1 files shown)


greydot:
P.S. I also implemented KVM_GET_FPU code, but it's not really tested and I'm not sure whether I should include it in this PR.

dinosaure:
> P.S. While debugging this, I also implemented KVM_GET_FPU code, but it's not really tested in the field (seems to work fine tho) and I'm not sure whether I should include it in this PR.

The best is probably to make another PR to review on it :+1:. Otherwise, this PR looks fine to me.

reynir:
Hi!

It's been about a year since I last looked at this. From reading my comments in #548 it seems we should maybe support the `qSupported` GDB debug protocol query. If I remember correctly that way we can tell GDB the register layout(?!) but then again I remember it was very difficult to find any documentation on the protocol.

Looking at the XML file it seems this change is correct. Maybe GDB can guess the layout just from the length of the hex string?? I'm in favor of merging this as this is at least less wrong.

reynir (approved):

dinosaure:
Thanks!

PR #570, opened 2024-02-28
Do not discard .gnu.note, GNU ld complains when we link object files

### LINKED ISSUE #562: GNU ld wants executable stack
Hi devs, I recently noticed that GNU ld complains about executable stack with the following warning:
```
x86_64-solo5-none-static/solo5_xen.o: requires executable stack (because the .note.GNU-stack section is executable)
```

As there is a `/DISCARD/` section in the linker scripts https://github.com/Solo5/solo5/blob/8d8517b12dbfdbe1e2e61e9a2fa07a2038a6b8b7/bindings/solo5_hvt.lds#L155, maybe it possible/needed to also add `.note.GNU-stack`?
I don't currently have access to my laptop, but it may be possible to find where's needed executable stack following https://www.redhat.com/en/blog/linkers-warnings-about-executable-stacks-and-segments?

dinosaure:
Fix #562 (/cc @palainp)


dinosaure:
Yes, I agree to put the `.note.GNU`into assembly code and discard it at the link time :+1:.

> That warning comes from the caml compiler directory, so it might be relevant to also investigate there.

You mean with `ocaml-solo5`? Not sure to understand well this warning. The warning will be deleted then in next versions of the linker?

palainp:
> You mean with `ocaml-solo5`? Not sure to understand well this warning. The warning will be deleted then in next versions of the linker?

The only `amd64.o` I've found is in `~/.opam/default/.opam-switch/build/ocaml-solo5.xxx/ocaml/runtime/` and ocaml has assembly files (e.g. https://github.com/ocaml/ocaml/blob/trunk/runtime/amd64.S). It seems that this warning has been recently added and reading https://discuss.ocaml.org/t/ld-error-missing-note-gnu-stack-section/12478.

So to me this is unrelated to solo5, the commit proposed by @Kensan LGTM.

To move forward I wonder if the best is to add `(flags (:standard -cclib "-z noexecstack"))` in `ocaml-solo5` (if possible) or PR at ocaml to add a similar `.section .note.GNU-stack,"",%progbits` in the asm files?

dinosaure:
> To move forward I wonder if the best is to add (flags (:standard -cclib "-z noexecstack")) in ocaml-solo5 (if possible) or PR at ocaml to add a similar .section .note.GNU-stack,"",%progbits in the asm files? EDIT: As it seems there's no issue about that in the ocaml repository, it's probably relevant to ocaml-solo5 :)

So, I will let this PR opens but we probably should check if the warning continues to appear if we update `ocaml-solo5`.

hannesm:
Is there any solution for this PR? Should it be included in solo5, or can it be left out (and closed)?

shym:
I would suggest to follow the same strategy than what was merged in the compiler in [ocaml/ocaml#13735], namely to do just as the C compiler does: the logic the C compiler uses to determine whether it should output the `.note.GNU-stack` is more complex than it appears (I looked into `clang` at least, I don’t remember for GCC) and they are surprising special cases (`%` is not always the symbol chosen by the compiler, for instance), so copying the C compiler output is a lot safer.
So, we could do so by adding in `bindings/GNUmakefile`:
```make
noexecstack.c:
	printf 'int f() {\n  return 0;\n}\n' > $@

noexecstack.s: noexecstack.c
	$(CC) $(CFLAGS) $(CPPFLAGS) -S $< -o $@

noexecstack.h: noexecstack.s
	grep -F .note.GNU-stack $< > $@
```
and adding `#include "noexecstack.h"` in the `.S` files instead of using directly the note.

dinosaure:
Thanks for your solution. I still think that this PR is not yet needed for Solo5 as we discovered. I will still keep this PR open the time that we get the same issue again but let's keep Solo5 simple as long as we don't need to deal with `.note.GNU-stack`.

shym:
The issue is really in Solo5. The patches to OCaml fixed this issue in the code generated by the compiler. The fix in the compiler is doing something similar to what I wrote in my comment above: it calls the C compiler and mimicks its behaviour later on.

hannesm:
@shym oh. or maybe you have a good idea for how to patch solo5? It may be based on @Kensan branch, or your idea from above. Any PR that fixes the warning would be great.

PR #573, opened 2024-04-12
Initial support for OpenBSD 7.5

shym:
This PR is a WIP as tests are failing but it makes at least the build succeed on OpenBSD 7.5.
Credits to @dustanddreams for identifying the breaking changes.

All the tests errors I see now are:

```
ld.lld: error: relocation refers to a symbol in a discarded section: __retguard_546
>>> defined in /home/build/solo5/toolchain/bin/../lib/x86_64-solo5-none-static/solo5_hvt.o
```

or similar with `solo5_stub.o`.

I think I’ve read somewhere similar errors being tackled previously. Any idea for a fix?

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -109,14 +109,16 @@ EOM
 
 cc_check_header()
 {
-    ${CC} ${PKG_CFLAGS} -x c -o /dev/null - <<EOM >/dev/null 2>&1
-#include <$@>
+    (
+        for h in "$@"; do echo "#include <$h>"; done
+        cat << EOM
 
 int main(int argc, char *argv[])
 {
     return 0;
 }
 EOM
+    ) | ${CC} ${PKG_CFLAGS} -x c -o /dev/null - >/dev/null 2>&1
 }
 
 cc_check_lib()
--- a/configure.sh
+++ b/configure.sh
@@ -284,6 +284,15 @@ case ${HOST_CC_MACHINE} in
     amd64-*openbsd*)
         CONFIG_HOST_ARCH=x86_64 CONFIG_HOST=OpenBSD
         CONFIG_HVT_TENDER=1
+        echo "${prog_NAME}: Checking for dev/vmm/vmm.h availability: "
+        if CC="${HOST_CC}" PKG_CFLAGS="" \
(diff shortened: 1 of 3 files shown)


adamsteen:
I did the initial port, I expect the ‘ret_guard’ protections need to be turned off in the linker or compiler, I don’t have the time to look too deeply atm, but will have a read over the next week to see what I find

adamsteen:
Just read [clang-local](https://man.openbsd.org/clang-local)

`clang includes the retguard security feature on amd64, arm64, mips64, powerpc and powerpc64. This feature can be disabled with the -fno-ret-protector or -fno-stack-protector options.`

Should be enough to add `-fno-ret-protector` to ‘configure.sh’, oops look like we already do https://github.com/Solo5/solo5/blob/master/configure.sh#L438

sorry not sure, seems like something in the chain does not have `-fno-ret-protector`, that’s where I would look

hannesm:
What is the status of this PR? Also, is there any update with OpenBSD 7.6 now released?

shym:
I had no time to make any progress on that work.
(I imagine the changes in that PR will be required anyway even if they are not enough.)

omegametabroccolo:
I got solo5 to build + tests to pass on OpenBSD-current. On top of the changes proposed in this PR I:

- removed uses of `vrp_continue` and `vrp_irq` in `tenders/hvt/hvt_openbsd_x86_64.c`, since they were removed from the `struct vm_run_params` data structure
- added `--Wl,--no-execute-only` to every target that does some linking in `tests/Makefile.tests`
- ran `configure.sh` with `TARGET_OBJCOPY=llvm-objcopy`, which comes with the base OpenBSD install

After that, `gmake` runs smoothly. I don't know how you'd want to integrate these changes, but feel free to ask me anything.

adamsteen:
It's great to see OpenBSD working again! I wish you well with the rest of the Mirage ecosystem.

omegametabroccolo:
Should I maybe open a new pull request? I'm not very familiar with how things are done on github.

dinosaure:
> Should I maybe open a new pull request? I'm not very familiar with how things are done on github.

Yes, a new pull-request will be good for us. Thanks!

PR #574, opened 2024-04-18
Fetch `stdatomic.h` from system on FreeBSD

shym:
This PR adds `stdatomic.h` to the set of headers that must be copied over from `/usr/include` on FreeBSD because it is not included in the clang resources on that system.

`stdatomic.h` is not listed in the C standard as a freestanding header but many atomic operations (in particular all the atomic operations the OCaml runtime uses) are really compiler builtins.

### DIFF
--- a/toolchain/gen-headers.sh
+++ b/toolchain/gen-headers.sh
@@ -72,6 +72,7 @@ if CC=${CONFIG_TARGET_CC} cc_is_clang; then
         FreeBSD|OpenBSD)
             SRCDIR=/usr/include
             SRCS="float.h stddef.h stdint.h stdbool.h stdarg.h"
+            [ "${CONFIG_HOST}" = "FreeBSD" ] && SRCS="${SRCS} stdatomic.h"
             DEPS="$(mktemp)"
             CC=${CONFIG_TARGET_CC} cc_get_header_deps ${SRCDIR} ${SRCS} \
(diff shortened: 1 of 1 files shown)


hannesm:
to me, this looks fine. I'm curious whether other systems need this as well? especially OpenBSD?

shym:
I expected so too but, at least on OpenBSD 7.3, `stdatomic.h` is in the Clang resources and I could compile the OCaml 5.2 runtime which needs this header with the current `solo5` package.

hannesm:
The reason may be that the `gen_headers.sh` does a `cd /usr/include` and `cc -M float.h stddef.h stdint.h stdbool.h stdarg.h` to figure out all dependencies of these header files -- eventually an OpenBSD system includes atomic.h/stdatomic.h in these dependencies...

hannesm:
hmm, on a OpenBSD 7.4 system I have access to, there's just no atomic.h, neither a stdatomic.h...

shym:
On an OpenBSD 7.5, I see:
```
$ echo '#include <stdatomic.h>' > test.c
$ cc -H -c test.c
. /usr/lib/clang/16/include/stdatomic.h
[...]
```

hannesm (approved):
looks fine to me, good to include into the next release

dinosaure:
Thanks!

PR #575, opened 2024-04-27
Docs: add mention of ip forwarding and NAT on host system

panglesd:
I got stuck on that for a while, would have give up without the help of Mirage folks around!

I think it is enough to just give the keywords that will allow a reader to investigate further.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -179,7 +179,7 @@ ifconfig tap100 inet 10.0.0.1 netmask 255.255.255.0
 ```
 
 In case your unikernel needs to access internet (which is not the case in this
-example), you will need to add "ip frowarding" and a NAT to your host system.
+example), you may need to add "ip forwarding" and a NAT to your host system.
 
 ## _hvt_: Running on Linux, FreeBSD and OpenBSD with hardware virtualization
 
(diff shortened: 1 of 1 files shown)


dinosaure (approved):

docs/building.md
```diff
@@ -178,6 +178,9 @@ cd /dev
 ifconfig tap100 inet 10.0.0.1 netmask 255.255.255.0
 ```
 
+In case your unikernel needs to access internet (which is not the case in this
+example), you will need to add "ip frowarding" and a NAT to your host system.
```
reynir, line 182:
```suggestion
example), you may need to add "ip forwarding" and a NAT to your host system.
```
what is annoying about networking is that a bridge may be enough if you have available IPs (e.g. if there is a router doing NAT for you or you're lucky to have ample IP addresses) - it all depends very much about your setup.

None the less I am in favor of this as it gives a hint that you may need to do some additional setup (whatever that is)

reynir (approved):

docs/building.md
panglesd: Thanks! 'frowarding" :facepalm:

dinosaure:
Thanks!

PR #576, opened 2024-05-06
Ensure reproducibility of Solo5

dinosaure:
This patch adds a new option availabe since GCC 8 and Clang 10 which delete absolute path from produced object files. It does not have huge implication on the execution of tenders and unikernels but it allows us to install Solo5 on some systems which check that installed files (like solo5_{hvt,...}.o) don't include user informations.

- [ ] check the version of the compiler used


hannesm (approved):
looks good to me

dinosaure:
Let's merge it, but we should assert now that we don't support GCC < 10 and Clang < 8.

PR #578, opened 2024-05-08
Check GET_TSC_KHZ return code correctly & update Intel SDM reference

reynir:
We check the ioctl for `KVM_GET_TSC_KHZ` being -1 only, but according to the documentation it will return `-EIO` (`-5`) when the TSC is unstable (which we check for in errno).  https://www.kernel.org/doc/html/latest/virt/kvm/api.html#kvm-get-tsc-khz

I also updated a reference to Intel SDM which now lives in chapter 24 instead of chapter 23. Maybe we should version the references somehow...

### DIFF
--- a/tenders/hvt/hvt_kvm.c
+++ b/tenders/hvt/hvt_kvm.c
@@ -75,7 +75,7 @@ struct hvt *hvt_init(size_t mem_size)
     if (hvb->vcpurun == MAP_FAILED)
         err(1, "KVM: VCPU mmap failed");
 
-    /* XXX(reynir): probably very unlikely to fail... */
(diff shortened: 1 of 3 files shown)


tenders/hvt/hvt_kvm.c
```diff
@@ -75,6 +75,12 @@ struct hvt *hvt_init(size_t mem_size)
     if (hvb->vcpurun == MAP_FAILED)
         err(1, "KVM: VCPU mmap failed");
 
+    /* XXX(reynir): probably very unlikely to fail... */
```
hannesm, line 78:
```suggestion
    /* (reynir, 2024-05-08): probably very unlikely to fail... */
```

hannesm (approved):
looks fine to me

PR #579, opened 2024-05-14
Fix tests with QEMU 9.0

wegank:
`-no-acpi` was removed in QEMU 9.0, see https://www.qemu.org/docs/master/about/removed-features.html#no-acpi-removed-in-9-0.

### DIFF
--- a/scripts/virtio-run/solo5-virtio-run.sh
+++ b/scripts/virtio-run/solo5-virtio-run.sh
@@ -146,7 +146,7 @@ kvm|qemu)
     hv_addargs -m ${MEM}
 
     # Kill all default devices provided by QEMU, we don't need them.
-    hv_addargs -nodefaults -no-acpi
+    hv_addargs -nodefaults -machine acpi=off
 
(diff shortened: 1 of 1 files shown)


reynir:
How backwards compatible is this? From the link it sounds like `-machine acpi=off` might be a new thing in QEMU 9.0.

wegank:
I don't think it is that new, I'm at least seeing `-machine acpi=off` in QEMU 6.0 (https://gitlab.com/qemu-project/qemu/-/blob/stable-6.0/softmmu/vl.c#L3311-3313).

reynir (approved):
It's even in 5.0 (released April 2020): https://gitlab.com/qemu-project/qemu/-/blob/stable-5.0/softmmu/vl.c?ref_type=heads#L3520-3523

Debian oldstable ships qemu 5.2 so it seems pretty safe \o/

hannesm (approved):
looks good to me

dinosaure:
Thanks!

Issue #580, opened 2024-09-04
Multiple virtio net devices


palainp:
Dev devs,

For experimentation purposes using mirageOS unikernels with GNS3, I need to use multiple network devices (at least 2, e.g. for a firewall unikernel). So far, using virtio target has worked well, but according to the code, the implementation is currently limited to a single net device: https://github.com/Solo5/solo5/blob/5478c255911e1dae6c9ab1f448daddf0fb46ea37/bindings/virtio/virtio_net.c#L283-L284

Maybe I'm getting it wrong with virtio and can use a different target, or maybe I can work on support for multiple net interfaces (in which case, what would be the best way to code this functionality?).

Any insight would be helpful :)
Best.

hannesm:
Dear @palainp, there has been work earlier on virtio and multiple network devices https://github.com/Solo5/solo5/pull/377

As far as I remember, the main issue was: who needs that? and the other about naming -- in hvt & spt we have this very nice naming (`solo5-hvt --net:service=tapN`) -- and we haven't found a nice way to transport names in virtio -- so how would such a mapping happen (one choice would be "first NIC passed is the first named interface"). If there's some idea how to do that, I guess we can just merge such a PR (see at the earlier PR for discussions).

palainp:
Dear @hannesm, thank you for your reply, I missed that PR. To answer the first question about "who needs that?" I'm now one of them :)
For the second question I'll check the PR and play with it a bit (probably it'll need some sort of rebase). Thanks for the pointer!

hannesm:
I agree that it is useful to have. I find the manifest in solo5 very nice, but do not know enough of virtio to understand whether / how such a mapping could take place. AFAICT from operational experience, it is just NIC_0..NIC_n, no names, and as an operator you've to remember which network interface is connected where.

palainp:
Solved by #581

PR #581, opened 2024-09-14
Update virtio

palainp:
Dear Madam or Sir,

I recently needed to use multiple network devices with virtio, and as @hannesm pointed out in #580, there was some work done by @ricarkol a while back. So this PR is an update from #377 to the current head, and tested with https://github.com/palainp/simple-fw.

After an online discussion, @reynir suggested following the naming conventions for devices in the linux ecosystem (e.g. sd[a-z] for block devices - unsure what is after sdz -, and eth[0-9] for network devices - I don't know if it's better to extend to "10" or "a" to keep a 4 letter device -). To me that seems a good solution to tackle the problem of naming and discovering the corresponding devices, currently they are assigned in the manifest order. What I'm wondering is how to deal with the mapping between manifest index and device index for blocks and networks that will be unaligned.

Best.

### DIFF
--- a/bindings/virtio/virtio_net.c
+++ b/bindings/virtio/virtio_net.c
@@ -243,8 +243,7 @@ static void virtio_net_config(struct virtio_net_desc *nd,
     outb(pci->base + VIRTIO_PCI_STATUS, VIRTIO_PCI_STATUS_DRIVER_OK);
 }
 
-/* Returns 1 if there is a pending used descriptors for us to read.
-FIXME: palainp, the description does not match the behaviour of the function */
+/* Set ready for each net interface if there is pending data to read. */
 static void virtio_net_pkt_poll(virtio_set_t *ready_set)
 {
     unsigned idx;
--- a/bindings/virtio/virtio_net.c
+++ b/bindings/virtio/virtio_net.c
@@ -28,9 +28,28 @@
 #define VIRTIO_NET_F_MAC (1 << 5) /* Host has given MAC address. */
 
 #define PKT_BUFFER_LEN 1526
+#define VIRTIO_NET_MTU 1500
 
-static struct virtq recvq;
-static struct virtq xmitq;
+extern struct mft *virtio_manifest;
+
+struct virtio_net_desc {
+    uint16_t pci_base; /* base in PCI config space */
+    struct virtq xmitq;
+    struct virtq recvq;
+    uint8_t net_mac[6];
+    uint16_t mtu;
+    solo5_handle_t handle;
+};
+
+#define VIRTIO_NET_MAX_ENTRIES  MFT_MAX_ENTRIES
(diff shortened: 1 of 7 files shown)


hannesm:
Thanks a lot for this PR. Would you mind to split the last commit into a separate PR, since it is unrelated to multiple virtio devices.

hannesm:
> I don't know if it's better to extend to "10" or "a" to keep a 4 letter device -).

I'd go for `10`.

> To me that seems a good solution to tackle the problem of naming and discovering the corresponding devices, currently they are assigned in the manifest order. What I'm wondering is how to deal with the mapping between manifest index and device index for blocks and networks that will be unaligned.

Would you mind to expand a bit on "unaligned"? I.e. what is the situation you've in mind?

I wonder: are network and block devices in a manifest grouped together?

I appreciate this PR, and it looks like you already have used it quite a bit for real. It may be worth to document how virtio & multiple devices can be used with solo5/mirage (but I fail to find a good spot for that documentation right now).

bindings/virtio/virtio_net.c
```diff
@@ -224,40 +243,48 @@ void virtio_config_network(struct pci_config_info *pci)
     outb(pci->base + VIRTIO_PCI_STATUS, VIRTIO_PCI_STATUS_DRIVER_OK);
 }
 
-/* Returns 1 if there is a pending used descriptor for us to read. */
-int virtio_net_pkt_poll(void)
+/* Returns 1 if there is a pending used descriptors for us to read.
+FIXME: palainp, the description does not match the behaviour of the function */
```
hannesm, line 247: would you mind to solve this fixme? :)

hannesm (approved):
looks fine to me!

bindings/virtio/virtio_net.c
palainp: Thank you, applied in [0f92621](https://github.com/Solo5/solo5/pull/581/commits/0f926216d4f428ce2b5e086ee9d877f462979dc6).

palainp:
> would you mind for the first commit to amend the commit message and add a "co-authored-by" line for ricardo?

Thank you @hannesm for your review, that line was already present, but I don't know why the avatar does not comes up. Even a blank line does not help :(

palainp:
> now, if you make this line: `Co-Authored-By: Ricardo Koller <ricarkol@google.com>`, GitHub will make this a nice link

Unfortunately it's not enough, last time it worked with a blank line, but not anymore :(

dinosaure:
@palainp what is the status of this PR? Should we merge it for the next release?

palainp:
@dinosaure to me this PR works fine and have {net,block}_devices names accorded to the manifest. I'd like to have it for the next release and I'll try to work on the naming convention in another PR if you're ok with that?

dinosaure:
Ok, let's merge! Thanks!

PR #582, opened 2024-09-19
spt: Fix hang in solo5_yield

reynir:
Same fix as in 97a66f6645499fafe77deadbce6d80c73641ea19

Calling solo5_yield(0, ...) results in  passing a struct itimerspec with an it_value of zero in both fields to timerfd_settime(), which disarms the timer, which causes the subsequent call to epoll_pwait() to hang forever if no other file descriptors (external events) have been registered with the waitset.

This addresses https://github.com/mirage/mirage-solo5/issues/97

Dune cache was interfering and this patch only worked once I disabled the dune cache.

### DIFF
--- a/bindings/spt/net.c
+++ b/bindings/spt/net.c
@@ -107,6 +107,12 @@ void solo5_yield(solo5_time_t deadline, solo5_handle_set_t *ready_set)
         }
     };
     /*
+     * Ensure that it.it_value is always non-zero, otherwise the following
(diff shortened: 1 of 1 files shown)


dinosaure (approved):

reynir:
I think it could be nicer both in spt and hvt to check if `deadline` is zero and then call `epoll_pwait()` with 0 as timeout instead of -1, but it requires a bit more code and I don't feel confident with that task.

dinosaure:
Thanks!

PR #583, opened 2024-10-08
hvt block: remove redundant check and assignment

hannesm:
The very same condition and assignment is still present 2 just after the comment

Discovered while looking for the default sector size with @PizieDust


dinosaure (approved):

reynir (approved):
Oops! I am to blame. Thanks!

dinosaure:
Thanks!

PR #584, opened 2024-10-09
Don't remove the first command line token

palainp:
Dear devs,
To match the code from Xen binding (see commit 5d3a214), if virtio boots via multiboot, do not remove the first token (it is actually an argument).
Best.


hannesm (approved):
looks good to me, thanks!

dinosaure:
Thanks, as for Xen, I trust on you about that. However, I do have some questions about GCM.

PR #587, opened 2024-10-10
Since #584 and #517, we must take into account a possible name of the executable

dinosaure:
Tests expect only arguments on `cmdline` which is not true anymore for Xen and Virtio according #584 & #517. This commit is a fix for tests for Xen and Virtio cases.

A double look from @palainp and @hannesm will be appreciate :+1: (and a `{sudo|doas} ./run-test.sh` too :slightly_smiling_face:).

### DIFF
--- a/tests/test_hello/test_hello.c
+++ b/tests/test_hello/test_hello.c
@@ -40,8 +40,10 @@ int solo5_app_main(const struct solo5_start_info *si)
 
     puts("'\n");
 
-    /* "Hello_Solo5" will be passed in via the command line */
-    if (strcmp(si->cmdline, "Hello_Solo5") == 0)
+    /* "Hello_Solo5" will be passed at the end of the command line
+     * NOTE: on hvt, we only have "Hello_Solo5", on Xen and Virtio, we have the
+     * name of the executable. */
+    if (strcmp(si->cmdline + len - 11, "Hello_Solo5") == 0)
         puts("SUCCESS\n");
 
     return SOLO5_EXIT_SUCCESS;
--- a/tests/test_net/test_net.c
+++ b/tests/test_net/test_net.c
@@ -374,12 +374,25 @@ static bool ping_serve(void)
     return true;
 }
 
+int is_alpha(char chr) {
+    return ((chr >= 0x61 && chr <= 0x7a) || (chr >= 0x41 && chr <= 0x5a));
+}
+
 int solo5_app_main(const struct solo5_start_info *si)
 {
     puts("\n**** Solo5 standalone test_net ****\n\n");
 
-    if (strlen(si->cmdline) >= 1) {
-        switch (si->cmdline[0]) {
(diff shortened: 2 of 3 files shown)


dinosaure:
> To me a test for that implies the creation of a disk image and qemu-ish.

Actually, that's what I run when I do `sudo ./run-test.sh`, we launch the unikernel with `qemu` and without this patch, tests fail.

palainp:
Thanks @dinosaure I'm to blame, `virtio/platform.c:parse_multiboot(.)` is used in all cases and not just with GRUB multiboot, whereas it's not the case for Xen. However, I haven't dug into the test scripts, but a manual start of text_hello works fine `on my laptop` without this PR:
```bash
$ qemu-system-x86_64 -kernel test_hello.virtio -nographic -append Hello_Solo5
[...]
Hello, World
Command line is: 'Hello_Solo5'
SUCCESS
Solo5: solo5_exit(0) called
Solo5: Halted
```
And it also works when a `disk.img` is created (i use https://github.com/palainp/simple-fw/blob/main/misc/create_img.sh) with grub as bootloader. So maybe the way tests are run for virtio adds this "opaque name"?

palainp:
And at the same time, probably there is no need for `placeholder` at https://github.com/Solo5/solo5/blob/d026c2f0fd604c8a105943f6b0260e90239ec60e/scripts/virtio-run/solo5-virtio-run.sh#L244C43-L244C54 anymore?

dinosaure:
> For the time being, maybe the easiest way to cut a release is to remove https://github.com/Solo5/solo5/pull/584 ? (and I'll use a placeholder in my image creation script :) )

Revert #584 will be easier, the question is: what we want? do we want to get the name of the binary or not?

From my perspective, if there's a valid reason for wanting the binary name on `virtio`, I can release Solo5 as is (with this patch). If it's not necessary, we can just revert and do a release. The second case is interesting because it ensures consistency between targets and argument management - but I can understand that it might be interesting to have the name of the binary.

hannesm:
I was not aware about the consequences of this patch -- and would prefer to have #584 reverted if that is fine with @palainp as well. What is the story on Xen, then? If I understand correctly, #517 is a good thing to have!?

dinosaure:
> If I understand correctly, https://github.com/Solo5/solo5/pull/517 is a good thing to have!?

Yes, it seems that the boot is different between Xen and `virtio` and that, for the former, #517 is necessary whereas it isn't for the latter.

palainp:
I wasn't aware of the consequence too /o\
For Xen the situation is ok , but for virtio qemu insert the name of the kernel before boot param when GRUB does not. So with #584 it's not possible anymore to boot using `qemu -kernel` command. I'm in favor to reverse #584 and I'll insert a placeholder name when I create a disk image.
I'm just worried that virtio is the only target that does not have only boot param as command line.
Thank you @dinosaure and @hannesm for the feedback!

dinosaure:
See #588

PR #588, opened 2024-10-11
Revert #584 & fix tests

dinosaure:


palainp:
Thank you @dinosaure , LGTM. And if qemu appears to have an option to disable the kernel name in the command line, 584 can be back :)

palainp (approved):

PR #590, opened 2025-01-24
Fetch `stdalign.h` from system on FreeBSD

shym:
This PR adds `stdalign.h` to the set of headers that must be copied over from `/usr/include` on FreeBSD because it is not included in the clang resources on that system. `stdalign.h` is used in OCaml 5.3.0 (discovered in CI for mirage/ocaml-solo5#148).


dinosaure (approved):

hannesm:
I guess for moving the ocaml-solo5 to 5.3.0, we'll need to have this merged and released.

shym:
Indeed. It could be a good opportunity to add the `x-maintenance-intent`, by the way.

dinosaure:
Thanks!

PR #593, opened 2025-03-06
add x-maintenance-intent to opam files

hannesm:
would be great if you could add these to https://github.com/ocaml/opam-repository/pull/27567 as well.


dinosaure:
Ah yes, I forget, I will add it into my PR on `opam-repository`.

PR #596, opened 2025-07-16
As unikraft, we increase the ringbuffer to have a better throughput on our network device

dinosaure:
As spotted here: https://github.com/unikraft/unikraft/blob/staging/drivers/virtio/ring/virtio_ring.c#L54. The throughput up to ~300MB/s instead of 260MB/s (/cc @shym who did some benchmarks on `hvt`).

### DIFF
--- a/bindings/virtio/virtio_ring.c
+++ b/bindings/virtio/virtio_ring.c
@@ -26,7 +26,7 @@
  * There is no official max queue size. But we've seen 4096, so let's use the
  * double of that.
  */
-#define VIRTQ_MAX_QUEUE_SIZE 8192
+#define VIRTQ_MAX_QUEUE_SIZE 32768
 
 
 /*


Firobe:
> @Firobe: are the baseline results for solo5-virtio similar to what you had measured? (And sorry if you saw the first draft of this message, I had used the wrong measures for some results)

Yes, I have similar results (included below), including the non-significant difference for this PR. Now keep in mind this specific benchmark measures a relatively narrow use case (mostly receiving data), and different tests might bring more significant improvements out. 

<details>
  <summary>My results</summary>
  
graph

</details>

reynir:
I tried to follow the unikraft code to figure out where the value `nr_desc` comes from in that function, but it was a deep rabbit hole of tracing back other functions and functions stored in structs. I don't think `VIRTQ_MAX_QUEUE_SIZE` is the size of the queue in unikraft. Possibly it's a value known to be likely greater than `nr_desc`, but then it's weird that they don't assert that.

PR #597, opened 2025-07-17
Silence stringop-overflow warning for SSP test

Firobe:
Recent versions of GCC have -Wstringop-overflow=2 enabled by default, which correctly detects a buffer overflow when we volontary smash the stack during the test. Silence it.


dinosaure:
Thanks! I got the error also with GCC. Let's merge.

Issue #598, opened 2025-07-27
The OpenBSD hvt tender is broken on -current


omegametabroccolo:
The way memory is allocated / handled has changed since OpenBSD 7.7 ([see commit](https://github.com/openbsd/src/commit/bde04a5b5e211fdc984547b13a749a6a425c4db3)). `solo5-hvt` still builds, but throws a triple fault when running any example.

I have a [branch](https://github.com/omegametabroccolo/solo5/tree/obsd-current-fix) with the appropriate fixes for -current. However, my changes in the memory handling code break released versions of OpenBSD, and I'm not sure what is the best way to support both (maybe some kind of `#ifdef` magic?).

dinosaure:
Hello, thank you for reporting this. I think the best solution is:
1) submit a PR so that Solo5 users are aware of this change
2) possibly add an `ifdef` for the `-current` version of OpenBSD

Then we can see if we can wait for an OpenBSD release to consolidate the ifdef or propose a Solo5 release that is compatible with `-current` and OpenBSD 7.7 (I would be more inclined towards the first approach since `-current` is not a release in itself). But we could discuss all this directly on the proposed PR.

omegametabroccolo:
I'll split this into two pull requests: one for fixing `test_ssp` that can be merged now, and another for the main fix.

PR #599, opened 2025-07-27
Fix test_ssp on OpenBSD

omegametabroccolo:
OpenBSD's `clang` doesn't know about `-Wstringop-overflow` on -current. This can be merged now. #598


hannesm (approved):
lgtm

Issue #605, opened 2025-10-08
MTU hardcoded to 1500 in tenders hvt, spt, muen


reynir:
I currently don't have a need for a different MTU, but now it is documented in this issue.
https://github.com/Solo5/solo5/blob/ff63477507a032583e2e2a91dc93eed5b27103cd/tenders/hvt/hvt_module_net.c#L126
https://github.com/Solo5/solo5/blob/ff63477507a032583e2e2a91dc93eed5b27103cd/tenders/spt/spt_module_net.c#L78
https://github.com/Solo5/solo5/blob/ff63477507a032583e2e2a91dc93eed5b27103cd/bindings/muen/muen-net.c#L219

ananos:
awesome, thanks for that -- some k8s deployments have bitten us hard with this! It's on our TODO list to fix it.

PR #606, opened 2025-10-08
Implement MTU detection on Linux for hvt, spt

reynir:
For the other tenders and other OSes it is still hardcoded to 1500.

This is an attempt at addressing #605.

For FreeBSD and OpenBSD I have not investigated how to detect the configured MTU.

There is no way to inform the unikernel if the MTU is adjusted on the host.

### DIFF
--- a/tests/tests.bats
+++ b/tests/tests.bats
@@ -84,6 +84,7 @@ setup() {
   NET0_IP=10.0.0.2
   NET1=tap101
   NET1_IP=10.1.0.2
+  NET2=tap102
 }
 
 teardown() {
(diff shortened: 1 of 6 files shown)


hannesm:
I like your patch, and pushed https://github.com/hannesm/solo5/commit/54ebd8e54260a9e36bee1afaf69a9102e7ef6036 which makes it working on FreeBSD as well (and I'm also very certain it will work on OpenBSD) :D For some reason unknown to me, I can't open a PR on your solo5 repository with that commit.

tests/tests.bats
```diff
@@ -644,6 +645,20 @@ xen_expect_abort() {
   expect_success
 }
 
+@test "net_mtu hvt" {
```
hannesm, line 648: the test net_mtu, is not present... did you forgot to git add it?

reynir: Yes, I forgot to add them. They are in https://github.com/Solo5/solo5/pull/606/commits/4064e8f432f1c1fda30cbe2f1db1e03390ca3d31

reynir:
I think this is ready for review. For Muen and Virtio I don't know what we can do.

Kensan:
> I think this is ready for review. For Muen and Virtio I don't know what we can do.

For Muen, this would need to be some kind of (static) configuration parameter, maybe passed to the unikernel via command-line or by extension of the manifest or some other mechnism.

PR #607, opened 2025-10-08
test_ssp: -Wstringop-overflow is only available for GNU C, not for clang

hannesm:
Otherwise, the build errors out due to -Wunknown-warning and -Werror


omegametabroccolo:
Since this has to do with clang we can strip the OpenBSD condition.

hannesm:
CI shows that this fixes the FreeBSD runner, merging.

Issue #604, opened 2025-10-01
Error "too many entries, maximum 64" with 64 devices


reynir:
With a manifest.json with exactly 64 devices I get an error saying there are too many entries and the maximum is 64:
```
solo5-elftool: manifest.json:4: .devices[]: too many entries, maximum 64
```

I suspect this is due to the implicit `MFT_FIRST_RESERVED` being counted towards the 64 max entries, but it's confusing to the user who may not be aware of this implementation detail. I guess we can modify the error message?!

<details>
<summary>

Contents of `manifest.json`

</summary>

```json
{
  "type": "solo5.manifest",
  "version": 1,
  "devices": [
      { "name": "1", "type": "NET_BASIC" },
      { "name": "2", "type": "NET_BASIC" },
      { "name": "3", "type": "NET_BASIC" },
      { "name": "4", "type": "NET_BASIC" },
      { "name": "5", "type": "NET_BASIC" },
      { "name": "6", "type": "NET_BASIC" },
      { "name": "7", "type": "NET_BASIC" },
      { "name": "8", "type": "NET_BASIC" },
      { "name": "9", "type": "NET_BASIC" },
      { "name": "10", "type": "NET_BASIC" },
      { "name": "11", "type": "NET_BASIC" },
      { "name": "12", "type": "NET_BASIC" },
      { "name": "13", "type": "NET_BASIC" },
      { "name": "14", "type": "NET_BASIC" },
      { "name": "15", "type": "NET_BASIC" },
      { "name": "16", "type": "NET_BASIC" },
      { "name": "17", "type": "NET_BASIC" },
      { "name": "18", "type": "NET_BASIC" },
      { "name": "19", "type": "NET_BASIC" },
      { "name": "20", "type": "NET_BASIC" },
      { "name": "21", "type": "NET_BASIC" },
      { "name": "22", "type": "NET_BASIC" },
      { "name": "23", "type": "NET_BASIC" },
      { "name": "24", "type": "NET_BASIC" },
      { "name": "25", "type": "NET_BASIC" },
      { "name": "26", "type": "NET_BASIC" },
      { "name": "27", "type": "NET_BASIC" },
      { "name": "28", "type": "NET_BASIC" },
      { "name": "29", "type": "NET_BASIC" },
      { "name": "30", "type": "NET_BASIC" },
      { "name": "31", "type": "NET_BASIC" },
      { "name": "32", "type": "NET_BASIC" },
      { "name": "33", "type": "NET_BASIC" },
[...]

PR #609, opened 2025-10-09
Elftool: report max entries w/o RESERVED_FIRST

reynir:
From the user's point of view it is confusing to have 64 entries in the .devices json field and get an error message saying the maximum number of entries is 64. The user should not need to know about the implicit MFT_RESERVED_FIRST entry.

This addresses #604.

I added a test, but unfortunately it doesn't work. The output is as expected, but the test fails for some reason. I don't know enough about bats to figure out why it complains.

### DIFF
--- a/tests/test_too_many_entries/GNUmakefile
+++ b/tests/test_too_many_entries/GNUmakefile
@@ -16,11 +16,8 @@
 # NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 # CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-include $(TOPDIR)/Makefile.common
-
 test_NAME := test_too_many_entries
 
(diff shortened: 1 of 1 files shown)


reynir:
I think I marked it WIP due to the test not working. The test prints the expected output, but bats is not happy. So I can remove the test again if we don't figure out how to make bats happy (help appreciated).

hannesm:
thanks, I'm happy to try this later.

hannesm (approved):

hannesm:
Interestingly, this is the only virtio test that is being built on the arm64 runner. And from the output, virtio is not enabled:
```
2025-10-23T09:52:17.7800927Z configure.sh: Enabled bindings: stub hvt spt.
2025-10-23T09:52:17.7801237Z configure.sh: Enabled tenders: hvt spt.
```

dinosaure (approved):

Issue #612, opened 2025-10-15
solo5-hvt ELF parser: assert can be trigerred by malformed ELF files


edwintorok:
`afl` found this one:
```
solo5-hvt: common/elf.c:479: elf_load_note: Assertion `note_size != 0 && note_size <= nhdr.h.n_descsz' failed.
[1]    70694 IOT instruction (core dumped)  solo5-hvt
```

This crashes before any code is executed, so is safe in that sense, but when `solo-hvt` is running potentially untrusted unikernels the coredumps could need special handling (e.g. `solo5-hvt` might need to be run with a coredump limit of 0).

I suggest replacing these assertions that are user triggerable with error handling that doesn't leave a coredump behind.
assertions should only be used for internal programming errors.

If needed I can share the corrupt ELF file:
`/var/home/edwin/git/solo5/out/default/crashes/id:000000,sig:06,src:000002,time:52084,execs:9138,op:int8,pos:24580,val:+1`

PR #613, opened 2025-10-16
Improve error handling

shym:
This PR replaces a couple of `assert`s with a proper, or at least better, handling of the corresponding error. In particular, it replaces `assert`s that check conditions that can, or maybe could, be controlled by the unikernel, either in its file format (hardening the ELF parser) or in the hypercalls it triggers. It also takes the opportunity to report (unspecified) error to the unikernels when an I/O operation failed on the host.

This is addressing #612, hopefully.

### DIFF
--- a/tenders/hvt/hvt_module_blk.c
+++ b/tenders/hvt/hvt_module_blk.c
@@ -26,8 +26,10 @@
 #define _FILE_OFFSET_BITS 64
 #include <assert.h>
 #include <err.h>
+#include <errno.h>
 #include <limits.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
@@ -69,7 +71,14 @@ static void hypercall_block_write(struct hvt *hvt, hvt_gpa_t gpa)
 
     ret = pwrite(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, wr->data, wr->len),
             wr->len, pos);
-    assert(ret == wr->len);
+    if (ret != wr->len) {
+        if (ret == -1)
+            fprintf(stderr, "Fatal error when writing: %s\n", strerror(errno));
+        else
+            fprintf(stderr, "Fatal error: wrote only %ld out of %ld bytes\n",
+                ret, wr->len);
+        exit(1);
+    }
     wr->ret = SOLO5_R_OK;
 }
 
@@ -100,7 +109,14 @@ static void hypercall_block_read(struct hvt *hvt, hvt_gpa_t gpa)
 
     ret = pread(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, rd->data, rd->len), rd->len,
             pos);
-    assert(ret == rd->len);
+    if (ret != rd->len) {
+        if (ret == -1)
+            fprintf(stderr, "Fatal error when reading: %s\n", strerror(errno));
+        else
+            fprintf(stderr, "Fatal error: read only %ld out of %ld bytes\n",
+                ret, rd->len);
+        exit(1);
+    }
     rd->ret = SOLO5_R_OK;
 }
 
--- a/tenders/hvt/hvt_module_net.c
+++ b/tenders/hvt/hvt_module_net.c
@@ -28,6 +28,7 @@
 #include <inttypes.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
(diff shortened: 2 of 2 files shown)


tenders/common/mft.c
```diff
@@ -142,6 +142,9 @@ const char *mft_type_to_string(mft_type_t type)
         case MFT_RESERVED_FIRST:
             return "RESERVED_FIRST";
         default:
-            assert(false);
+            /* This is used only to pretty print the content of the manifest so
+               a generic error is probably good enough, if only partially
+               informative */
+            return "<undefined type>";
```
reynir: Would it make sense to "print" the integer value of `type` in the string? So the developer of a new device / module has a better chance to figure out what went wrong. Then again if someone is writing a module they have a better chance of knowing to touch this part of the internals...

tenders/hvt/hvt_module_blk.c
```diff
@@ -56,8 +56,7 @@ static void hypercall_block_write(struct hvt *hvt, hvt_gpa_t gpa)
     ssize_t ret;
     off_t pos, end;
 
-    assert(wr->len <= SSIZE_MAX);
-    if (wr->offset >= e->u.block_basic.capacity) {
+    if (wr->len > SSIZE_MAX || wr->offset >= e->u.block_basic.capacity) {
```
reynir: Good catch!

tenders/hvt/hvt_module_blk.c
```diff
@@ -70,7 +69,11 @@ static void hypercall_block_write(struct hvt *hvt, hvt_gpa_t gpa)
 
     ret = pwrite(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, wr->data, wr->len),
             wr->len, pos);
-    assert(ret == wr->len);
+    if (ret != wr->len) {
+        /* Report that it failed somehow */
+        wr->ret = SOLO5_R_EUNSPEC;
+        return;
+    }
```
reynir: This changes the semantics so the unikernel can continue if the write fails. Which is perhaps okay. It's hard for the unikernel to act on it, though(?!)

From what I can read from `man -s2 pwrite write` it can return less if there is not enough disk space. Considering the file is checked to have a fixed size from the beginning I can only imagine this happening if the file is a) sparse, or b) has been truncated(!)

tenders/hvt/hvt_module_blk.c
```diff
@@ -102,7 +104,11 @@ static void hypercall_block_read(struct hvt *hvt, hvt_gpa_t gpa)
 
     ret = pread(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, rd->data, rd->len), rd->len,
             pos);
-    assert(ret == rd->len);
+    if (ret != rd->len) {
+        /* Report that it failed somehow */
+        rd->ret = SOLO5_R_EUNSPEC;
+        return;
+    }
```
reynir: I think this can only happen if the file was truncated(?)

tenders/hvt/hvt_module_blk.c
hannesm: It is not entirely clear to me whether it is worth to handle this situation gracefully. What is wrong about assertion failure and the unikernel exiting?

Handling this properly from the unikernel side may require some further information, i.e. transport the data from errno from the tender to the unikernel, and then deciding what to do in the unikernel.

tenders/hvt/hvt_module_blk.c
```diff
@@ -69,7 +71,14 @@ static void hypercall_block_write(struct hvt *hvt, hvt_gpa_t gpa)
 
     ret = pwrite(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, wr->data, wr->len),
             wr->len, pos);
-    assert(ret == wr->len);
+    if (ret != wr->len) {
+        if (ret == -1)
+            fprintf(stderr, "Fatal: error %d when writing\n", errno);
```
hannesm: Would a "%s" and "strerror(errno)" be suitable? So we don't need to decode the errno value manually.

tenders/hvt/hvt_module_blk.c
hannesm: Thanks, I only put a minor comment to use `strerror()`

edwintorok:
> Would you mind to rebase and squash, so the CI has a chance to run this PR? Also, is there a very small ELF binary that you could add for testing?

It is ~33KiB (attaching Zipped because github doesn't allow me to attach the binary directly)
[bug.zip](https://github.com/user-attachments/files/23095131/bug.zip)

tenders/hvt/hvt_module_net.c
```diff
@@ -72,15 +80,22 @@ static void hypercall_net_read(struct hvt *hvt, hvt_gpa_t gpa)
...
+    ssize_t ret;
 
     ret = read(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, rd->data, rd->len), rd->len);
     if ((ret == 0) ||
         (ret == -1 && errno == EAGAIN)) {
         rd->ret = SOLO5_R_AGAIN;
         return;
     }
-    assert(ret > 0);
+    if (ret <= 0) {
```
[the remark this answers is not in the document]
hannesm: let's keep that for a separate PR.

PR #618, opened 2025-10-23
hvt_module_net: do not return EAGAIN on a read of 0 bytes

### LINKED ISSUE #559: Solo5-hvt reads on network devices returns `SOLO5_R_AGAIN` on EOF
When a hypercall-initiated read on a network device returns 0 (EOF) the tender returns `SOLO5_R_AGAIN`. In mirage-net-solo5 this results in `listen` retrying the read. If the file descriptor is really closed this will end in an infinite loop.

https://github.com/Solo5/solo5/blob/bf29b8af11feec9dbc2e74ccdc69bc1e0cbf5dfe/tenders/hvt/hvt_module_net.c#L77-L82

The spt target does **not** do anything special in the case of `read(2)` returns 0 as far as I can tell.

It is unclear to me if a file descriptor for a tap device could ever return EOF. I observe this by abusing the file descriptor mechanism for passing network devices in https://github.com/reynir/mirage-cat

hannesm:
fixes #559 //cc @reynir

### DIFF
--- a/tenders/hvt/hvt_module_net.c
+++ b/tenders/hvt/hvt_module_net.c
@@ -75,8 +75,7 @@ static void hypercall_net_read(struct hvt *hvt, hvt_gpa_t gpa)
     int ret;
 
     ret = read(e->b.hostfd, HVT_CHECKED_GPA_P(hvt, rd->data, rd->len), rd->len);
-    if ((ret == 0) ||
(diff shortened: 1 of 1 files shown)


hannesm:
Let's close this, since there isn't too much interest in modifying the semantics.

PR #619, opened 2025-10-23
Define a `.note.GNU-stack` in assembler files when needed

shym:
Follow what the C compiler does and add the same `.note.GNU-stack` section in assembler files as the C compiler.

This silences the warnings such as:

```
ld: warning: cpu_vectors_x86_64.o: missing .note.GNU-stack section implies executable stack
```

I checked that this indeed makes unikernel builds silent in that [run](https://github.com/shym/ocaml-solo5/actions/runs/18754758959/job/53503673443#step:6:17) of OCaml/Solo5 CI.

### DIFF
--- a/bindings/GNUmakefile
+++ b/bindings/GNUmakefile
@@ -76,7 +76,7 @@ endif
 %.o: %.c %.d
 	$(COMPILE.c)
 
-%.o: %.S %.d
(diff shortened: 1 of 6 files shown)


hannesm:
If this attracts @Kensan @palainp @dinosaure, we're all set :)

Issue #396, opened 2019-09-09
Better error reporting in the ELF loader


mato:
The ELF loader has recently gained a lot of new functionality, and is also much more strict in what executables it will accept (#393, #386, #385). However, its error reporting is somewhat lacking. We should improve this at some point.

PR #624, opened 2025-12-11
Add better error reporting in the ELF loader

m-wkr:
Added a few extra goto error messages relating to elf headers, program headers, and note headers have been added in relation to #396.

### DIFF
--- a/tenders/common/elf.c
+++ b/tenders/common/elf.c
@@ -219,8 +219,8 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
         goto mem_cleanup;
     }
     if (nbytes != sizeof(Elf64_Ehdr)) {
-        warnx("%s: INV_EXE: elf header does not match the expected size"
-              " (%zu != %lu)", bin_name, nbytes, sizeof(Elf64_Ehdr));
+        warnx("%s: %s: elf header does not match the expected size"
+              " (%zu != %lu)", bin_name, INV_EXE, nbytes, sizeof(Elf64_Ehdr));
         goto mem_cleanup;
     }
     if (!ehdr_is_valid(ehdr)) {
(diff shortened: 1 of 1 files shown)


tenders/common/elf.c
```diff
@@ -211,14 +211,14 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
     if (nbytes < 0)
         goto out_error;
     if (nbytes != sizeof(Elf64_Ehdr))
-        goto out_invalid;
+        goto out_ehdr_size_disparity; 
     if (!ehdr_is_valid(ehdr))
-        goto out_invalid;
+        goto out_invalid;
```
hannesm: style: please avoid trailing whitespaces, here and in other lines. thanks.

reynir (changes requested):
Thanks!

I think this adds a lot of gotos (and labels!) which maybe goes a bit off from the current style. Since most of the labels are `warnx("..."); goto mem_cleanup` I wonder if it could be rewritten so the source locations of the gotos instead call `warnx("...")` and then `goto mem_cleanup`? Then we have fewer locations we `goto`.

tenders/common/elf.c
```diff
@@ -205,31 +207,47 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
...
-    if (nbytes != sizeof(Elf64_Ehdr))
-        goto out_invalid;
-    if (!ehdr_is_valid(ehdr))
-        goto out_invalid;
+    if (nbytes < 0) {
+        warnx("%s: pread_in_full for elf header while loading elf returned %zu < 0", bin_name, nbytes);
+        goto mem_cleanup;
+    }
+    if (nbytes != sizeof(Elf64_Ehdr)) {
+        warnx("%s: INV_EXE: elf header does not match the expected size (%zu != %lu)", bin_name, nbytes, sizeof(Elf64_Ehdr));
```
dinosaure: I actually prefer `"%s: "INV_EXE": ..."` to show that `INV_EXE` is actually a define (and not a part of the format string.

tenders/common/elf.c
```diff
@@ -210,15 +210,21 @@ void elf_load(int bin_fd, const char *bin_name, uint8_t *mem, size_t mem_size,
...
+        warnx("%s: invalid or unsupported executable", bin_name);
+        goto mem_cleanup;
+    }
     /*
      * e_entry must be non-zero and within range of our memory allocation.
      */
-    if (ehdr->e_entry < p_min_loadaddr || ehdr->e_entry >= mem_size)
-        goto out_invalid;
+    if (ehdr->e_entry < p_min_loadaddr || ehdr->e_entry >= mem_size) {
+        warnx("%s: e_entry does not fall within the memory allocation range", bin_name);
```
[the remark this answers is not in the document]
hannesm: see [6b4f536](https://github.com/Solo5/solo5/pull/624/commits/6b4f536c62435e4f3a620f7069b4ea8110f9abef)

PR #632, opened 2026-03-07
Add -Wshadow -Wpointer-arith -Wcast-qual -Wsign-compare

dinosaure:
This patch mainly try to lint our codebase with few warnings. It introduces also the usage of _Generic (C11) to be able to keep const qualifier between arguments and returned values (à la Rust).

### DIFF
--- a/bindings/virtio/start.c
+++ b/bindings/virtio/start.c
@@ -54,14 +54,14 @@ extern const struct mft1_note __solo5_mft1_note;
  * Will be initialised at start-up, and used by bindings to access (and
  * modify!) the in-built manifest.
  */
-struct mft *virtio_manifest = NULL;
+const struct mft *virtio_manifest = NULL;
 
 /* Will abort the program if any of the devices is not acquired. */
 static void mft_check_all_acquired()
 {
     bool fail = false;
     for (unsigned i = 0; i != virtio_manifest->entries; i++) {
-        if (!virtio_manifest->e[i].attached) {
+        if (i > 0 && !virtio_manifest->e[i].attached) {
             log(WARN, "Solo5: Device '%s' of type %s not attached.",
                     virtio_manifest->e[i].name, mft_type_to_string(virtio_manifest->e[i].type));
             fail = true;
@@ -89,9 +89,9 @@ static void _start2(void *arg __attribute__((unused)))
      * Get the built-in manifest out of the ELF NOTE and validate it.
      * Once validated, it is available for access globally by the bindings.
      */
-    struct mft *mft;
+    const struct mft *mft;
     size_t mft_size;
-    mft_get_builtin_mft1_unconst(&__solo5_mft1_note, &mft, &mft_size);
+    mft_get_builtin_mft1(&__solo5_mft1_note, &mft, &mft_size);
     if (mft_validate(mft, mft_size) != 0) {
         log(ERROR, "Solo5: Built-in manifest validation failed. Aborting.\n");
         solo5_abort();
@@ -102,12 +102,11 @@ static void _start2(void *arg __attribute__((unused)))
      * Attach the first entry if its type is really MFT_RESERVED_FIRST.
      * Abort otherwise.
      */
-    struct mft_entry *e = mft_get_by_index(virtio_manifest, 0, MFT_RESERVED_FIRST);
+    const struct mft_entry *e = mft_get_by_index(virtio_manifest, 0, MFT_RESERVED_FIRST);
     if (e == NULL) {
         log(ERROR, "Solo5: first entry in manifest is not of type MFT_RESERVED_FIRST\n");
         solo5_abort();
     }
-    e->attached = true;
 
     mem_init();
     time_init();
--- a/Makefile.common
+++ b/Makefile.common
@@ -30,7 +30,7 @@ include $(TOPDIR)/Makeconf
 # target.
 #
 CC := __V=$(V) $(TOPDIR)/toolchain/bin/$(CONFIG_TARGET_TRIPLE)-cc
-CFLAGS := -std=c11 -Wall -Wextra -Werror -O2 -g
+CFLAGS := -std=c11 -Wall -Wshadow -Wpointer-arith -Wcast-qual -Wsign-compare -Wwrite-strings -Wextra -Werror -O2 -g
 CPPFLAGS :=
 LD := __V=$(V) $(TOPDIR)/toolchain/bin/$(CONFIG_TARGET_TRIPLE)-ld
 LDFLAGS :=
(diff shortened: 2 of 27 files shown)


dinosaure:
I think it's ready to merge. The change is not really deep, just add few `const` and the usage of `_Generic` (C11) to be able to keep `const` between arguments and returned values. We are 2026, I think it's ok to move on. Probably @MisterDA will be interesting also by this PR.

hannesm:
I struggle with the gcc specific enabling and disabling of specific warnings, can we instead fix the code and have gcc not warn?

dinosaure:
> I struggle with the gcc specific enabling and disabling of specific warnings, can we instead fix the code and have gcc not warn?

Unfortunately, these `pragma` is related to a _not safe_ but common pattern in C about `const`.
- [here](https://github.com/Solo5/solo5/pull/632/changes#diff-6cdd533e0dc01d6abd04cc878dd8cb5170ca796af7330d1490a182928f0d49cbR280-R284) the comment above explain why we should be able to "unconst" _safely_
- [here](https://github.com/Solo5/solo5/pull/632/changes#diff-ae2d2d38b55277685491de96720e0b508cd549df364977430de514e880dc9cf6R86-R99), the name of the function is explicitly `unconst`
- and about `mft_get_by_{name,index}`, it's the same story because on one side, we would like to write data and, on the other side, we should only have a read access. This is why we have the `_Generic` which ensures that if we gives a const, we should get a const but if we give a value (which can be written), we also return a value which can be writtern

May be my experience with C is not enough to find a "good" example, but as far as I can say, the pattern to "unconst" _unsafely_ (according to `Wcast-qual`) returned values seems common in C.

dinosaure:
Probably @haesbaert or @Kensan have more knowledge (and a good opinion) about `const` in C :+1:

dinosaure:
I finally found something which seems good enough (at least, for my perspective). I deleted `#pragma` and use `_Generic` to provide a `const` function of `mft_get_by_{name,index}` & `mft_get_builtin_mft1`. I followed the same style with the `_unconst` function and use the `_Generic` to dispatch correctly the caller to the right implementation (and, I suspect that the compiler is able to optimize at some points).

A fix is needed for the virtio backend where we would like to set `virtio_manifest` to where we have our manifest but as a `const` value. `virtio` has an empty entry and we don't verify if this one was correctly attached or not.

The only remaining `pragma` is also about `virtio` when we would like to write a page into a block device but I think this one is well documented to say: it's safe to ignore `-Wcast-qual` in that case.

MisterDA:
Your MR reminds me of [N2973 Qualifier-preserving standard library functions](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2973.pdf). You may find something that helps in the "Implementation" subsection. I've also found <https://inbox.sourceware.org/libc-alpha/mvm5x8wpsmi.fsf@suse.de/T/>.
I also suggest the [`-Wwrite-strings`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wwrite-strings) C compiler flag.

dinosaure:
Thanks @MisterDA, I was able to remove few lines and propose something more clean than before (without repetition). I also applied `-Wwrite-strings`. I think, regardless documentation, this PR is good for me.

dinosaure:
A gently ping to @MisterDA to have a formal review (and if @Kensan or @haesbaert would like to participate).

MisterDA:
Just a note, in my experience `-Wcast-qual` doesn't play well with the OCaml or POSIX API. Sometimes you need to cast away a `const` or a `volatile`, that makes the compiler complains. If it works here that's nice but I wouldn't push to hard for this warning.

dinosaure:
For clarity, I rebased this PR to one commit, the diff is really small. I think it's ok to merge. `-Wcast-qual` is a bit restrictive and ask us to use `_Generic` (C11) but it leads us to write good C code. I will merge it when the CI is green.

MisterDA:
C++ has [`const_cast`](https://en.cppreference.com/w/cpp/language/const_cast.html) which could come in handy if there was an equivalent in C, when `-Wcast-qual` is enabled. I think it can be emulated with modern GCC/Clang with:

```c
/* const_cast(ptr) can only be used inside functions, on pointers */

#if (defined(__GNUC__) && __GNUC__ >= 14) || \
    (defined(__clang_major__) && __clang_major__ >= 19)
#define const_cast(ptr)                               \
 ({                                                   \
    _Pragma("GCC diagnostic push")                    \
    _Pragma("GCC diagnostic ignored \"-Wcast-qual\"") \
    (__typeof_unqual__(*(ptr)) *) (ptr);              \
    _Pragma("GCC diagnostic pop")                     \
  })
#else
#define const_cast(ptr) (ptr)
#endif

int main(void) {
    const int * const p0 = (void *)0;
    int *p1 = const_cast(p0);
    return 0;
}
```

dinosaure:
I think it's ok to merge now.

bindings/virtio/start.c
```diff
@@ -54,14 +54,14 @@ extern const struct mft1_note __solo5_mft1_note;
...
-struct mft *virtio_manifest = NULL;
+const struct mft *virtio_manifest = NULL;
 
 /* Will abort the program if any of the devices is not acquired. */
 static void mft_check_all_acquired()
 {
     bool fail = false;
     for (unsigned i = 0; i != virtio_manifest->entries; i++) {
-        if (!virtio_manifest->e[i].attached) {
+        if (i > 0 && !virtio_manifest->e[i].attached) {
```
hannesm, line 64: I don't understand this change as part of this PR...

bindings/virtio/start.c
```diff
@@ -102,12 +102,11 @@ static void _start2(void *arg __attribute__((unused)))
      * Attach the first entry if its type is really MFT_RESERVED_FIRST.
      * Abort otherwise.
      */
-    struct mft_entry *e = mft_get_by_index(virtio_manifest, 0, MFT_RESERVED_FIRST);
+    const struct mft_entry *e = mft_get_by_index(virtio_manifest, 0, MFT_RESERVED_FIRST);
     if (e == NULL) {
         log(ERROR, "Solo5: first entry in manifest is not of type MFT_RESERVED_FIRST\n");
         solo5_abort();
     }
-    e->attached = true;
```
hannesm, line 110: I also don't understand this change.

dinosaure: To be precise, the manifest is currently annotated with `const`, which prevents us from modifying the first entry `e`. This entry is a special entry in the case of virtio that does not correspond to a network interface or a block device.

The idea prior to this commit was to verify that the first entry in our manifest was indeed a special entry and then treat it as attached if the test passed.

Next, the `mft_check_all_acquired` function checks whether all entries (including our first special entry) have been successfully attached.

Since the `const` prevents us from modifying our first entry, and since this entry does not correspond to any device and ultimately serves only to verify that the manifest is correct, I have removed the `e->attached = true` (due to the `const`) and we do not check whether the first entry is attached or not (hence the `i > 0 &&`). The test for the first entry is still present, and if it fails, we have a `solo5_abort()`.

I prefer this approach, which better highlights the fact that the first entry is a bit special. I can add a comment to explain why we don’t check `virtio_manifest->e[0]` if you wish.

hannesm: thanks for your explanation.

bindings/virtio/start.c
hannesm: wouldn't the same be to start from i = 1 in the for loop?

A comment why i=0 is special would be very welcome.

PR #633, opened 2026-03-09
Use `clang-format`

dinosaure:
On top of #632, this PR introduce a new GitHub action which verifies that we respect a certain way to write our code. I also applied `clang-format` on the full codebase which generates a huge diff. I really tried to limit the diff and tweak a bit `clang-format`. We have two rules: `format` which calls `clang-format -i` and `format-check` which is used by our CI. I also completed the `clean` rules into the `bindings/` directory to delete `noexecstack.{c,h,s}`.


hannesm:
I'm not against a formatter, but suggest (a) to add it to `.git-blame-ignore-revs`, and (b) run it in CI as something that pushes to a PR.

I appreciate the work and the CI job, but I find it much nicer to not bother a contributor to act on some formatting issues. I understand that e.g. dnsvizor/mollymawk have such a CI job which unfortunately at the moment only works for branches of the original repository (not on forks). Maybe this is already the case here?

dinosaure:
I’ve added the `.git-blame-ignore-revs` file. As for CI, the first thing is that I haven’t found a satisfactory CI solution that allows me to create a commit applying `clang-format`, and I’m not keen on using a GitHub Action that isn’t really well maintained.

So I’ve created a new PR that adds `docs/contributing.md` (see #634), which explains how to apply `clang-format` and that contributors should add the relevant commit to the `.git-blame-ignore-revs` file.

I tried to keep the diff as minimal as possible, but not only were quite a few files imported, and the question of style is often very subjective. So I tried to achieve a configuration that was roughly consistent with what I could see in the code.

hannesm:
cool. the ci check action is already a great thing so we don't end up with unformatted code.

If you're happy with this, please feel free to merge.

Issue #635, opened 2026-03-31
We never use `HVT_TENDER_ENABLE_CAPSICUM`


dinosaure:
When I looked into the FreeBSD support of `hvt`, I saw some code about `CAPSICUM` but it seems that we have typo when we would like to enable capsicum, we define `DHVT_FREEBSD_ENABLE_CAPSICUM` (and not `HVT_FREEBSD_ENABLE_CAPSICUM`)
https://github.com/Solo5/solo5/blob/a333cbb20152ecfad9bfd4957aa7f8b237408600/tenders/GNUmakefile#L73

So, as far as I understand, everything related to capsicum is just a dead code (and we probably never tested it...). Should we try to re-integrate it (I suspect the code does not work also) or just remove it?

hannesm:
Good catch. I tested this on my laptop, and the tests (for hvt and virtio) run fine. I'll submit a PR fixing the typo in a second.

hannesm:
see #636

PR #637, opened 2026-03-31
hvt: optimize net interface

dinosaure:
Here is an initial attempt to optimise our unikernels with regard to hvt. To fully understand the original problem: every time we want to write an Ethernet frame, our execution path is as follows:
1) we prepare a value
2) we write using `out`; this is a VM-exit
3) we are in the tender’s address space and execute `read()`
3) we prepare the result
4) we return to the VM

What is **very** costly here is the VM-exit (`out`), which significantly degrades our performance if we run a ‘fair’ benchmark comparison with VirtIO.

## Overview

The idea behind this PR is to offer something very similar to what VirtIO can provide with Virtqueues. It therefore involves implementing queues that are shared between the host and the guest so that they can exchange information.
In this case, the guest sends “entries”, which are actions the host must perform (read from and write to a network interface), and the host sends confirmation of these actions along with the result. In our case, we are only interested in the result of the read operation. Initially, Solo5 would only return SOLO5_R_OK or fail when writing:

https://github.com/Solo5/solo5/blob/a333cbb20152ecfad9bfd4957aa7f8b237408600/tenders/hvt/hvt_module_net.c#L46-L70

To enable the guest to transmit information to the host, we must be mindful of certain barriers (as was the case with VirtIO, #630) and there may be instances where we wish to “kick” our host’s thread.

Indeed, our host may be in a state of waiting for entries (the unikernel has not sent any entries). In this case, the thread goes to sleep whilst reading a file descriptor. As regards KVM, this file descriptor is derived from ioeventfd (it is a file descriptor that can be associated with a specific memory address). As regards OpenBSD and FreeBSD, this file descriptor is one derived from pipe() (thanks to @haesbaert for giving me the tip already present in Miou).

If our thread is waiting, it updates a shared variable called needs_kick. If the unikernel sees this value set to 1, it performs a VM-exit to wake up the host thread. On KVM, this VM-exit is inexpensive as KVM handles it directly. On OpenBSD/FreeBSD, there will be a proper VM-exit which involves writing to the other end of our pipe().

As for the unikernel, since we have very little available in this space, I chose to statically allocate what is necessary for our “ring”. In this case, when a unikernel wishes to write, we need to “hold” its buffer. This buffer will only become available after our host thread has written it to the interface. Once again, we’re not interested in write confirmation. So we’ll simply attempt to write until the queue is full. The buffers are allocated via the .bss segment.

The ring is a bit special, however. It must exist for the guest but also for the host. The idea is to reserve an area just before the stack. This is why there are now guest_mem_size and mem_alloc_size. The first refers to what the unikernel can use and where the stack must start. The second refers to everything allocated for the guest (which includes what is needed for our ring). Thus, a unikernel requesting 512MB will only have 510MB (as the ring takes up 2MB). The ring is only allocated if a network interface is present.

Since the buffers are statically allocated, there is a scenario where the user might wish to configure the MTU to more than 2048 bytes (the maximum size of a buffer that can be written). In this case, we ‘fall back’ to the hypercall scenario and therefore perform a VM-exit. Furthermore, increasing the MTU will therefore tend to degrade performance.

## Benchmark

It should be noted first of all that the performance gain is specifically observable on Linux. I haven’t run any benchmarks on FreeBSD/OpenBSD, but I think there must still be a gain. Here is a benchmark using IPerf3 without that pull request:
```
Connecting to host 10.0.0.2, port 5201
[  6] local 10.0.0.1 port 36664 connected to 10.0.0.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  6]   0.00-1.00   sec  80.1 MBytes   671 Mbits/sec    0    178 KBytes       
[  6]   1.00-2.00   sec  86.0 MBytes   721 Mbits/sec    0    178 KBytes       
[  6]   2.00-3.00   sec  85.0 MBytes   713 Mbits/sec    0    178 KBytes       
[  6]   3.00-4.00   sec  82.2 MBytes   690 Mbits/sec    0    178 KBytes       
[  6]   4.00-5.00   sec  83.5 MBytes   700 Mbits/sec    0    178 KBytes       
[  6]   5.00-6.00   sec  84.9 MBytes   712 Mbits/sec    0    178 KBytes       
[  6]   6.00-7.00   sec  85.2 MBytes   715 Mbits/sec    0    178 KBytes       
[  6]   7.00-8.00   sec  80.5 MBytes   675 Mbits/sec    0    178 KBytes       
[  6]   8.00-9.00   sec  83.1 MBytes   697 Mbits/sec    0    178 KBytes       
[  6]   9.00-10.00  sec  85.6 MBytes   718 Mbits/sec    0    178 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  6]   0.00-10.00  sec   836 MBytes   701 Mbits/sec    0            sender
[  6]   0.00-10.00  sec  0.00 Bytes  0.00 bits/sec                  receiver

iperf Done.
```

And here is the result with this PR:
```
Connecting to host 10.0.0.2, port 5201
[  6] local 10.0.0.1 port 46470 connected to 10.0.0.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  6]   0.00-1.00   sec   254 MBytes  2.13 Gbits/sec    0    178 KBytes       
[  6]   1.00-2.00   sec   261 MBytes  2.19 Gbits/sec    0    178 KBytes       
[  6]   2.00-3.00   sec   262 MBytes  2.20 Gbits/sec    0    178 KBytes       
[  6]   3.00-4.00   sec   258 MBytes  2.17 Gbits/sec    0    178 KBytes       
[  6]   4.00-5.00   sec   260 MBytes  2.18 Gbits/sec    0    178 KBytes       
[  6]   5.00-6.00   sec   252 MBytes  2.12 Gbits/sec    0    178 KBytes       
[  6]   6.00-7.00   sec   256 MBytes  2.14 Gbits/sec    0    178 KBytes       
[  6]   7.00-8.00   sec   263 MBytes  2.20 Gbits/sec    0    178 KBytes       
[  6]   8.00-9.00   sec   257 MBytes  2.16 Gbits/sec    0    178 KBytes       
[  6]   9.00-10.00  sec   242 MBytes  2.03 Gbits/sec    0    178 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  6]   0.00-10.00  sec  2.50 GBytes  2.15 Gbits/sec    0            sender
[  6]   0.00-10.00  sec  0.00 Bytes  0.00 bits/sec                  receiver

iperf Done.
```

As you can see, the performance gain is quite significant. It is based on our implementation of `mnet` and `mkernel`. The IPerf3 code is available [here](https://git.robur.coop/robur/iperf3). I haven’t carried out a comparison with `mirage-tcpip`, however (there may be issues relating to the scheduler, particularly between Miou/`mkernel` and `lwt`).

## Tests

A test has been added to verify the integrity of the data transmitted over the network. In theory, the test could work perfectly well without this PR (it simply sends a ping with data), but I wanted to check that, under high load, I wasn’t misinterpreting the indices and buffers.

## Improvements

The number of comments is inversely proportional to my confidence in the code. In this case, there are a lot of comments. I hope the code is sufficiently documented, but above all it needs to be tested. In short, it’s a major change, but I think it’s definitely worth it to close the performance gap between hvt and VirtIO or Xen.

### DIFF
--- a/include/hvt_ring.h
+++ b/include/hvt_ring.h
@@ -52,7 +52,7 @@
 struct hvt_ring_entry {
     uint32_t operation;
     uint32_t handle;
-    uint64_t data; /* GPA of the data buffer */
+    HVT_GUEST_PTR(const void *) data; /* GPA of the data buffer */
     uint32_t len;
     uint32_t id;
     uint64_t _reserved;
--- a/include/hvt_ring.h
+++ b/include/hvt_ring.h
@@ -27,6 +27,7 @@
 #ifndef HVT_RING_H
 #define HVT_RING_H
 
+#include <stddef.h>
 #include <stdint.h>
 
 #define HVT_RING_SIZE 1024
@@ -93,6 +94,16 @@ struct hvt_ring {
     struct hvt_ring_entry entries[HVT_RING_SIZE];
     struct hvt_ring_commit commits[HVT_RING_SIZE];
 };
+
+/*
+ * Verify that guest-written and host-written fields live on distinct cache
+ * lines (64 bytes each), and that the data arrays start after exactly 2
+ * cache lines of index/control fields.
+ */
+_Static_assert(offsetof(struct hvt_ring, ent_head) == 64,
+               "ent_head must start at cache line 1 (offset 64)");
+_Static_assert(offsetof(struct hvt_ring, entries) == 128,
+               "entries[] must start after 2 cache lines (offset 128)");
 
 /*
  * Memory barriers.
--- a/include/hvt_ring.h
+++ b/include/hvt_ring.h
@@ -1,0 +1,145 @@
+/*
+ * Copyright (c) 2026 Contributors as noted in the AUTHORS file
+ *
+ * This file is part of Solo5, a sandboxed execution environment.
+ *
+ * Permission to use, copy, modify, and/or distribute this software
+ * for any purpose with or without fee is hereby granted, provided
+ * that the above copyright notice and this permission notice appear
+ * in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+ * NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * hvt_ring.h: Shared ring buffer definitions for ioeventfd-base I/O.
+ *
+ * This header file is used by both the tender (host) and the guest (bindings).
+ */
+
+#ifndef HVT_RING_H
+#define HVT_RING_H
+
+#include <stdint.h>
+
+#define HVT_RING_SIZE 1024
+#define HVT_RING_MASK (HVT_RING_SIZE - 1)
+#define HVT_RING_BUF_SIZE                                                      \
+    2048 /* per-slot data buffer (>= max ethernet frame) */
+
+/*
+ * Number of spin iterations the I/O thread performs before blocking on
+ * the eventfd. Tuned for ~1-2ns of spinning at typical CPU frequencies.
+ */
+#define HVT_RING_POLL_ITERS 4096
+
+/*
+ * Ring operations.
+ */
+#define HVT_RING_NET_WRITE 1
+#define HVT_RING_NET_READ  2
+
+/*
+ * Submission entry: written by the guest, consumed by the host.
+ */
+struct hvt_ring_entry {
+    uint32_t operation;
+    uint32_t handle;
+    uint64_t data; /* GPA of the data buffer */
+    uint32_t len;
+    uint32_t id;
+    uint64_t _reserved;
+};
+
+/*
+ * Commit entry: written by the host, consumed by the guest.
+ */
+struct hvt_ring_commit {
+    uint32_t id;
+    int32_t ret;
+    uint32_t len;
+    uint32_t _reserved;
+};
+
+/*
+ * Shared ring structure. Indices are separated into distinct cache lines to
+ * avoid false sharing between guest and host.
+ *
+ * Cache line 0: written ONLY by the guest (ent_tail, com_head)
+ * Cache line 1: written ONLY by thes host (ent_head, com_tail, needs_kick)
+ *
+ * This ensures each side only writes to its own cache line, eliminating
+ * false sharing bounces.
+ */
+struct hvt_ring {
+    /* Cache line 0: written ONLY by the guest */
+    volatile uint32_t ent_tail; /* produced by guest */
+    volatile uint32_t com_head; /* consumed by guest */
+    uint8_t _pad0[56];
+
+    /* Cache line 1: written ONLY by the host */
+    volatile uint32_t ent_head; /* consumed by host */
+    volatile uint32_t com_tail; /* produced by host */
+    volatile uint32_t needs_kick; /* set by host before sleeping */
+    uint8_t _pad1[52];
+
+    struct hvt_ring_entry entries[HVT_RING_SIZE];
+    struct hvt_ring_commit commits[HVT_RING_SIZE];
+};
+
+/*
+ * Memory barriers.
+ *
+ * hvt_wmb() / hvt_rmb(): used for the SPSC ring indices ([ent_tail],
+ * [com_tail]). On x86 TSO, stores are not re-ordered with sotrees and loads
+ * are not re-ordered with loads, so a compiler barrier suffices.
+ *
+ * hvt_mb(): full Store-Load barrier. Required for the [needs_kick] protocol
+ * where both sides do "store mine; load theirs". x86 TSO allows a younger load
+ * to bypass an older store to a different address (stored-buffer forwarding),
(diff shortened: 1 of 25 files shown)


hannesm:
Great work. Earlier discussion in https://github.com/Solo5/solo5/pull/290. I don't have the time at the moment to review, and hope someone else will review this amazing PR :)

include/hvt_ring.h
```diff
@@ -0,0 +1,145 @@
...
+
+    /* Cache line 1: written ONLY by the host */
+    volatile uint32_t ent_head; /* consumed by host */
+    volatile uint32_t com_tail; /* produced by host */
+    volatile uint32_t needs_kick; /* set by host before sleeping */
+    uint8_t _pad1[52];
+
+    struct hvt_ring_entry entries[HVT_RING_SIZE];
+    struct hvt_ring_commit commits[HVT_RING_SIZE];
+};
```
reynir, line 96: Maybe add `_Static_assert(offsetof(ent_head) == 64)` and `_Static_assert(offsetof(hvt_ring_entry) == 128)` or so? If it's important that they are offset by 64 bytes.

hannesm:
Thanks again for your PR.

On the top level, you described the path of an ethernet frame being sent. What is the path for receiving an ethernet frame? And does this PR as well touch the receiving side (and if yes, how much threads and ringbuffers are there? One for receive, one for send? Or one in total? Is there one ringbuffer per network device (I think the answer is no)).

dinosaure:
> What is the path for receiving an ethernet frame?

About reading Ethernet frame, this is where we also use the same _entries_ ring buffer (so one for reading & writing) and a "commit" of the operation from the `pthread` (on the host side). The idea is simple, we put a new entry on the _entries_ ring buffer to read with the buffer given by the user (so no `memcpy`) and we wait a new "commit" from the _commits_ ring buffer (this is where we have difference with writing where the write operations does **not** fill the _commits_ ring buffer).

> Is there one ringbuffer per network device (I think the answer is no).

Then, we have only one ring buffer for all net devices. Each entries have their "handle".

dinosaure:
I finally managed a test on FreeBSD (with a fix for #635 to see if CAPSICUM allows us to spawn a thread). This is the results:
```
root@mirage:~ # iperf3 -c 10.0.42.2
Connecting to host 10.0.42.2, port 5201
[  5] local 10.0.42.1 port 12938 connected to 10.0.42.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.02   sec  49.5 MBytes   408 Mbits/sec    0   64.2 KBytes       
[  5]   1.02-2.06   sec  51.6 MBytes   415 Mbits/sec    0   64.2 KBytes       
[  5]   2.06-3.01   sec  46.5 MBytes   414 Mbits/sec    0   64.2 KBytes       
[  5]   3.01-4.01   sec  49.2 MBytes   412 Mbits/sec    0   64.2 KBytes       
[  5]   4.01-5.00   sec  49.0 MBytes   414 Mbits/sec    0   64.2 KBytes       
[  5]   5.00-6.00   sec  48.8 MBytes   408 Mbits/sec    0   64.2 KBytes       
[  5]   6.00-7.01   sec  49.0 MBytes   407 Mbits/sec    0   64.2 KBytes       
[  5]   7.01-8.00   sec  49.2 MBytes   418 Mbits/sec    0   64.2 KBytes       
[  5]   8.00-9.00   sec  48.2 MBytes   405 Mbits/sec    0   64.2 KBytes       
[  5]   9.00-10.02  sec  50.9 MBytes   418 Mbits/sec    0   64.2 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.02  sec   492 MBytes   412 Mbits/sec    0            sender
[  5]   0.00-10.02  sec  0.00 Bytes  0.00 bits/sec                  receiver

iperf Done.
root@mirage:~ # iperf3 -c 10.0.42.2
Connecting to host 10.0.42.2, port 5201
[  5] local 10.0.42.1 port 21477 connected to 10.0.42.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.01   sec  48.0 MBytes   400 Mbits/sec    0   64.2 KBytes       
[  5]   1.01-2.00   sec  48.1 MBytes   405 Mbits/sec    0   64.2 KBytes       
[  5]   2.00-3.00   sec  47.1 MBytes   395 Mbits/sec    0   64.2 KBytes       
[  5]   3.00-4.00   sec  47.8 MBytes   401 Mbits/sec    0   64.2 KBytes       
[  5]   4.00-5.00   sec  49.0 MBytes   410 Mbits/sec    0   64.2 KBytes       
[  5]   5.00-6.03   sec  49.0 MBytes   400 Mbits/sec    0   64.2 KBytes       
[  5]   6.03-7.00   sec  46.1 MBytes   399 Mbits/sec    0   64.2 KBytes       
[  5]   7.00-8.03   sec  50.2 MBytes   409 Mbits/sec    0   64.2 KBytes       
[  5]   8.03-9.00   sec  45.6 MBytes   395 Mbits/sec    0   64.2 KBytes       
[  5]   9.00-10.02  sec  49.6 MBytes   407 Mbits/sec    0   64.2 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.02  sec   481 MBytes   402 Mbits/sec    0            sender
[  5]   0.00-10.02  sec  0.00 Bytes  0.00 bits/sec                  receiver

iperf Done.
```

The first `iperf3` is without this PR and the second is with this PR. So I don't observe a regression here or an improvement (I expected a little improvement but it's correct to see no changes on FreeBSD, the real trick is the `ioeventfd` which is available only on Linux).

As far as I can say, everything works. I did not test on OpenBSD and if someone is interesting to test it, go ahead but nothing should change for OpenBSD also (as nothing change for FreeBSD).

I also tested for a long term with unikernels ([`annuaire`](https://github.com/dinosaure/annuaire)) to see if everything works and: 1) we can observe a real improvement when it's about resolving a domain-name from root DNS servers 2) I did not notice bad (or unexpected) behaviors. These tests use a version of Solo5 with this PR and #638.

I think it's time to merge it and make a release :tada:.

PR #638, opened 2026-05-15
add seccomp sandbox to hvt on linux

### LINKED ISSUE #282: hvt_drop_privileges() behaviour for Linux and FreeBSD

pocopepe:
hvt_drop_privileges() on linux has been a stub since the issue went up back in 2020. this wires in a seccomp-bpf filter via libseccomp, using the same install pattern as spt_core.c: build the rules, export the BPF to a memfd, release libseccomp, then get the filter up and running ourselves with the raw seccomp() syscall.

allowlist is what the run loop actually pokes - ioctl, read, write, pread64/pwrite64, epoll_pwait, timerfd_settime, clock_gettime, exit_group, rt_sigreturn - plus a handful dumpcore needs on abort. anything else gets SIGSYS.

one open question - the debug build's dumpcore module needs 9 extra syscalls (getpid, openat, close, writev, lseek, brk, mmap, munmap, mincore), and since hvt_seccomp_linux.c is compiled once and linked into both binaries, production carries them too. happy to split per build if you guys would prefer the tighter posture, just wanted to land the filter first :)

Closes #282

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,5 +21,6 @@ Michael LeMay
 Nikhil AP
 Ricardo Koller (IBM)
 Stefan Berger (IBM)
+Viju Sanjai
 Waldir Pimenta
 Wei Chen (ARM)
--- a/bindings/hvt/net.c
+++ b/bindings/hvt/net.c
@@ -94,7 +94,7 @@ solo5_result_t solo5_net_write(solo5_handle_t handle, const uint8_t *buf,
         memcpy(write_bufs[idx], buf, size);
         ent->operation = HVT_RING_NET_WRITE;
         ent->handle = handle;
-        ent->data = (uint64_t)(uintptr_t)write_bufs[idx];
+        ent->data = write_bufs[idx];
         ent->len = size;
         ent->id = ring_req_id++;
         ring_submit(net_ring);
@@ -142,7 +142,7 @@ solo5_result_t solo5_net_read(solo5_handle_t handle, uint8_t *buf, size_t size,
 
         ent->operation = HVT_RING_NET_READ;
         ent->handle = handle;
-        ent->data = (uint64_t)(uintptr_t)buf;
+        ent->data = buf;
         ent->len = size;
         ent->id = ring_req_id++;
 
--- a/bindings/hvt/net.c
+++ b/bindings/hvt/net.c
@@ -19,12 +19,91 @@
  */
 
 #include "bindings.h"
+#include "hvt_ring.h"
 
 static const struct mft *mft;
+static struct hvt_ring *net_ring;
+static uint32_t ring_req_id;
+/* Here, we’re doing things differently from VirtIO, which allocates buffers
+ * using `malloc()`. Our buffers will be in the .bss section! */
+static uint8_t write_bufs[HVT_RING_SIZE][HVT_RING_BUF_SIZE];
+
+static inline void cpu_relax(void)
+{
+#if defined(__x86_64__)
+    __asm__ __volatile__("pause" ::: "memory");
+#elif defined(__aarch64__)
+    __asm__ __volatile__("yield" ::: "memory");
+#endif
+}
+
+/*
+ * Submit a request to the ring and kick if needed.
+ * Does NOT wait for completion, caller decides whether to wait.
+ */
+static inline void ring_submit(struct hvt_ring *ring)
+{
+    /* as VirtIO, we add wmb() here to add a new entry. */
+    hvt_wmb();
+    ring->ent_tail++;
+
+    /* Kick suppression: only signal the I/O thread via ioeventfd if it has
+     * indicated it is about to sleep (needs_kick == 1). When the thread is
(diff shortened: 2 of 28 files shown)


hannesm:
Thanks for your contribution. We're slightly busy at the moment; it'll take some time to review it.

pocopepe:
no rush at all, I'll be around whenever you get to it

dinosaure:
I did a quick look and it's an interesting proposal indeed.

> happy to split per build if you guys would prefer the tighter posture

I think it's a good path.

pocopepe:
thanks, already tried a split version on another branch, if the seccomp filter looks good I'll be happy to merge it in here.

dinosaure:
For me, the list of syscalls seems good. I can give a try to some of my unikernels but, thanks to Solo**5**, we should not have much more syscalls when it's about running an unikernel.

However, are you aware of this PR #637, which makes quite significant changes to the way `solo5-hvt` transmits Ethernet packets? It would be brilliant if your PR could incorporate this work, and we could then release a single update to Solo5 (with a breaking change).

pocopepe:
wasn't aware of it, just got looking into it now. did a quick syscall trace on your pr, and combining it with the filter needs a couple changes; the net I/O thread means the filter has to go `SECCOMP_FILTER_FLAG_TSYNC` to cover it, plus 8 new syscalls; `futex, munmap, madvise, exit, close, mmap, mprotect, rt_sigprocmask`. otherwise the thread runs unsandboxed and shutdown hangs on `pthread_join`→`futex`.

only thing i'm unsure about: `mmap` and `mprotect` are in that list just because the I/O thread is still mid-startup when the filter kicks in. if the thread finished starting before we lock things down, those two wouldn't need allowing at all. is there a clean spot to get the thread fully up before drop_privileges? happy to hear what you think.
(could also leave the thread uncovered and just add `futex, close`, but that defeats the point of the filter imo.)

dinosaure:
> is there a clean spot to get the thread fully up before drop_privileges

Normally, after this line:
https://github.com/Solo5/solo5/pull/637/changes#diff-3cedc7030541356007d7edb059d08a192dd9060d52ba045b219d2fbd3fbfa9c4R534-R535

the `pthread` is fully setup. `hvt_drop_privileges` is called really at the end (after `setup_modules`) so we should not include `mmap` and `mprotect`. Can you explain about `madvise` & `rt_sigprocmask`? I think it's related to `pthread`

pocopepe:
those come from [`free(ta)`](https://github.com/Solo5/solo5/blob/11ac6a522f023f49809391faffa9e16ed09d9d54/tenders/hvt/hvt_module_net.c#L337) in `io_thread_net_fn`; it's the thread's only alloc call, so glibc spins up a 128mb arena just to free one struct. free `ta` off the thread (in `kill_net_pthread` after the join, or make it static) and `mmap`/`mprotect`/`munmap` all drop. any reason it's freed on the thread?
my side stays light either way: filter goes `SECCOMP_FILTER_FLAG_TSYNC` to cover the thread, plus a few allowlist entries. runs clean with test_net_ring.

worth moving the free to keep `mmap`/`mprotect` out?

dinosaure:
> any reason it's freed on the thread?

Yeah it's needed, the `ta` value does go outside the `setup` function (and can not be transferred to the caller to be free-ed then elsewhere). So all syscalls seems legitimate. So can you:
1) rebase on top of #637 (and include required syscalls for our pthread)
2) split out for dbg syscalls

It will be perfect then and I will try to test it on my own way.

pocopepe:
done. rebased on #637, TSYNC covers the I/O thread, dumpcore syscalls split to debug only. ready for you to test, lemme know if this looks good to you :)

dinosaure:
Ok I finally tested it with some unikernels and it seems fine (at least, Linux tells me that the filter is applied and I'm able to manipulate a net interface and a block device). Let's merge.

PR #639, opened 2026-05-28
Prepare v0.11.0

dinosaure:


CHANGES.md
```diff
@@ -1,3 +1,37 @@
...
+  change mainly concerns Linux and the use of `eventfd(2)` to avoid a full VM
+  exit whenever one wishes to write to a network interface. A few benchmarks
+  (using `iperf3`) have been carried out on Linux and FreeBSD. For the former,
+  an improvement was observed (around three times more data can be transferred
+  per second), whilst no regression was observed for the latter (though no
+  improvement either).
+
+  This change **does not involve** an ABI change, meaning that the tender can
+  still run unikernels that have not been compiled with this version and vice
+  versa (a unikernel compiled with this version can be run by an older tender).
```
reynir:
```suggestion
  This change **does not involve** a breaking ABI change, meaning that the tender can
  still run unikernels that have not been compiled with this version and vice
  versa (a unikernel compiled with this version can be run by an older tender).
```
Maybe nit picking, but I would say the ABI *has* changed although in a way that is compatible.

PR #640, opened 2026-06-11
Port Solo5 to DragonFly BSD and nvmm

mneumann:
nvmm(4) is NetBSD's Virtual Machine Monitor which we ported to DragonFly BSD.

It should be trivial to port this further to NetBSD.

This only builds with clang (e.g. "pkg ins llvm18") and not with the gcc from base.

To build, I used:

    env TARGET_CC=clang18 TARGET_LD=ld.lld18 sh configure.sh
    gmake

### DIFF
--- a/.github/workflows/dragonfly.yml
+++ b/.github/workflows/dragonfly.yml
@@ -16,8 +16,11 @@ jobs:
         with:
           usesh: true
           prepare: |
-            pkg install -y gmake git llvm18
+            pkg install -y gmake git llvm18 gcc9
           run: |
             git config --global --add safe.directory /home/runner/work/solo5/solo5
             env TARGET_LD=ld.lld18 TARGET_CC=clang18 ./configure.sh
+            gmake
+            gmake clean
+            env TARGET_CC=gcc9 ./configure.sh
             gmake
--- a/.github/workflows/dragonfly.yml
+++ b/.github/workflows/dragonfly.yml
@@ -1,0 +1,23 @@
+name: DragonFly
+on: [push, pull_request]
+jobs:
+  Build:
+    name: AMD64
+    strategy:
+      matrix:
+        operating-system: [ubuntu-latest]
+    runs-on: ${{ matrix.operating-system }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: DragonFly action
+        id: test
+        uses: vmactions/dragonflybsd-vm@v1
+        with:
+          usesh: true
+          prepare: |
+            pkg install -y gmake git llvm18
+          run: |
+            git config --global --add safe.directory /home/runner/work/solo5/solo5
+            env TARGET_LD=ld.lld18 TARGET_CC=clang18 ./configure.sh
+            gmake
--- a/configure.sh
+++ b/configure.sh
@@ -502,6 +502,14 @@ case ${CONFIG_HOST} in
     DragonFly)
         TARGET_LD="${TARGET_LD:-ld}"
         TARGET_OBJCOPY="${TARGET_OBJCOPY:-objcopy}"
+        TARGET_CC="${TARGET_CC:-cc}"
+        if CC="${TARGET_CC}" cc_is_gcc; then
+            major_version=$(echo __GNUC__ | ${TARGET_CC} -E -x c - | tail -n 1)
+            if [ "${major_version}" -lt 9 ]; then
+                err "gcc version ${major_version} of ${TARGET_CC} unsupported on DragonFly"
+                die "gcc 9+ or clang required on DragonFly"
+            fi
+        fi
         ;;
     *)
         die "Unsupported host system: ${CONFIG_HOST}"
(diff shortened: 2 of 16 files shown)


hannesm:
This looks fantastic. Some questions, though
- would you mind to adhere to clang-format?
- would you be able to add a .github/workflows/dragonflybsd (I guess https://github.com/vmactions/dragonflybsd-vm should be suitable) -- so at least building on DragonFlyBSD is checked in CI?

mneumann:
@hannesm perfect. I'll clang-format it. CI: I have never tried this myself on github. I only know that SeanTAllen from https://github.com/ponylang/ponyc has setup CI for DragonFly.

mneumann:
@hannesm I'll try later to add CI

mneumann:
@hannesm I added CI using your suggested method and rebased the whole thing into one commit. Building with gcc should also work, but there is an issue with toolchain/gen-headers.sh.

dinosaure:
> but there is an issue with toolchain/gen-headers.sh.

Do you know what is the issue? We probably can help on this topic.

mneumann:
@dinosaure:

These lines cause issues:

```
    # For GCC there isn't an equivalent of -nostdlibinc, so we need to
    # appropriate all of its internal headers.
    SRCDIR="$(${CONFIG_TARGET_CC} -print-file-name=include)"
    [ -d "${SRCDIR}" ] || die "Cannot determine gcc include directory"
    cp -R "${SRCDIR}/." ${DESTDIR} || \
        die "Failure copying host headers"
```

```
~/Dev/solo5 % cc -print-file-name=include
include
```

This ends up executing:

```
cp -R include/. include/x86_64-solo5-none-static
```

And this leads to this:

```
cp: include/x86_64-solo5-none-static/./x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static/x86_64-solo5-none-static: name too long (not copied)
```

dinosaure:
Is it possible to have the version of GCC used? We probably, at least, do a test and fail with a better message.

mneumann:
@dinosaure: 8.3

```
 % gcc -v
Using built-in specs.
COLLECT_GCC=/usr/libexec/gcc80/gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc80/lto-wrapper
Target: x86_64-pc-dragonflybsd
Configured with: DragonFly/x86_64 system compiler (c c++ LTO)
Thread model: posix
gcc version 8.3 [DragonFly] Release/2019-02-22 
```

It works with gcc 11

hannesm:
For historical reasons, the gen-headers.sh was developed when gcc was only used on GNU/Linux systems, while clang was used on Open/FreeBSD systems. Now, the purpose of gen-headers.sh is to copy the required header files, sometimes distributed by the system, or by the C compiler, to the cross-compilation environment. This is a bit differently on BSD vs Linux, since on BSD we need to compute all required headers (using the cc_get_header_deps function), while on Linux a `gcc -print-file-name=include` was sufficient.

I've no deep insight why this changed on DragonFlyBSD between gcc 8.3 and 11.x. I'm curious whether a llvm-only build of solo5 is sufficient for DragonFlyBSD? Will the base system stay with gcc 8.3 for a long time (or are there plans to either upgrade to a newer gcc or to llvm)? If there is a roadmap, I'd just wait. Otherwise, this requires digging into how to find out where are the header files that need to be copied.

mneumann:
@hannesm @dinosaure It seems to be an issue with the base compiler of DragonFly, which is gcc 8. In commit [22b31f2](https://github.com/Solo5/solo5/pull/640/commits/22b31f280da6dfd1714241c898fc18ffbae7e2b0) I added a version check. gcc 9+ works on DragonFly, or clang.

mneumann-adgentek:
@hannesm ping :). let me know if I can do anything to get this merged.

dinosaure (approved):
Sorry for the long process but consider it as merged, thanks for your work!

PR #642, opened 2026-06-23
Stamp version.h from $SOLO5_VERSION for tarball pins

samoht:
`gen_version_h.sh` needs a version string when `.git` is absent. Today the only fallback is `version.h.distrib`, shipped in a `make distrib` release tarball. An opam tarball pin (or any exported git snapshot, e.g. a GitHub codeload archive) has neither a Git tree nor a `version.h.distrib`, so the build fails at `gen_version_h.sh`.

This PR adds `$SOLO5_VERSION` as a third source, after `git describe` and `version.h.distrib`, and passes it from the opam files as `%{version}%`. So `opam pin add solo5 <tarball>` and an opam-repository release from a codeload archive both build without a Git tree.

The macOS cross-build PR #641 carries this as a temporary commit so its CI can tar-pin solo5; that commit drops once this lands.

### DIFF
--- a/opam/solo5-cross-aarch64.opam
+++ b/opam/solo5-cross-aarch64.opam
@@ -11,9 +11,9 @@ license: "ISC"
 dev-repo: "git+https://github.com/solo5/solo5.git"
 build: [
   ["env" "TARGET_CC=aarch64-linux-gnu-gcc" "TARGET_LD=aarch64-linux-gnu-ld" "TARGET_OBJCOPY=aarch64-linux-gnu-objcopy" "./configure.sh" "--prefix=%{prefix}%"]
-  [make "V=1"]
+  ["env" "SOLO5_VERSION=%{version}%" make "V=1"]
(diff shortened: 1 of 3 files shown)


hannesm:
I struggle to understand why this is needed. Now given the amount of maintenance of solo5, I'm not very comfortable to add this much more compexity (in the shell script, as well as opam pushing through some environment variable) for which I don't understand the use case. Why can't you use a git pin or a released version?

samoht:
The current issue is that solo5 breaks if you try tin pin install it from a tarball. I agree this is a narrow use-case and I'm fine to drop it if you don't think that's useful.

dinosaure:
> try tin pin install it from a tarball.

It is, in fact, far too specific when other suitable solutions (such as pin a Git repository) are available and do not have the problem described here.

samoht:
Fair. Thanks for reviewing this anyway!

PR #644, opened 2026-06-29
Apply HVT_DROP_PRIVILEGES=0 on solo5-hvt-debug's objects

dinosaure:
The purpose of solo5-hvt-debug is only for debug and we don't need to drop privileges when we would like to run a HVT unikernel (solo5-hvt-debug interacts with multiple things if we take into account dumpcore and gdb). This commit splits the way to build solo5-hvt-debug and apply -DHVT_DROP_PRIVILEGES=0 on sources. By this way, only solo5-hvt drops privileges.

### DIFF
--- a/tenders/GNUmakefile
+++ b/tenders/GNUmakefile
@@ -44,11 +44,18 @@ all_TARGETS += $(common_LIB)
 %.o: %.c %.d
 	$(HOSTCOMPILE.c)
 
+%.debug.o: %.c %.debug.d
+	$(HOSTCOMPILE.c)
+
 %.o: %.S %.d
 	$(HOSTCOMPILE.S)
 
+%.debug.o: %.debug.S %.debug.d
+	$(HOSTCOMPILE.S)
+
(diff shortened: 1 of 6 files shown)


hannesm:
stupid question, but isn't it worth to as well have the debug & dumpcore stuff be seccomp-protected with the right system calls? or is it too much effort for nothing?

dinosaure:
I think that, according to this comment, the original intention was to support sandboxing only for `solo5-hvt`, and that `solo5-hvt-debug` should not drop privileges.
https://github.com/Solo5/solo5/blob/d31546c84591a023f706c4a75d207a76f2cbdb69/tenders/hvt/hvt.h#L41-L49

 Furthermore, I think that given the `dumpcore` module (which handles a file) and the `gdb` module (which communicates with `gdb`), we would end up having to support a lot of system calls, and I don’t really have the time to work through them one by one.

hannesm:
I understand and approve @dinosaure

PR #645, opened 2026-06-30
Verify that we have, at least, GCC 10 to compile Solo5

dinosaure:
Our `configure.sh` will fail if we don't have GCC 10. Spotted by #576 and #640 also. /cc @mneumann can you confirm that GCC 10 is available on DragonFly (and update your PR to install it instead of GCC 9)? We also should do the same for Clang (and verify that we have, at least, Clang 8) but we can do that later (and in another PR).

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -262,15 +262,18 @@ HOST_CC_MACHINE=$(${HOST_CC} -dumpmachine)
     die "Could not run '${HOST_CC} -dumpmachine', is your compiler working?"
 echo "${prog_NAME}: Using ${HOST_CC} for host compiler (${HOST_CC_MACHINE})"
 
+cc_version=$(${HOST_CC} -dumpversion)
+cc_major=$(echo "$cc_version" | cut -d '.' -f 1)
+if [ -z "$cc_major" ]; then
+    die "Impossible to recognize the version of cc: ${cc_version}"
+fi
 if CC="${HOST_CC}" cc_is_gcc; then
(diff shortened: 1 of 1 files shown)


hannesm:
from #576 I get we need gcc >= 8 or clang >= 10... also with the DragonFlyBSD port, it seems that gcc9 is sufficient there. :)

configure.sh
```diff
@@ -269,7 +269,7 @@ if CC="${HOST_CC}" cc_is_gcc; then
         die "Impossible to recognize the version of GCC: ${gcc_version}"
     fi
 
-    if [ "$major" -lt 10 ]; then
+    if [ "$major" -lt 9 ]; then
         die "GCC 10 (or later) is required"
```
hannesm: `GCC 9 (or later)`

mneumann:
@dinosaure gcc9 is available on DragonFly via "pkg install". gcc10 is **not**! 
FWIW, gcc11 to gcc15 are available.

dinosaure:
Ok let's keep GCC 9. If people agreed with this PR, I can merge it.

hannesm:
this is fine, maybe add the clang version check in the same commit/PR?

hannesm:
I pushed a commit which as well checks the clang version. Please back it out if you're not happy with it.

It uses `cc -dumpversion` instead of `cc --version` to avoid parsing...

dinosaure:
Thanks!

PR #646, opened 2026-07-01
Use GCC 9 as the host compiler for DragonFly

dinosaure:


.github/workflows/dragonfly.yml
```diff
@@ -19,8 +19,8 @@ jobs:
             pkg install -y gmake git llvm18 gcc9
           run: |
             git config --global --add safe.directory /home/runner/work/solo5/solo5
-            env TARGET_LD=ld.lld18 TARGET_CC=clang18 ./configure.sh
+            env HOST_CC=gcc9 TARGET_LD=ld.lld18 TARGET_CC=clang18 ./configure.sh
```
hannesm: here shouldn't we use clang18?

PR #647, opened 2026-07-01
Be able to softly power-off a virtio unikernel

dinosaure:
This PR wants to solve a very old issue #499 and took a big inspiration on #500 and #501. I think mostly @mbacarella or @kit-ty-kate will be interested by such PR. I tried it on Google Compute Engine and I can confirm that an unikernel power-off correctly. May be, at a time, the big issue was about the unavailability of ACPI tables (due to `pagetable.S`).


dinosaure:
Oh the ratio is interesting but it's mostly due to a refactor of `pagetable.S` (and the use of `rept`).

kit-ty-kate:
> @kit-ty-kate will be interested by such PR

I'm not using GCE anymore, but nonetheless it's a nice improvement to see

PR #648, opened 2026-07-02
Use a clean GIT_VERSION value instead of generating one from a Git source tree which was updated (we deleted the -Werror)

dinosaure:
In the release path, we take everything and delete the -Werror argument into the Makefile.common (which is tracked by Git). The GIT_VERSION results with a `-dirty` suffix. This patch permits to pass a clean GIT_VERSION regardless some changes.

/cc @hannesm

Noticed on the last release, `solo5-hvt` shows me `v0.11.0-dirty`.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -51,7 +51,7 @@ gen-version-h:
 
 distrib-gen-version-h:
 	@echo "GEN $(VERSION_H).distrib"
-	scripts/gen_version_h.sh $(VERSION_H).distrib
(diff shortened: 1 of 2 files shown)


hannesm:
could we instead run the gen_version before we remove the Werror?

dinosaure:
Hmmhmm, it assumes a certain order of execution here:
https://github.com/Solo5/solo5/blob/5b41bd6c900fe0ab2fce9b5ef4d95dbd871c023c/GNUmakefile#L66

And I'm not sure that we can trust `make` on that topic :confused:

hannesm:
Let's close this, and if it happens again, revise our thoughts about that.

PR #652, opened 2026-07-10
Discard the unused .eh_frame unwind tables

samoht:
Every unikernel image maps `.eh_frame` and `.eh_frame_hdr`, but nothing in a Solo5 guest unwinds through them: the tenders don't, and the runtimes that target Solo5 (OCaml/MirageOS, freestanding C) either use their own unwinding or none at all. Because the sections are `SHF_ALLOC` they land in a `PT_LOAD` segment, so they are loaded into every image and `strip` cannot reclaim them.

This moves `.eh_frame`/`.eh_frame_hdr` to `/DISCARD/` in all six bindings' linker scripts, the same way the scripts already discard the GNU NOTEs. It removes about 0.5 MiB from every image: a hello-world unikernel drops from 1,2M to 600k, and larger images shrink by the same absolute amount.

A guest that genuinely relies on DWARF unwinding (C++ exceptions, Rust `panic=unwind`) would need these tables. If that is a use case worth keeping, I am happy to gate the discard behind an option instead.

### DIFF
--- a/bindings/solo5_hvt.lds
+++ b/bindings/solo5_hvt.lds
@@ -93,10 +93,6 @@ SECTIONS {
         *(.rodata)
         *(.rodata.*)
     } :rodata
-    .eh_frame :
-    {
-        *(.eh_frame)
-    }
 
     . = ALIGN(CONSTANT(MAXPAGESIZE));
     _erodata = .;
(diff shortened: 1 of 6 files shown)


dinosaure:
> A guest that genuinely relies on DWARF unwinding (C++ exceptions, Rust panic=unwind) would need these tables. If that is a use case worth keeping, I am happy to gate the discard behind an option instead.

It is, in fact, necessary. Solo5 is not exclusive to OCaml.

hannesm:
Does the solo5-hvt-debug with the gdb interaction use eh_frame? I'm not sure whether this is part of the test suite.

> Because the sections are SHF_ALLOC they land in a PT_LOAD segment, so they are loaded into every image and strip cannot reclaim them.

Is there another path to solve this? Can there be arguments passed to "strip" to allow discarding these? Can it be marked differently? Would a custom strip (how hard can it be?) be sensible? Are there different strip utilities with the same behaviour, or are they all the same?

PR #655, opened 2026-07-20
Clean up various small points

shym:
This PR proposes various clean-ups that came up while preparing another PR. It is opened independently as really unrelated to the goal of that future PR.


hannesm:
I looked through these commits, and they look fine to me.

PR #656, opened 2026-07-20
Add arm64 macOS support in the build system

shym:
This PR is another take (compare with #641) to add some support for building on arm64 macOS. It is built on top of #655 and adds only 4 commits.
It is designed to make no difference for currently supported platforms.

The main changes are:
- a new `--disable-elftool` configure option (that does what you’d expect),
- a new `--path-prepend` configure option, to add new entries in `PATH` before invoking tools in the toolchain wrappers; this is necessary to be able to use homebrew’s `llvm` and `lld` packages (which we need to generate ELF unikernels) which do not install the binaries in a directory in `PATH` to avoid conflicts with versions of the same tools provided by macOS,
- new cases in `configure.sh` to enable arm64 macOS.

Compared with #641, this PR doesn’t modify opam packages and leave such a change for another PR.


configure.sh
```diff
@@ -52,9 +52,17 @@ Options:
     --prefix=DIR
         Installation prefix (default: /usr/local).
 
+    --path-prepend=VAL
+        Prepend VAL to PATH in toolchain. For instance:
+        --path-prepend=/usr/local/llvm-21/bin:
+        (note the required explicit ':' separator at the end)
```
dinosaure, line 58: Is the requirement to include the colon a convention? I would prefer that we (`./configure.sh`) are the ones to add the colon, or that this be verified.

configure.sh
```diff
@@ -400,7 +425,22 @@ fi
...
+case ${HOST_CC_MACHINE} in
+    arm64-*darwin*|aarch64-*darwin*)
+        TARGET_CC="${TARGET_CC:-clang -target aarch64-linux-gnu}"
+        # Without the following flag, I see errors such as:
+        # ld.lld: error: runtime/libcamlrun.a(domain.b.o):(function caml_thread_running_on_expected_domain: .text+0x1c): relocation R_AARCH64_TLSLE_ADD_TPREL_HI12 out of range: 18446744073709547536 is not in [0, 16777215]; references 'previous_domain_id'
+        # By changing the possible TLS size, the generated code use different
+        # instructions with which a negative offset is admissible
+        # Is there a better fix?
+        # Because of this doubt, make it overridable, unlike in the other case
+        TARGET_CC_CFLAGS="${TARGET_CC_CFLAGS:--mtls-size=32}"
```
dinosaure, line 437: I need more time to understand the implication of that and it seems really tricky.

dinosaure (commented):
We can probably split out `DISABLE_ELFTOOL` into another PR which more general and can be easily acceptable for Solo5. I have some reservations about `path-prepend` (although the idea seems ok to me). Finally, we'll take the time to understand `-mtls-size 32` and its implications.

configure.sh
shym: I agree.
The initial rationale was thinking about portability to Windows (which is still a step further, I admit). But on second thought, as we are using those in `sh` scripts, I’d guess that even Windows’ `sh` would take care of interpreting `:`. So I went in #658 with adding the separator in the configure script, and while at it, support various occurrences of the option.

configure.sh
shym: I’d like to understand better where the breaking relocation is located. My guess is that clang makes assumptions about the layout that will be used in the linker script so that it assumes the offset will be positive and that those assumptions are not held in our linker scripts but I didn’t understand the full story there.

shym:
I extracted #657 and #658. Let’s take some more time to understand what is happening with those relocations.

shym:
I’ve spent some time trying to understand a bit better what happens with the TLS size parameter I had to tweak.
With both homebrew’s clang and debian’s aarch64-gcc, I see the following assembler code for `caml_thread_record_domain_id` (that fails to link):
```asm
caml_thread_record_domain_id:           // @caml_thread_record_domain_id
	mrs	x8, TPIDR_EL0
	add	x8, x8, :tprel_hi12:previous_domain_id
	add	x8, x8, :tprel_lo12_nc:previous_domain_id
	str	x0, [x8]
	ret
```
but then homebrew’s lld fails to link it in libcamlrun, while debian’s aarch64-ld succeeds.
As both compilers generate the same assembler code, I suppose that the issue lies in how homebrew’s lld interprets the linker script (I don’t know very much about linker scripts, but my understanding is that lld doesn’t pretend to be fully compatible with GNU ld on their interpretation).

For the record, with the parameter to set the TLS size to 32, I get the following assembler code in homebrew’s clang:

```asm
caml_thread_record_domain_id:           // @caml_thread_record_domain_id
	movz	x8, #:tprel_g1:previous_domain_id       // high 16 bits
	mrs	x9, TPIDR_EL0
	movk	x8, #:tprel_g0_nc:previous_domain_id    // low 16 bits
	str	x0, [x9, x8]
	ret
```

so the linker is not constrained on the offset being a positive number.

I couldn’t reproduce the issue with debian’s clang. @samoht didn’t see that issue either IIUC. I wonder whether that appears only on some versions of lld.

PR #657, opened 2026-07-22
Add a `--disable-elftool` configure option

shym:
Make it possible to build the project without the ELF tool. This is particularly interesting on non-ELF systems (read macOS) where the tool doesn’t compile.

Extracted from #656.


dinosaure (approved):

hannesm:
thanks

PR #660, opened 2026-08-10
Roundup the memory area for our ringbuffer when we have, at least, one net device

### LINKED ISSUE #659: attaching a net device leaves guest memory not 2mb aligned
Since v0.11.0, attaching a network device to `solo5-hvt` makes the tender report a guest memory size that isn't 2MB aligned. `hvt_mem_size_roundup()` rounds guest memory to `X86_GUEST_PAGE_SIZE` (2MB), but `hvt_net_reserve_ring()` carves the ring off the top with only 4KB alignment and hands that straight to `guest_mem_size`.

Found it with solo5's own test unikernels, this is v0.12.0:

```
pocopepe@pocopepe:~/contrib/test/solo5/tests$ sudo ../tenders/hvt/solo5-hvt test_hello/test_hello.hvt 2>&1 | grep -E "Memory map|heap >="
Solo5:       heap >= 0x30c000 < stack < 0x20000000
pocopepe@pocopepe:~/contrib/test/solo5/tests$ sudo timeout 5 ../tenders/hvt/solo5-hvt --net:service0=tap100 test_net/test_net.hvt 2>&1 | grep -E "Memory map|heap >="
Solo5: Memory map: 513 MB addressable:
Solo5:       heap >= 0x30c000 < stack < 0x201f3000
```

`0x201f3000` is `0xd000` below `0x20200000`, which is `sizeof(struct hvt_ring)`.

Same thing on v0.10.1 stays aligned and doesn't gain the extra MB:

```
pocopepe@pocopepe:~/contrib/test/solo5-v0.10.1$ sudo ./tenders/hvt/solo5-hvt --net:service0=tap100 tests/test_net/test_net.hvt 2>&1 | grep -E "Memory map|heap >="
Solo5:       heap >= 0x10c000 < stack < 0x20000000
```

We hit this with Rumprun unikernels under urunc. Every networked image dies during guest boot, non-networked ones on the same tender are fine. Rumprun gives the whole reported range to its page allocator (`heap_size = mem_size - heap_start`), so it overlaps the stack. Nothing signals the change either, a working Mirage unikernel and a failing Rumprun one both declare ABI version 2. Changing the mask to `~0x1FFFFFULL` fixes it and costs at most 2MB. `tests/run-tests.sh` passes 81/81 before and after, the suite just doesn't have a guest that leans on the alignment.

Seems to be a one liner change, I'm happy to open a pr with the change and maybe a test to prevent regression if someone could double check bug itself.

dinosaure:
It seems that the memory given to the unikernel must be page-aligned. We roundup the needed memory for our ringbuffer and, by this way, we ensure to give a page-aligned memory area to the unikernel. It seems that it's a requirement for rumprun.

It should fix #659

### DIFF
--- a/tenders/hvt/hvt.h
+++ b/tenders/hvt/hvt.h
@@ -104,9 +104,10 @@ void hvt_net_reserve_ring(struct hvt *hvt, struct mft *mft);
 void hvt_mem_size_roundup(size_t *mem_size);
 
 /*
- * Returns the extra guest memory needed for the network ring buffer, or 0 if
- * no NET_BASIC device is attached in the manifest. Must be called after
- * command-line parsing has set the attached flags.
+ * Returns the extra guest memory needed for the network ring buffer, rounded
+ * up to the architecture page boundary, or 0 if no NET_BASIC device is
+ * attached in the manifest. The return value is roundup on the system page.
+ * Must be called after command-line parsing has set the attached flags.
  */
 size_t hvt_net_mem_overhead(struct mft *mft);
 
--- a/tenders/hvt/hvt_boot_info.c
+++ b/tenders/hvt/hvt_boot_info.c
@@ -107,11 +107,11 @@ void hvt_boot_info_init(struct hvt *hvt, hvt_gpa_t gpa_kend, int cmdline_argc,
         if (ring_active) {
             bi->host_features |= HVT_FEATURE_RING_IO;
             bi->net_ring = hvb->net_ring_gpa;
-            /*
-             * Exclude the ring area from usable guest memory so the
-             * guest's heap allocator does not grow on it.
-             */
-            bi->mem_size = hvb->net_ring_gpa;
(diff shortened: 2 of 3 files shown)


pocopepe:
I'd ended up with something pretty close locally, masked the ring gpa down to 2mb rather than rounding the reserve up. Lands on the same address as far as I can tell, though going through hvt_mem_size_roundup probably ages better since it picks up the granularity per arch on its own. Either way I've tested yours against the image that got me looking into this and it seems to sort it.

dinosaure:
Ok, so you confirm that this patch works for you for rumprun? I think, we never ensure loud that the memory given to the unikernel must be aligned.

pocopepe:
took a while to run the whole e2e suite of where I found the bug, works for rumprum, ran the full HVT rumprun suite against roundup-ringbuffer-size, all passing now.

dinosaure:
Ok, thanks, we will cut a release soon.

PR #661, opened 2026-08-11
remove dead per module seccomp hook from hvt on linux

pocopepe:
Was reading the v0.12.0 changelog and hit the solo5-hvt-debug fix, [#644](https://github.com/Solo5/solo5/pull/644),  the crash fix for my seccomp work in [#638](https://github.com/Solo5/solo5/pull/638). Turns out the fix left some of my old code behind: the per-module seccomp hook and dumpcore's rules that used it, both dead since debug stopped sandboxing entirely.

Cleaned it up. solo5-hvt itself untouched, still sandboxed, still builds/runs/tests fine.


dinosaure (approved):
A double check will be appreciate 👍

hannesm:
looks fine to me

PR #662, opened 2026-08-12
docs: fix broken header links in architecture.md

danila-permogorskii:
The links to the public headers pointed at the installed layout ($PREFIX/include/solo5/) rather than the repository layout, where the headers live directly in include/. All five render as 404s on GitHub. The link on line 134 was additionally missing the leading ../ needed to escape docs/.

Also drop the reference to .travis.yml, removed in [5e56936](https://github.com/Solo5/solo5/commit/5e56936), and mention the GitHub Actions workflows instead.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,7 +11,6 @@ Individual Contributors
 Adrian-Ken Rueegsegger
 Adam Steen
 Dan Williams (IBM)
-Danila Permogorskii
 Emery Hemingway (Genode Labs)
 Felix Moebius
 Gabriel Jaldon
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@ Individual Contributors
 Adrian-Ken Rueegsegger
 Adam Steen
 Dan Williams (IBM)
+Danila Permogorskii
 Emery Hemingway (Genode Labs)
 Felix Moebius
 Gabriel Jaldon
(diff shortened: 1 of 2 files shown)


danila-permogorskii:
Hi! This is my first start contributing to solo5. I'm planning to work on this project continiously, so I decided to start with something small and improve it incrementally

hannesm:
this looks good. maybe worth to adjust the docs and repository with reality -- that surf-build is not around anymore (since the host got lost), and so the `build.sh` isn't actually used anywhere... but that can be done in a separate PR (also depending on whether any maintainer actually thinks we will bring surf & CI back)

hannesm:
For such a small PR I wouldn't put myself into the AUTHORS file, but rather wait a couple more contributions, maybe even something substantial. But of course there is no clear rule when to add yourself as an author.

danila-permogorskii:
I agree with you and I can remove it. I just interpreted it as a rule, but I see your point

dinosaure (approved):

danila-permogorskii:
> this looks good. maybe worth to adjust the docs and repository with reality -- that surf-build is not around anymore (since the host got lost), and so the `build.sh` isn't actually used anywhere... but that can be done in a separate PR (also depending on whether any maintainer actually thinks we will bring surf & CI back)

Got it

PR #667, opened 2026-08-16
configure.sh: pass -no-pie to the compiler driver, not ld

Zaneham:
Follow-up to #629. `-no-pie` is a gcc driver flag, but it went in inside the `-Wl,` list, so ld gets it instead. Recent ld happens to accept it. GNU ld 2.35 doesn't, and parses it as `-n -o -pie`, which writes every unikernel to a file called `-pie` while make still exits 0.

Tested on ppc64le with ld 2.35, where this takes the build from no unikernels at all to all 24, and they run.


hannesm:
thanks, this looks good to me /cc @Firobe

Firobe (approved):

PR #668, opened 2026-08-16
Add riscv64 support to the spt target

Zaneham:
Adds riscv64 to the spt target, following the ppc64le port. Three new files plus one arm on each of the existing `#if` ladders, and a build-only CI job. The tender runs fine under qemu-user but the guest it maps at a fixed low address doesn't, so the suite needs real hardware.

Built and tested on a riscv64 Debian 13 box, all eleven spt tests pass including `test_tls`. Ran the same tests on ppc64le against main and against this branch to check the shared files, identical results. Logs are [here](https://github.com/Zaneham/bits-and-bobs/tree/master/solo5/2026-08-16-riscv64-spt).

btw RISC-V is Variant I with `tp` one past the end of the TCB, so thread variables start at `tp` and `solo5_tls_init`'s self-pointer write would land on `.tdata`. I've suppressed it for riscv64 rather than reserving a gap, since nothing reads it when `tp` is directly readable. Happy to do it differently.


Zaneham:
I had my NPP workspace open when I got the email so I just quickly fixed it and tested it using Clang on my machine so it should be alright now.

hannesm:
I've not much knowledge about Linux, RISC-V, and spt... this looks nice (as in no surprises, more ifdefs, etc.).

Do you have plans for hvt & RISC-V ~~(I guess the related question is whether there's kvm support for RISC-V)~~?

I guess the other question is whether there are boards (or servers) available with hypervisor extensions? Do you own any/have access to some?

EDIT: figured out there's KVM support for RISC-V in Linux.

Zaneham:
I checked my hardware and I don't have access to those extensions. Not to say I couldn't in the future (I got sent some hardware from a company that makes RISC-V gpus but I'm still getting that box set up). I'll see if I can find somewhere that grants hardware access too.

It's definitely not a blocker for me though because `qemu-system-riscv64` emulates H and that's how most KVM/RISC-V was done before hardware was even manufactured. 

So yes, I'd be up for it. Been enjoying operating systems so far.

dinosaure:
Can you add your `clang-format` commit into our `.git-blame-ignore-revs`? Otherwise, I'm ok with this PR (just a concerns about how we can properly test it but that's fine).

Zaneham:
Forgot to reply but I've added it @dinosaure :-)

dinosaure:
Ok I think it's good to merge, we will describe it however as experimental (as it's the code for your aarch64 cross-compiler). Thanks for your work.

PR #670, opened 2026-08-18
Allocate our ring buffer inside the memory asked by the user and introduce the new --no-net-ring option

dinosaure:
The ring-buffer for net interfaces are now located inside the asked amount of memory. We also keep the page-aligned constrainst of the heap and the stack for rumprun and we introduce the new option --no-net-ring which does not set the ring-buffer for the guest.


bindings/hvt/net.c
```diff
@@ -205,5 +205,7 @@ void net_init(const struct hvt_boot_info *bi)
 
     if ((bi->host_features & HVT_FEATURE_RING_IO) && bi->net_ring != 0) {
         net_ring = (struct hvt_ring *)(uintptr_t)bi->net_ring;
+	log(INFO, "Solo5:        net @ (0x%llx - 0x%llx)\n",
```
hannesm: can we add something with the word "ring" or "ringbuffer", please?

tenders/hvt/hvt_main.c
```diff
@@ -227,6 +228,12 @@ int main(int argc, char **argv)
             argc--;
             argv++;
         }
+	if (strncmp("--no-net-ring", *argv, 13) == 0) {
+	    no_net_ring = 1;
+	    matched = 1;
+	    argc--;
+	    argv++;
+	}
```
hannesm: should be documented in the usage (line 102)

tenders/hvt/hvt_main.c
```diff
@@ -252,20 +259,13 @@ int main(int argc, char **argv)
         err(1, "Could not install signal handler");
 
     hvt_mem_size(&mem_size);
-
-    size_t net_overhead = hvt_net_mem_overhead(mft);
-    if (net_overhead > 0) {
-        mem_size += net_overhead;
-        hvt_mem_size_roundup(&mem_size);
-    }
```
hannesm, line 260: are hvt_net_mem_overhead and hvt_mem_size_roundup used at all now?

where does this "stack is aligned to 2MB" happen?

hannesm:
indeed testing, we should have something to ensure the 2MB alignment from #659

bindings/hvt/net.c
reynir: This is also indented with a tab while the context uses spaces

tenders/hvt/hvt_main.c
reynir: It is done in `hvt_net_reserve_ring()` https://github.com/Solo5/solo5/blob/469c1a8d935b503fa553aef152d1d57d095fa3a4/tenders/hvt/hvt_module_net.c#L76-L99 called 7 lines below this comment.

tenders/hvt/hvt_main.c
```diff
@@ -252,20 +259,13 @@ int main(int argc, char **argv)
...
-
     struct hvt *hvt = hvt_init(mem_size);
 
     elf_load(elf_fd, elf_filename, hvt->mem, hvt->guest_mem_size,
              HVT_GUEST_MIN_BASE, hvt_guest_mprotect, hvt, &gpa_ep, &gpa_kend);
     close(elf_fd); /* Done with ELF binary */
 
-    hvt_net_reserve_ring(hvt, mft);
+    if (no_net_ring == 0) hvt_net_reserve_ring(hvt, mft);
```
reynir: Since we no longer adjust `mem_size` depending on hvt_ring or no I wonder if we can move this and the argument into `hvt_module_net.c`

tenders/hvt/hvt_main.c
hannesm: ..thanks, and while reading, I likely don't know much... but why `2 * reserve`? And why is nothing done with `reserve` later in that function (so, we do memset but only (`sizeof (struct hvt_ring)`).. so for me to understand, there's due to `roundup` some potential gap with memory we don't touch, right?

tenders/hvt/hvt_main.c
```diff
@@ -227,6 +228,12 @@ int main(int argc, char **argv)
             argc--;
             argv++;
         }
+	if (strncmp("--no-net-ring", *argv, 13) == 0) {
```
reynir: I suspect we need to check more precisely it is exactly `--no-net-ring` and not `--no-net-ringlalalala` :)

tenders/hvt/hvt_main.c
reynir: I'm guessing maybe no since we currently call this before `hvt_vcpu_init()` and we call `setup_modules()` after `hvt_vcpu_ini()`.

tenders/hvt/hvt_main.c
reynir: `reserve` is basically `sizeof(struct hvt_ring)` rounded up, yes. So yes, there's potentially a gap of unusable memory (can we expand the ring to fill this gap?). I don't quite understand whythe `2 * reserve` check.

We use `reserve` to subtract from `guest_mem_size` / compute where the hvt_ring is located.

dinosaure: At this stage, `hvt->guest_mem_size` is the full memory that the unikernel can have (say that `--mem=32`, `guest_mem_size = 32MB`). So `reserve` is `sizeof(struct hvt_ring)` rounded up and we place the `gpa_ring` at `guest_mem_size - roundup(sizeof(struct hvt_ring))`. `2 * reserve` is an arbitrary factor, I need to think a bit more because I suspect that, at a time, I took into account reserved memory area.

tenders/hvt/hvt_main.c
dinosaure: Yes, it must be done on the `hvt_main.c` and it's global (not per net devices).

tenders/hvt/hvt_main.c
```diff
@@ -227,6 +229,12 @@ int main(int argc, char **argv)
             argc--;
             argv++;
         }
+	if (strcmp("--no-net-ring", *argv) == 0) {
```
palainp: I would keep the `strncmp` version, e.g. `strncmp("--no-net-ring", *argv, 14) == 0` which catches the trailing '\0' and keep the string comparison bounded to n characters. Note that the `strcmp` implementation checks each characters from the first parameters only, so both are equivalent, but that would not be the case with `strcmp(*argv, "--no-net-ring")` which could lead to a very long loop.
My note is (EDIT: not) not for this PR, but maybe https://github.com/Solo5/solo5/blob/469c1a8d935b503fa553aef152d1d57d095fa3a4/bindings/lib.c#L134 should be written like (without the n part) https://github.com/Solo5/solo5/blob/469c1a8d935b503fa553aef152d1d57d095fa3a4/bindings/lib.c#L178

palainp: Actually no, the loop would not be so long as each characters are still compared ;) but anyway it's a matter of taste, I prefer the 'n' version ;)

palainp: And looking twice, `strcmp` is already used multiple times for command line arguments, so please do not take care of my suggestion ;)

dinosaure:
I think this PR is ready, probably we need a check from @pocopepe (about rumprun). I think we still have some rooms to improve the net ring but, at least, let's go back to our initial statement about memory plus the page aligned memory addressable thing.

pocopepe:
cloned the branch and gave it a run, all good on my end. rumprun boots fine and the #659 case stays fixed.

dinosaure:
Let's merge it and cut a release, I think everybody agreed with this PR now.

PR #673, opened 2026-09-06
Fix argument after --mem= being processed twice

danila-permogorskii:
The second command line pass in both tenders uses sequential `if`
statements where the branches are meant to be mutually exclusive. Once
the `--mem=` branch consumes its argument and advances `argv`, control
falls through to the next `if`, which evaluates `handle_cmdarg(*argv,
mft)` on the *following* argument within the same iteration.

On the success path this is accidentally harmless: two arguments are
consumed and `argc` is decremented twice, which is correct. On the
failure path the module diagnostic is emitted, `matched` is already set
so nothing is reported, and the next iteration runs the same argument
through `handle_cmdarg()` again — printing the diagnostic twice.

Before:

    $ ./tenders/spt/solo5-spt --mem=2 --block:nosuch=x tests/test_hello/test_hello.spt
    solo5-spt: Resource not declared in manifest: 'nosuch'
    solo5-spt: Resource not declared in manifest: 'nosuch'
    solo5-spt: Invalid option: `--block:nosuch=x'

    $ ./tenders/spt/solo5-spt --block:nosuch=x tests/test_hello/test_hello.spt
    solo5-spt: Resource not declared in manifest: 'nosuch'
    solo5-spt: Invalid option: `--block:nosuch=x'

After, both report once. `solo5-hvt` behaves identically before and
after, with `--mem=4`.

There is a second, latent consequence: when `--mem=` is the last option,
`handle_cmdarg()` is called on the kernel filename. It declines today
only because every module handler starts with a `strncmp` against its
own `--option:` prefix.

The fix gives each option its own branch ending in `continue`, making
the branches mutually exclusive by construction and removing the
`matched` flag. I chose this over an `if`/`else if` chain because the
chain makes clang-format break `--no-net-ring`'s condition across lines;
happy to switch if you prefer that shape.

hvt is included because #670 added `--no-net-ring` as a third branch in
the same chain, so the pattern is actively growing.

Tested on x86_64 Linux: `tests/run-tests.sh` shows no change from
baseline (the only failures are the `virtio` tests, which need
`qemu-system-x86_64`). Guest argument passing verified unchanged with
and without `--mem=`, with `--`, and with a valid `--block:` option
following `--mem=`.


reynir:
TL;DR: the existing code is not right, but doesn't actually cause buggy behavior.

I like the change, but I can't reproduce that it's being called twice. So when we process `--mem=` we increment `argv` and then the next argument is processed. If it's a match we further increment `argv` before looping - so here we don't process the same argv entry twice. If it's not a match we call `usage()` and exit.

It *does* make it possible to bypass the `argv[0] = '-'` check in the while loop. This is pretty harmless, though. The worst that can happen is a module an argument that doesn't start with `-`.

reynir (approved):

danila-permogorskii:
> TL;DR: the existing code is not right, but doesn't actually cause buggy behavior.
> 
> I like the change, but I can't reproduce that it's being called twice. So when we process `--mem=` we increment `argv` and then the next argument is processed. If it's a match we further increment `argv` before looping - so here we don't process the same argv entry twice. If it's not a match we call `usage()` and exit.
> 
> It _does_ make it possible to bypass the `argv[0] = '-'` check in the while loop. This is pretty harmless, though. The worst that can happen is a module an argument that doesn't start with `-`.

Hi, Reynir! 

Thank you for the review and look at my output below. I'm using wsl and Ubuntu 26.04. Can you provide the command you used? It helps me to understand the problem better

This is a command what I used to reproduce:
`
./tenders/spt/solo5-spt --mem=2 --block:nosuch=x tests/test_hello/test_hello.spt`


This is my output what I've got right now:
```
danila@olsi:~/solo5$ ./tenders/spt/solo5-spt --mem=2 --block:nosuch=x tests/test_hello/test_hello.spt
solo5-spt: Resource not declared in manifest: 'nosuch'
solo5-spt: Resource not declared in manifest: 'nosuch'
solo5-spt: Invalid option: `--block:nosuch=x'
usage: solo5-spt [ CORE OPTIONS ] [ -- ] KERNEL [ ARGS ]
KERNEL is the filename of the unikernel to run.
ARGS are optional arguments passed to the unikernel.
Core options:
  [ --mem=512 ] (guest memory in MB)
    --help (display this help)
Compiled-in modules: core net block
Compiled-in module options:
    --x-exec-heap (make the heap executable). WARNING: This option is dangerous and not recommended as it makes the heap and stack executable.
    --net:NAME=IFACE | @NN (attach tap at IFACE or at fd @NN as network NAME)
  [ --net-mac:NAME=HWADDR ] (set HWADDR for network NAME)
    --block:NAME=PATH (attach block device/file at PATH as block storage NAME)
  [ --block-sector-size:NAME=SECTORSIZE ] (set sector size for block device NAME; must be a power of two greater than or equal 512)
```

And the same for hvt:

Command:
`./tenders/hvt/solo5-hvt --mem=2 --block:nosuch=x tests/test_hello/test_hello.hvt`

Output:
```
danila@olsi:~/solo5$ ./tenders/hvt/solo5-hvt --mem=2 --block:nosuch=x tests/
test_hello/test_hello.hvt
solo5-hvt: Resource not declared in manifest: 'nosuch'
solo5-hvt: Resource not declared in manifest: 'nosuch'
solo5-hvt: Invalid option: `--block:nosuch=x'
usage: solo5-hvt [ CORE OPTIONS ] [ MODULE OPTIONS ] [ -- ] KERNEL [ ARGS ]
KERNEL is the filename of the unikernel to run.
ARGS are optional arguments passed to the unikernel.
Core options:
  [ --mem=512 ] (guest memory in MB)
  [ --no-net-ring ] (disable the net ring buffer)
    --help (display this help)
    --version (display version information)
Compiled-in modules: core block net
Compiled-in module options:
    --block:NAME=PATH (attach block device/file at PATH as block storage NAME)
  [ --block-sector-size:NAME=SECTORSIZE ] (set sector size for block device NAME; must be a power of two greater than or equal 512)
    --net:NAME=IFACE | @NN (attach tap at IFACE or at fd @NN as network NAME)
  [ --net-mac:NAME=HWADDR ] (set HWADDR for network NAME)
```

Thank you a lot for you feedback! I want to invest more time to the contribution and it helps me to understand this project better :-)

reynir:
Ah! That makes sense. So in that case the block module prints a message but doesn't match so we don't increment `argv`.

I tested by adding a `warnx("this is from block module");` in `handle_cmdarg()` in `hvt_module_blk.c` and then trying `./tenders/hvt/solo5-hvt --mem=64 --no-net-ring  ~/workspace/mirage-skeleton/device-usage/disk-lottery/dist/lottery.hvt` and adding a `--block:storage=storage` (which the unikernel expects) and different permutations. If I do `solo5-hvt --mem=64 --quux /path/to/lottery.hvt` it does indeed print twice for `--quux` (in my modified solo5-hvt).

I think this is a good change.

danila-permogorskii:
> it does indeed print twice for `--quux` (in my modified solo5-hvt).



> Ah! That makes sense. So in that case the block module prints a message but doesn't match so we don't increment `argv`.
> 
> I tested by adding a `warnx("this is from block module");` in `handle_cmdarg()` in `hvt_module_blk.c` and then trying `./tenders/hvt/solo5-hvt --mem=64 --no-net-ring ~/workspace/mirage-skeleton/device-usage/disk-lottery/dist/lottery.hvt` and adding a `--block:storage=storage` (which the unikernel expects) and different permutations. If I do `solo5-hvt --mem=64 --quux /path/to/lottery.hvt` it does indeed print twice for `--quux` (in my modified solo5-hvt).
> 
> I think this is a good change.

Thank you, reynir! If you'll have enough time, please check the #674 . May be your independent technical review may help to find something wrong

reynir:
> Thank you, reynir! If you'll have enough time, please check the #674 . May be your independent technical review may help to find something wrong

Sorry, I don't understand this part of GNU make

PR #674, opened 2026-09-06
build: Give debug objects their own dependency files

danila-permogorskii:
`tenders/GNUmakefile` builds two objects from each source, `%.o` and
`%.debug.o`. Both pattern rules produce the same stem, so both write
`$*.Td` and rename it to `$*.d`.

**1. Parallel builds race.** 2 failures in 6 clean builds at `-j16` under
WSL2 with 6 vCPUs (Windows host: 16 logical CPUs), a different file each time:

    mv: cannot stat 'hvt/hvt_cpu_x86_64.Td': No such file or directory
    make[1]: *** [GNUmakefile:48: hvt/hvt_cpu_x86_64.debug.o] Error 1
    mv: cannot stat 'hvt/hvt_kvm.Td': No such file or directory
    make[1]: *** [GNUmakefile:48: hvt/hvt_kvm.debug.o] Error 1

At `-j$(nproc)` (`-j6`) it did not reproduce in 6 attempts, though I first hit
it with plain `make -j$(nproc)` on a partially-built tree. On a different
configuration the same test failed 3 of 6 times at `-j16`, there on the `%.o` side of the
collision (`GNUmakefile:45`) rather than the `%.debug.o` side — whichever rule
finishes second is the one that fails, which is what makes it a race rather
than a broken rule. The low rate at default parallelism is presumably why CI
has not hit it.

**2. Debug objects have no header dependency tracking.** The rules
already reference a file that nothing creates:

    $ find tenders -name '*.debug.d' | wc -l
    0

`tenders/GNUmakefile:47` declares `%.debug.o: %.c %.debug.d`, line 57 has
`%.debug.d: ;`, line 58 lists it in `.PRECIOUS`, and `all_DEPS` at line 146
computes `foo.debug.d` — but the recipe writes `foo.d`. Line 154's
`include $(wildcard $(all_DEPS))` then drops the missing file silently.
The effect:

    $ stat -c %y tenders/hvt/hvt_module_gdb.debug.o
    2026-09-06 11:12:41
    $ touch tenders/hvt/hvt.h && make
    $ stat -c %y tenders/hvt/hvt_module_gdb.debug.o
    2026-09-06 11:12:41

**3. `make clean` leaves orphans**, for the modules built only as debug
objects:

    $ make clean && ls tenders/hvt/*.d
    tenders/hvt/hvt_module_dumpcore.d
    tenders/hvt/hvt_module_gdb.d

Consequence 2 is the damaging one. `struct hvt_module` goes into the
`modules` linker section and `hvt_main.c:46` walks it with `m++`, which
requires every contributing object to agree on the struct size. I hit
this on a long-lived tree: `hvt_module_gdb.debug.o` was nine minutes
older than the `hvt.h` it was compiled against and had never been
rebuilt, carrying a 64-byte `struct hvt_module` where every other module
object had 56. `solo5-hvt-debug` then aborts:

    solo5-hvt-debug: hvt/hvt_main.c:48: setup_modules:
    Assertion `m->ops.setup' failed.

and `dumpcore hvt` fails. `make clean` hides it, which is why it does
not show up in CI — it only bites developers with long-lived trees.

**The fix** adds a `DEPBASE` variable defaulting to `$*`, and overrides
it to `$*.debug` for the `%.debug.o` rules. `bindings/` and `elftool/`
have no `%.debug.o` rules and their dependency filenames are unchanged.
This uses a GNU Make pattern-specific variable value; the project
already requires GNU Make (hence `GNUmakefile`).

I also considered keying the files off `$@` (`foo.o.d`,
`foo.debug.o.d`), which would prevent the whole class rather than this
instance. I did not propose it because it renames every dependency file
and needs matching changes in all three of `tenders/`, `bindings/` and
`elftool/` — two of which do not have the problem. Happy to do it that
way if you prefer.

**After the patch:** 11 `.debug.d` files exist and each targets its own
object; touching `hvt.h` rebuilds the debug objects; every `__module_*`
symbol is `0x38`; `make clean` leaves nothing; 5 of 5 clean builds at
`-j16` succeed. `tests/run-tests.sh` is unchanged from baseline —
`82 tests, 12 failures`, all `virtio` for lack of
`qemu-system-x86_64` locally.


hannesm:
was there any so-called AI used for preparing this PR (code/patch? comment above?)?

danila-permogorskii:
> was there any so-called AI used for preparing this PR (code/patch? comment above?)?

Hi, yes. I'm using ai for prepare the description and code analysing. But the code itself/commits written by me manually. I checked the OCaml rules for today and ready to rewrite the description if it's necessary.

dinosaure:
I think the description is too wordy and difficult to understand (especially for someone who isn’t fluent in English). To put it more simply, this omission essentially stems from #644 (there’s no need to go into quite so much detail...). I am still curious as to why `DEPBASE` is necessary, though.

The review process in question isn’t just about determining whether the patch is correct or not, but also about being able to present such a proposal in a way that makes it understandable and legitimate.

To be clear, I would be more inclined to reject such pull requests even if they might improve Solo5: it’s also a matter of respecting the time we can devote to maintaining Solo5, and reading this PR is, quite frankly, very frustrating. More specifically, reworking the commit and its description (to avoid an LLM-generated output) and fully understanding the purpose of DEPBASE would lead us to eventually merge this pull request.

PR #678, opened 2026-09-09
Check the version of the C compiler on invocations

shym:
The Solo5 toolchain is tied to the version of the C compiler that was found when it was generated. This adds a new special header that is included on every compilation invocations to check that the version of the C compiler is the same as when it was built.

If I understand correctly, the issue was raised by @kit-ty-kate, @n-osborne and @Firobe. The idea to use a lightweight header to check the used compiler was suggested by @edwintorok.


Firobe:
The existing issue can be observed for example in the beginning of this thread: https://github.com/mirage/mirage-crypto/issues/276

dinosaure (approved):

n-osborne:
Thanks.
Just a small remarks about the warning message: is the `it` in `reinstall it` a bit ambiguous?
Maybe a more verbose `reinstall the opam solo5 package` would be clearer?

shym:
Note that it could be the `solo5-cross-aarch64` package, so I added a (squashable if accepted) commit that will propagate the name of the package being built to that header. (Maybe it be better placed as a configure argument, though; so I just fix it that way).

hannesm:
I'm impressed by the amount of positive feedback this PR has, so there is likely some need.

Now, I wonder what the API stability of C compilers is (both claimed and practical experience)? Are we sure that any patch version may break solo5? I mean of course we can stick to the conservative approach and always ask the user to recompile, at the cost of wasting CPU cycles...

edwintorok:
> I'm impressed by the amount of positive feedback this PR has, so there is likely some need.
> 
> Now, I wonder what the API stability of C compilers is (both claimed and practical experience)? Are we sure that any patch version may break solo5? I mean of course we can stick to the conservative approach and always ask the user to recompile, at the cost of wasting CPU cycles...

I think this is currently a warning, instead of a hard error. So if something does fail later on, then you'd have some hint about what went wrong and how to fix it,  instead of some cryptic error.
I don't think there'd be expectations of API stability when it comes to a compiler's internal headers, those are supposed to match exactly the compiler you're using (it is installed by the same RPM package as the compiler itself, so in practice they'd never diverge, unless something has made a copy, like solo5...).
Would be nice if we could avoid copying them in the first place, but symlinking might run into problems still.
Looking at its path, on my system it includes the major version of the compiler, so it'd be expected to break at least across major versions:
```
$ rpm -q --whatprovides /usr/lib64/gcc/x86_64-suse-linux/16/include/stdarg.h
gcc16-16.2.0+git9497-3.1.x86_64
```

hannesm:
@edwintorok yes, so I don't argue that on major version changes that is the right path forward.

about the implementation, I understand the following:
- toolchain/gen_headers.sh uses gcc-version-check.h.in or clang-version-check.h.in, fills in the data, and moves it to solo5-compiler-check.h
- clang-version.h and gcc-verison.h are needed for the invocation (CC -E -P)
- that solo5-compiler-check.h is included by cc.in
- configure.sh now passes around a "package" so the warning is specific about which opam package to reinstall

this sounds ok, but is quite some machinery - would it be worth in gen_headers.sh to drop a comment on what is done and why?

more notes:
- about the solo5-compiler-check.h being a warning -- I wonder whether it should be an error instead?
- about the `#if.. #endif` in the gcc-version-check.h.in/clang-version-check.h.in, should there be an `#else` branch for the unlikely situation where suddenly `cc` is a different compiler kind?
- about naming: would it be possible to have them unified? I like `solo5-compiler-check.h`, so could it be `clang-compiler-check.h.in` and `gcc-compiler-check.h.in` (or, is there a path where we only have a single `solo5-compiler-check.h.in` (and `compiler-version.h`)?

thanks for your work on this! :)

shym:
First of all, thank you for your reviews!

> about the implementation, I understand the following: […]

All correct! And I admit that it’s pretty involved, with some information that should be kept consistent spread across multiple files, that would be harder to maintain than what’s really necessary. The couple of rounds of work on the sibling PR in OCaml/Unikraft (mirage/ocaml-unikraft#18) give me also ideas for a simpler version of the generation of the header to include (but the case in OCaml/Unikraft is simpler to start with, as less platforms are actually supported). I’ll (eventually) update this PR with a more localized version (and with comments).

> * about the solo5-compiler-check.h being a warning -- I wonder whether it should be an error instead?

That was a middle path chosen to avoid the mandatory rebuilds on every compiler updates, if that turns out not to be necessary.

> * about the `#if.. #endif` in the gcc-version-check.h.in/clang-version-check.h.in, should there be an `#else` branch for the unlikely situation where suddenly `cc` is a different compiler kind?

I’m not sure I understand: the first clause in the `#if` test checks that there’s no change from GCC to clang or vice versa. Isn’t it what you have in mind?

> * about naming: would it be possible to have them unified? I like `solo5-compiler-check.h`, so could it be `clang-compiler-check.h.in` and `gcc-compiler-check.h.in` (or, is there a path where we only have a single `solo5-compiler-check.h.in` (and `compiler-version.h`)?

Yes, I think the OCaml/Unikraft’s PR shows a way to merge all those.

But maybe I should have started with an analysis of the current status which explains the dependencies.

- When using GCC, `gen-headers.sh` copies all its internal headers; those headers are located, at least in all the distributions I have seen until now, in directories containing at least the GCC major number but I assume that’s (only) to make different major versions co-installable rather than to state that minor version updates of the compiler should be compatible with any other minor version of the same major version (ie I agree with Edwin :-). Note that NixOS seems to use the full version number in its paths, rather unsurprisingly. This is why this PR proposed to warn whenever even the patch level changes: better be warned than lost, we might say.

  Another idea in the GCC space would be to go the same way as what is done in OCaml/Unikraft and use `-isystem $(gcc -print-file-name=include)` in `cc.in` instead of the directory in which all the content of that include directory has been copied. This is related to the current PR because it makes more sense when combined with a sanity check that the version is indeed found since `gcc -isystem /nonexistant` generates no error.

  A final idea after thinking about all this: we might refine further the sanity check so that, when the major version number of the compiler changes, we generate an error, but when it’s only the minor or patch-level version number, we generate only a warning.

- When using clang, it depends on the platform:
  1. on FreeBSD, OpenBSD and DragonFly, we copy some headers from `/usr/include`; so there, the actual version number of the compiler is not the relevant information to check for consistency; there we might be looking for other relevant `#define`s, such as one with the OS version number, maybe?
  2. sometimes we don’t copy anything and we don’t even keep track of where the compiler is (according to opam, the only supported OS that’s not in the first case is Linux), so I guess no check is really necessary there.

shym:
I’ve updated (and rebased) this PR with the changes I had in mind in my previous comment (except for the change of pointing `-isystem` directly to GCC’s directory instead of copying it), to make those ideas a bit clearer.
I triggered a run ([commit](https://github.com/shym/solo5/commit/2b7942f113bb90955b2bee3494dab046a817e8ec)) to log the generated header.

hannesm:
Thanks for your work, although this is getting more obscure to me.

For starters, it is good that you check that the used compiler for compiling solo5 and the compiler used now are the same (so indeed switching from gcc to clang will report an error).

So why is the full version checked for gcc and not for clang? Are there any guarantees that I've missed? Why is the FreeBSD (major) version in the clang case? What about OpenBSD, DragonFlyBSD, Linux?

I find this new version pretty ad-hoc (and a rather complex adjustment of configure and build of solo5) and would like to understand what are the goals? If all it is about "providing a better error message", the question is still whether this should be a suggestion (warning) or an error -- and how conservative it should be.

shym:
> although this is getting more obscure to me.

Sorry about that :-/

> So why is the full version checked for gcc and not for clang? Are there any guarantees that I've missed?

Because GCC internal headers are copied (and we don’t assume that there are compatibility guarantees between even patch-level versions), while Clang headers are not.

> Why is the FreeBSD (major) version in the clang case? What about OpenBSD, DragonFlyBSD, Linux?

Because some FreeBSD (and OpenBSD and DragonFly) system headers are copied (not Linux’s). But OpenBSD and DragonFly’s clang do not expose their versions in a macro (`__OpenBSD__` and `__DragonFly__` are just set to 1, so checking that macro seems useless).

I should add those two points to the comment in `gen-headers` if they’re helpful.

shym:
To make progress on this PR, I’ve:
- kept only the test we all agree upon: whether the compiler is Clang or GCC and its major version number; this simpler test makes it possible to remove a lot of the clutter,
- extracted into a separate commit the very noisy work of propagating the name of the opam package from the `opam` file up to the generated header.

The separation into 2 commits makes it a lot easier to read through, I hope (at least I find it simpler for myself one week later).

PR #679, opened 2026-09-16
Use the intended compiler in amd64 ubuntu CI tests

shym:
Set both `HOST_CC` and `TARGET_CC` so that clang compiler is used instead of the default gcc on that platform for the “clang” matrix variant. (The `CC` environment variable is never used by the build system)

This was noticed in a test run for #678: see the [generated header](https://github.com/shym/solo5/actions/runs/35084228634/job/104755139381#step:4:75) for that job.


dinosaure:
I don't understand why but your change broke one test:
```
not ok 32 wnox spt
# (in test file tests.bats, line 402)
#   `[ "$status" -eq 139 ] # SIGSEGV' failed
# solo5-spt: test_wnox/test_wnox.spt: invalid executable: phdr[2] requests WRITE and EXEC permissions
```

hannesm:
maybe wnox with clang on Linux is broken? and only works with gcc... before this PR as I understand only gcc was used on Linux.

shym:
That’s also my understanding: clang was not properly tested and that test doesn’t work with it.
Run locally on a amd64 linux, where `cc` (the default) is `gcc`:
```
$ ./configure.sh
[…]
$ make -j8
[…]
$ readelf -Wl tests/test_wnox/test_wnox.spt

  LOAD           0x001000 0x0000000000100000 0x0000000000100000 0x003de8 0x003de8 R E 0x1000
  LOAD           0x005000 0x0000000000104000 0x0000000000104000 0x000011 0x000011 RW  0x1000
  LOAD           0x000238 0x0000000000000000 0x0000000000000000 0x000000 0x000000 R   0x1000
  TLS            0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000     0x8
  LOAD           0x001000 0x0000000000105000 0x0000000000105000 0x000000 0x000058 RW  0x1000

#####################################

$ make clean
[…]
$ TARGET_CC=clang ./configure.sh
[…]
$ make -j8
[…]
$ readelf -Wl tests/test_wnox/test_wnox.spt

  LOAD           0x001000 0x0000000000100000 0x0000000000100000 0x004648 0x004648 R E 0x1000
  LOAD           0x006000 0x0000000000105000 0x0000000000105000 0x000011 0x000011 RWE 0x1000
  LOAD           0x000238 0x0000000000000000 0x0000000000000000 0x000000 0x000000 R   0x1000
  TLS            0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000     0x8
  LOAD           0x001000 0x0000000000106000 0x0000000000106000 0x000000 0x000058 RW  0x1000
```

shym:
I’m curious: does this test succeed with clang on other platforms (such as FreeBSD, where it’s the default cc nowadays, isn’t it?)?

dinosaure:
I get some insights here: #680. It seems that GCC/ld and clang is not a funny mix.

hannesm:
So for the failure, I'm curious (as discussed with @dinosaure) whether there's any use for gcc & ld.lld (from clang) OR clang & gnu ld? considering how tightly clang & ld.lld // gcc & gnu ld are developed, I would drop the combinations from the support matrix (as @dinosaure pointed out, in configure on FreeBSD using clang with gnu ld is already rejected).

any thoughts? is there any linux distribution mixing clang & gnu ld?

hannesm:
I figured that clang & gnu ld is common on Linux systems, so we should support that combination (as it is tested with this PR).

IMHO this is good to be merged, although it uncovers a failed thing, so we should take a look (review) of #680.

shym:
> IMHO this is good to be merged, although it uncovers a failed thing, so we should take a look (review) of #680.

Great!
It would also make sense to merge it as part of #680.

dinosaure:
Integrated into #680

dinosaure:
Thanks!

PR #680, opened 2026-09-17
Force split rodata on spt artifacts

dinosaure:


dinosaure:
Ok it seems that we have some orphans sections:
```
make[2]: Entering directory '/home/runner/work/solo5/solo5/tests/test_wnox'
CC test_wnox.o
ELFTOOL manifest.c
CC manifest.o
CCLD test_wnox.stub
CCLD test_wnox.hvt
/usr/bin/ld: warning: test_wnox.hvt has a LOAD segment with RWX permissions
CCLD test_wnox.spt
/usr/bin/ld: warning: test_wnox.spt has a LOAD segment with RWX permissions
CCLD test_wnox.virtio
/usr/bin/ld: warning: test_wnox.virtio has a LOAD segment with RWX permissions
CCLD test_wnox.muen
/usr/bin/ld: warning: test_wnox.muen has a LOAD segment with RWX permissions
CCLD test_wnox.xen
/usr/bin/ld: warning: test_wnox.xen has a LOAD segment with RWX permissions
SUBST test_wnox.xl
make[2]: Leaving directory '/home/runner/work/solo5/solo5/tests/test_wnox'
```

I'm currently reading this blog post: https://maskray.me/blog/understanding-orphan-sections and it seems related to a mix between `clang` and `ld` (from GCC)...

dinosaure:
It’s clearly a work in progress, but it seems that the combination of Clang and GNU ld doesn’t work well together. It appears that Clang generates quite a few sections, and GNU ld doesn’t know where to place them, so it puts them (it seems) in the `data` segment by default. However, some of these sections require the X flag by default, which causes `wnox` to fail. In this PR, we’re adding the `--orphan-handler=error` option to identify all "orphan" sections and specify where to place them. **This is a debug-only option** that we should remove if we want to keep the `debug_*` sections in particular (or set it to `warn`). Next, we specify where to place some of them (specifically in the `data` segment) and force this segment to be RW rather than X (in the hope that the linker will fail if any of the sections REALLY needs the X flag, which I’m not sure about).

In short, this is a pull request that needs reworking but which explains more precisely the problem we’re having with #679.

configure.sh
```diff
@@ -513,6 +513,9 @@ case ${CONFIG_HOST} in
         TARGET_LD="${TARGET_LD:-ld}"
         TARGET_OBJCOPY="${TARGET_OBJCOPY:-objcopy}"
         TARGET_CC_LDFLAGS="-Wl,--build-id=none -no-pie"
+        if CC="${TARGET_CC}" cc_is_clang && ! LD="${TARGET_LD}" ld_is_lld; then
+          TARGET_CC_LDFLAGS="${TARGET_CC_LDFLAGS} -Wl,--orphan-handling=warn"
```
dinosaure: Do we want to keep it? It shows multiple warnings about `debug_` sections, we can remove the specific commit which introduces this option if we want.

shym: By `sort -u`-ing those entries, I see:
```
/usr/bin/ld: warning: orphan section `.comment' from ... `.comment'
/usr/bin/ld: warning: orphan section `.debug_abbrev' from ... `.debug_abbrev'
/usr/bin/ld: warning: orphan section `.debug_addr' from ... `.debug_addr'
/usr/bin/ld: warning: orphan section `.debug_aranges' from ... `.debug_aranges'
/usr/bin/ld: warning: orphan section `.debug_frame' from ... `.debug_frame'
/usr/bin/ld: warning: orphan section `.debug_info' from ... `.debug_info'
/usr/bin/ld: warning: orphan section `.debug_line' from ... `.debug_line'
/usr/bin/ld: warning: orphan section `.debug_line_str' from ... `.debug_line_str'
/usr/bin/ld: warning: orphan section `.debug_loclists' from ... `.debug_loclists'
/usr/bin/ld: warning: orphan section `.debug_rnglists' from ... `.debug_rnglists'
/usr/bin/ld: warning: orphan section `.debug_str' from ... `.debug_str'
/usr/bin/ld: warning: orphan section `.debug_str_offsets' from ... `.debug_str_offsets'
/usr/bin/ld: warning: orphan section `.igot.plt' from ... `.igot.plt'
/usr/bin/ld: warning: orphan section `.iplt' from ... `.iplt'
/usr/bin/ld: warning: orphan section `.note.GNU-stack' from ... `.note.GNU-stack'
/usr/bin/ld: warning: orphan section `.rela.data.multiboot' from ... `.rela.dyn'
/usr/bin/ld: warning: orphan section `.rela.got' from ... `.rela.dyn'
/usr/bin/ld: warning: orphan section `.rela.iplt' from ... `.rela.dyn'
/usr/bin/ld: warning: orphan section `.rela.text' from ... `.rela.dyn'
```
Maybe we could also copy the standard handling of those sections from `ld --verbose`?

dinosaure: That’s sort of what worries me. From what I understand, Solo5 seems to be quite flexible when it comes to sections and where they should be placed, depending on the behaviour of the linker used. We could be exhaustive today, but that would require keeping a closer eye on developments in GNU ld (which is clearly under-specified) and lld. The real question is: do we really need to manage all these sections manually at the moment?

I also think that this flexibility allows us to work ‘fine’ across different platforms (from Linux to FreeBSD via OpenBSD), whereas meticulously tweaking every section generated by GCC or Clang and deciding where to place them could lead to silent failures (with warnings) or outright failures (with errors).

For the sake of maintenance and the time we don’t necessarily have, I would be in favour of removing this option.

shym: > For the sake of maintenance and the time we don’t necessarily have, I would be in favour of removing this option.

That’s sensible.

dinosaure: I mean, it's actually the case, not sure if we can automate a way to fill our link scripts automatically according to what GCC or clang give to us.

hannesm: To me, these sections could just be removed from the resulting binary. I mean for the runtime we manually map the sections we need / we have. But those additional sections, I think we can just drop.

I don't know much about linker scripts / linkers, but maybe there's a "drop sections not mentioned here"?

shym: Just in case, I’ve stumbled upon a `/DISCARD/ : { *(*) }` in `ld`’s manual that might do what you have in mind and there’s also the option `--orphan-handling=discard`. But, as someone who has spent quite some time looking at a unikernel in a debugger (through QEMU for OCaml/Unikraft, but anyway), I’d prefer to keep the current behaviour, which is IIUC to keep the debug info, or at least to offer some way to keep it (a flag (I’m thinking another Solo5-specific `-z` one) and the command-line discard option are probably a simpler way to go for that).

hannesm:
@dinosaure IMHO it is fine to remove the commit with the option -- but I'd defer it to you whether to keep it or not.

@shym thanks for your pointer to DISCARD. I guess this should if at all be a separate PR -- and as you mention, some of the debug sections are useful.

